PR #6, opened 2024-04-08
Rework mux demux

bwiesmayr:
Reworked the struct manipulators to use the configurable FB infrastructure, including the possibility to have error markers in STRUCT_MUX/STRUCT_DEMUX. Fixed problems in update type, which led to losing the configuration of visible children when updating a struct manipulator.


plugins/org.eclipse.fordiac.ide.fbtypeeditor.fbtester/.project
azoitl, line 1: this file should be removed

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
```diff
@@ -19,22 +19,46 @@
...
+import org.eclipse.fordiac.ide.model.libraryElement.impl.ConfigurableFBManagement;
 
 public class ChangeStructCommand extends AbstractUpdateFBNElementCommand {
 
-	private final StructuredType newStruct;
+	private final DataType newStruct;
+	private final String newVisibleChildren;
+
+	public ChangeStructCommand(final StructManipulator mux) {
+		super(mux);
```
azoitl: would this be better a this call to the constructor below?

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
bwiesmayr: yes but I didn't want to have the public constructor with "DataType", although I internally need to handle DataType because it could be an ErrorDataType which is then not a struct.... so I propose to wait for some tests and then possibly do further cleanups.

azoitl (approved):

PR #10, opened 2024-04-08
Bug 583131 - Update instances after type change

azoitl:
If a type entry has changed it will notify its corresponding editors to trigger a type update


azoitl:
migrated from @oberlehner's gerrit commit: https://git.eclipse.org/r/c/4diac/org.eclipse.4diac.ide/+/206876

PR #12, opened 2024-04-08
Add the option to set retain options for FB pins

azoitl:


azoitl:
pushed as migration from gerrit on behave of @MandKastner   https://git.eclipse.org/r/c/4diac/org.eclipse.4diac.ide/+/206819

PR #14, opened 2024-04-08
Bug 583134 - Show error message in not loadable editors

azoitl:


bwiesmayr (approved):
code looks good, very useful patch

PR #23, opened 2024-04-09
Merge Develop to freeze

azoitl:


oberlehner (dismissed):

azoitl:
@oberlehner only if I do a rebase and merge. If I do a merge commit, which is the right thing for these kinds of PRs then it should be fine.

PR #24, opened 2024-04-09
border crossing reconnect

lindhuber:


oberlehner:
image

This use case is not working.

oberlehner:
image
Moving from here to here should resized the lower subapp

azoitl (approved):

PR #33, opened 2024-04-10
Add library delete handling

pazi146:
on library link deletion:
let the user decide if dependency gets removed


oberlehner:
It works for that case:

image

But not for that case:
image



image

pazi146:
Can you show the .project file?
I suspect that "WORKSPACE_LOC\.lib\lib5-0.5.3\MANIFEST.MF" isn't properly linked
(I also think you created this project with an intermediate version as the linked folder should be named "lib5" instead of "lib5-0.5.3")

oberlehner:
> Can you show the .project file? I suspect that "WORKSPACE_LOC.lib\lib5-0.5.3\MANIFEST.MF" isn't properly linked (I also think you created this project with an intermediate version as the linked folder should be named "lib5" instead of "lib5-0.5.3")

You are right. i tried it with lib5 and it worked well.

pazi146:
Can you show the .project file?
I suspect that "WORKSPACE_LOC\.lib\lib5-0.5.3\MANIFEST.MF" isn't properly linked
(I also think you created this project with an intermediate version as the linked folder should be named "lib5" instead of "lib5-0.5.3")

PR #37, opened 2024-04-10
Update instances in editor

azoitl:
On change of dependent types the editor is automatically performing an update type on the children using a type.

Type save only informs about affected types but does not change them anymore.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeEditor.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeEditor.java
@@ -26,9 +26,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.Objects;
-import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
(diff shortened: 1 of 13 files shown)


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeEditor.java
```diff
@@ -157,7 +152,7 @@ private void createSaveDialog(final IProgressMonitor monitor) {
 		// Depending on the button clicked:
 		switch (fbSaveDialog.open()) {
 		case DEFAULT_BUTTON_INDEX:
-			doSaveWithUpdate(monitor, fbSaveDialog.getDataHandler().getCollectedElements());
+			doSaveInternal(monitor);
```
oberlehner, line 155: I think this should be executed no matter what the user selects

oberlehner (commented):
It works perfectly fine.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeEditor.java
azoitl: No because if the user selects cancel we must not save.

oberlehner: Ok, I thought the requirements were that it would be saved no matter what the user said. But maybe I'm remembering that wrong.

PR #56, opened 2024-04-16
Resolve target pin labels in follow connection handler

MandKastner:
The follow connection handler can now directly resolve the border labels for connections in an expanded subapp and will take the ref element of the border pin


azoitl:
Cherrypicked changes for freeze branch. Removed Java 21 feature.

Feature works as intended.

PR #59, opened 2024-04-16
Use new struct search in save of structs

azoitl:
Removed now unecessary update of struct instances.


oberlehner:
The search works fine!
What is currently not working is:
-Update of initial value in untyped subapp and struct member

oberlehner (approved):

azoitl:
> The search works fine! What is currently not working is: -Update of initial value in untyped subapp and struct member

I think this is something to be fixed in the data type editor. Would look into that in a separate PR.

PR #60, opened 2024-04-17
Fix LibraryElement in Manifest

pazi146:
fixed LibraryElement tag in xsd
added LibraryElement handling to conversion method

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library.model/model/library.xsd
+++ b/plugins/org.eclipse.fordiac.ide.library.model/model/library.xsd
@@ -17,11 +17,22 @@
       <xsd:element maxOccurs="unbounded" minOccurs="0" name="Attribute" nillable="true" type="xsd:string"/>
     </xsd:sequence>
     <xsd:attribute name="Name" type="xsd:string"/>
-    <xsd:attribute name="SymbolicName" type="xsd:string"/>
+    <xsd:attribute name="SymbolicName" type="SymbolicName"/>
     <xsd:attribute name="Comment" type="xsd:string"/>
   </xsd:complexType>
+  <xsd:simpleType name="SymbolicName">
+    <xsd:restriction base="xsd:token">
+      <xsd:pattern value="[a-zA-Z][-_a-zA-Z0-9]*"/>
+    </xsd:restriction>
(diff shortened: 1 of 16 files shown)


plugins/org.eclipse.fordiac.ide.library.model/model/library.xsd
```diff
@@ -59,5 +77,15 @@
...
-  <xsd:complexType name="LibraryElement"/>
+  <xsd:complexType name="LibraryElement">
+      <xsd:simpleContent>
+        <xsd:extension base="NameSpaceFilter">
+        </xsd:extension>
+      </xsd:simpleContent>
+  </xsd:complexType>
+  <xsd:simpleType name="NameSpaceFilter">
+    <xsd:restriction base="xsd:token">
+      <xsd:pattern value="((_[a-zA-Z0-9]|[a-zA-Z])(_?[a-zA-Z0-9])*|\*\*?)(::((_[a-zA-Z0-9]|[a-zA-Z])(_?[a-zA-Z0-9])*|\*\*?))*"/>
```
ernstblechaPT, line 88: Identifiers can not start with a digit - I also would have expected the pattern to be more like `([_a-zA-Z][_a-zA-Z0-9]*|\*\*?)(::([_a-zA-Z][_a-zA-Z0-9]*|\*\*?))*`

ernstblechaPT (changes requested):

plugins/org.eclipse.fordiac.ide.library.model/model/library.xsd
pazi146: The given pattern doesn't allow a digit at the start - mind the underscore at the very start ("\_3" is a valid identifier according to the standard)
Also the repeated use of underscore isn't allowed and the identifier also can't end with one ("a__b" and "foo_" are not valid) - this forces the pattern to be more complex

ernstblechaPT: You are right, the rule for multiple underscore threw me off

ernstblechaPT (approved):

azoitl (approved):

PR #63, opened 2024-04-17
Gef marqueeselection

azoitl:


azoitl:
All tets pass and it is only cleaning GEF Classic API usage.

PR #64, opened 2024-04-17
Bug 582792 MVP set sleeptime in debug view

azoitl:


plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/FBDebugView.java
```diff
@@ -116,16 +118,22 @@ public Rectangle getClientArea(final Rectangle rect) {
 	private static final int NUM_COLUMNS = 1;
 	private KeyHandler sharedKeyHandler;
 	private RepeatEventAction repeatEventAction;
+	private static DebugTimeEditPart debugTimeEditPart;
```
mx990: Why is this field static?

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/FBDebugView.java
```diff
@@ -116,16 +118,22 @@ public Rectangle getClientArea(final Rectangle rect) {
...
 	public void createPartControl(final Composite parent) {
 		getSite().getWorkbenchWindow().getSelectionService().addSelectionListener(this);
 		GridLayoutFactory.fillDefaults().numColumns(NUM_COLUMNS).margins(0, 0).generateLayout(parent);
 		createGraphicalViewer(parent);
+		createTestRecorderComponent(parent);
 		createToolBarEntries();
 		hookDebugListeners();
 	}
 
+	private static void createTestRecorderComponent(final Composite parent) {
```
mx990: I would suggest renaming this method, since it does not seem to create a "test recorder" component.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -138,6 +235,29 @@ public void performApply(final ILaunchConfigurationWorkingCopy configuration) {
...
+			configuration.setAttribute(FBLaunchConfigurationAttributes.CLOCK_MODE,
+					FBLaunchConfigurationAttributes.MANUAL);
+		}
+		try {
+			final String debug_time = debugTimeText.getText();
+			if (debug_time != null) {
+				configuration.setAttribute(FBLaunchConfigurationAttributes.CLOCK_INTERVAL, debug_time);
+			} else {
+				configuration.setAttribute(FBLaunchConfigurationAttributes.CLOCK_INTERVAL, "0"); //$NON-NLS-1$
+			}
```
mx990: I believe `Text.getText()` never returns `null`? I would also suggest fixing the variable name to conform to the camel-case naming convention.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -98,12 +111,93 @@ protected Composite createEventComponent(final Composite parent) {
...
+		GridLayoutFactory.swtDefaults().numColumns(3).applyTo(c);
+		GridDataFactory.fillDefaults().grab(true, false).applyTo(c);
+
+		debugTimeText = new Text(c, SWT.BORDER);
+		debugTimeText.setText("0"); //$NON-NLS-1$
+		debugTimeText.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1));
+		debugTimeText.addSegmentListener(e -> updateLaunchConfigurationDialog());
+		debugTimeLabel = new Label(c, SWT.NONE);
+		debugTimeLabel.setText("ms"); //$NON-NLS-1$
+		debugTimeLabel.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false, 1, 1));
```
mx990: It seems the `Text` and `Label` both occupy half the available space, since they both grab horizontal space. I would suggest setting the `grabExcessHorizontalSpace` parameter to `false` here.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -14,33 +14,52 @@
...
 import org.eclipse.fordiac.ide.model.eval.fb.FBEvaluatorCountingEventQueue;
 import org.eclipse.fordiac.ide.model.eval.fb.FBEvaluatorExternalEventQueue;
 import org.eclipse.fordiac.ide.model.libraryElement.Event;
 
 public class FBLaunchEventQueue implements FBEvaluatorCountingEventQueue, FBEvaluatorExternalEventQueue {
 	private final AtomicBoolean repeat;
 	private final AtomicBoolean blocking;
+	private final AtomicBoolean system;
+	private final AtomicBoolean incremental;
+	private final AtomicBoolean manual;
```
mx990: I would suggest defining an enum (and using an AtomicReference here), since these modes are mutually-exclusive. Otherwise, switching between modes may cause race conditions where more than one mode is `true` or all of them are `false`.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -53,6 +72,7 @@ public Event receiveInputEvent() throws InterruptedException {
 			incrementEventCount(result);
 			if (repeat.get()) {
 				queue.add(result);
+				setDebugTime();
```
mx990: It seems this does not properly set the time for the first event when repeat is off.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -14,33 +14,52 @@
...
 import org.eclipse.fordiac.ide.model.libraryElement.Event;
 
 public class FBLaunchEventQueue implements FBEvaluatorCountingEventQueue, FBEvaluatorExternalEventQueue {
 	private final AtomicBoolean repeat;
 	private final AtomicBoolean blocking;
+	private final AtomicBoolean system;
+	private final AtomicBoolean incremental;
+	private final AtomicBoolean manual;
+	private Duration incrementDebugTime;
+	private Duration currentDebugTime;
```
mx990: I believe these `Duration`s should be wrapped in `AtomicReference`s, since they may be accessed concurrently.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -79,31 +100,137 @@ public AtomicInteger getCount(final Event ev) {
...
+	}
+
+	/**
+	 * Set the Evaluator Process.
+	 *
+	 * @param evaluatorProcess the current EvaluatorProcess
+	 */
+	public void setEvaluatorProcess(final EvaluatorProcess evaluatorProcess) {
+		evaluator = evaluatorProcess;
+	}
```
mx990, line 195: It seems this method is called from the *FB Debug* view to update the current evaluator, which does not work when the view is not open. It also generally seems both very fragile and an inversion of responsibilities.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -79,31 +100,137 @@ public AtomicInteger getCount(final Event ev) {
...
+
+	/**
+	 * Set Debug Time in Evaluator depending on mode selection.
+	 *
+	 */
+	public void setDebugTime() {
+		final Instant instant = Instant.ofEpochSecond(currentDebugTime.getSeconds(), currentDebugTime.getNano());
+		evaluator.getExecutor().setClock(Clock.fixed(instant, ZoneId.systemDefault()));
```
mx990: If the evaluator has not been set by the UI (when the *FB Debug* view is not open) this leads to an NPE.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -119,7 +213,10 @@ public void initializeFrom(final ILaunchConfiguration configuration) {
 			}
 			repeatEventCheckbox.setSelection(FBLaunchConfigurationAttributes.isRepeatEvent(configuration));
 			keepDebuggerRunningCheckbox
-			.setSelection(FBLaunchConfigurationAttributes.isKeepRunningWhenIdle(configuration));
+					.setSelection(FBLaunchConfigurationAttributes.isKeepRunningWhenIdle(configuration));
+			systemTimeRadio.setSelection(FBLaunchConfigurationAttributes.isSystem(configuration));
+			incrementTimeRadio.setSelection(FBLaunchConfigurationAttributes.isIncrement(configuration));
+			manualTimeRadio.setSelection(FBLaunchConfigurationAttributes.isManual(configuration));
```
mx990, line 181: It seems this does not initialize the time from the configuration, meaning the time is lost when reopening the launch configuration dialog.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -98,12 +111,93 @@ protected Composite createEventComponent(final Composite parent) {
...
+		final Composite c = new Composite(group, SWT.NONE);
+		GridLayoutFactory.swtDefaults().numColumns(3).applyTo(c);
+		GridDataFactory.fillDefaults().grab(true, false).applyTo(c);
+
+		debugTimeText = new Text(c, SWT.BORDER);
+		debugTimeText.setText("0"); //$NON-NLS-1$
+		debugTimeText.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1));
+		debugTimeText.addSegmentListener(e -> updateLaunchConfigurationDialog());
+		debugTimeLabel = new Label(c, SWT.NONE);
+		debugTimeLabel.setText("ms"); //$NON-NLS-1$
```
mx990, line 149: I believe when specifying a fixed time instead of an increment using mode *Manual*, a precise timestamp with a unit other than milliseconds may make sense here?

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/editparts/DebugTimeEditPart.java
```diff
@@ -0,0 +1,265 @@
...
+
+	private static final int NUM_COLUMNS = 1;
+	private static final int NUM_BUTTONS = 3;
+	private static EvaluatorProcess evaluator;
+	private static FBLaunchEventQueue eventQueue;
+	private final Group debugTimeGroup;
+	private static Button systemTimeRadio;
+	private static Button incrementTimeRadio;
+	private static Button manualTimeRadio;
+	private static Text debugTimeText;
```
mx990: Why are all of these fields static?

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -98,12 +111,93 @@ protected Composite createEventComponent(final Composite parent) {
...
+		debugTimeLabel.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false, 1, 1));
+
+		return group;
+	}
+
+	protected static void setCheckbox(final String text) {
+		// only the most recent selected box is checked
```
mx990: I believe this is not necessary, since all RADIO buttons inside a container should update their selected state to be mutually-exclusive automatically.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -42,15 +42,24 @@
...
+import org.eclipse.swt.widgets.Text;
 
 public abstract class FBLaunchConfigurationTab extends MainLaunchConfigurationTab {
 
 	private ComboViewer eventCombo;
 	private Button repeatEventCheckbox;
 	private Button keepDebuggerRunningCheckbox;
+	private static Button systemTimeRadio;
+	private static Button incrementTimeRadio;
+	private static Button manualTimeRadio;
```
mx990: Why are these fields static?

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -98,12 +111,93 @@ protected Composite createEventComponent(final Composite parent) {
...
+		manualTimeRadio = new Button(comp, SWT.RADIO);
+		manualTimeRadio.setText("Manual"); //$NON-NLS-1$
+		manualTimeRadio.addSelectionListener(new SelectionAdapter() {
+			@Override
+			public void widgetSelected(final SelectionEvent e) {
+				updateLaunchConfigurationDialog();
+				setCheckbox(manualTimeRadio.getText());
+			}
+		});
+		GridDataFactory.fillDefaults().applyTo(manualTimeRadio);
```
mx990, line 136: I would suggest aligning these buttons in a column instead of in one row and adding a more descriptive label to explain what each of these modes mean, such as:
- Use system clock
- Use fixed clock with specified time
- Increment clock after each event by specified amount

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -138,6 +235,29 @@ public void performApply(final ILaunchConfigurationWorkingCopy configuration) {
...
+					FBLaunchConfigurationAttributes.MANUAL);
+		}
+		try {
+			final String debug_time = debugTimeText.getText();
+			if (debug_time != null) {
+				configuration.setAttribute(FBLaunchConfigurationAttributes.CLOCK_INTERVAL, debug_time);
+			} else {
+				configuration.setAttribute(FBLaunchConfigurationAttributes.CLOCK_INTERVAL, "0"); //$NON-NLS-1$
+			}
+		} catch (final NumberFormatException | ArithmeticException e) {
```
mx990: It seems these exceptions are never thrown here?

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
```diff
@@ -98,12 +111,93 @@ protected Composite createEventComponent(final Composite parent) {
...
+		GridDataFactory.fillDefaults().applyTo(manualTimeRadio);
+
+		final Composite c = new Composite(group, SWT.NONE);
+		GridLayoutFactory.swtDefaults().numColumns(3).applyTo(c);
+		GridDataFactory.fillDefaults().grab(true, false).applyTo(c);
+
+		debugTimeText = new Text(c, SWT.BORDER);
+		debugTimeText.setText("0"); //$NON-NLS-1$
+		debugTimeText.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1));
+		debugTimeText.addSegmentListener(e -> updateLaunchConfigurationDialog());
```
mx990, line 147: I believe this class should override `updateLaunchConfigurationDialog()` to set an appropriate error message when entering an invalid value here.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchConfigurationAttributes.java
```diff
@@ -29,6 +31,14 @@ public interface FBLaunchConfigurationAttributes extends LaunchConfigurationAttr
 	String REPEAT_EVENT = "org.eclipse.fordiac.ide.debug.repeatEvent"; //$NON-NLS-1$
 	String KEEP_RUNNING_WHEN_IDLE = "org.eclipse.fordiac.ide.debug.keepRunningWhenIdle"; //$NON-NLS-1$
 
+	String CLOCK_MODE = "org.eclipse.fordiac.ide.debug.clockMode"; //$NON-NLS-1$
+
+	String SYSTEM = "system"; //$NON-NLS-1$
+	String INCREMENT = "increment"; //$NON-NLS-1$
+	String MANUAL = "manual"; //$NON-NLS-1$
```
mx990: I would suggest defining an enum for these values and using `toString()`/`valueOf()` to convert to/from a string representation for persistence.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/FBDebugView.java
azoitl: this should defintly not be static, will fix.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/FBDebugView.java
azoitl: removed it as it is clearer to not have it.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: fxed

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: true, fixed.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchConfigurationAttributes.java
azoitl: good idea

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: it is better to override `public boolean isValid(final ILaunchConfiguration launchConfig)` but yes you are right a message for the users should be there.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: should not be. fixed.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: like that suggestion. also extracted all the strings and removed the wrong non-nls tags

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: yes. fixed.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/view/editparts/DebugTimeEditPart.java
azoitl: I guess this came from wrong static methods. fixed

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: maybe in a future version of that feature, as the commit message says it is an mvp

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
azoitl: fixed

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: done

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: done

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: what would be a fix?

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: added a null check.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: I agree, however I remember when implementing the debug view that it was very hard to get the right elements at the right point in time. So I would leave it for now.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/fb/FBLaunchConfigurationTab.java
mx990: Yes, that is understandable.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/messages.properties
```diff
@@ -5,3 +5,14 @@ MainLaunchConfigurationTab_ErrorInitializingArguments=Error initializing argumen
 MainLaunchConfigurationTab_ErrorUpdatingArguments=Error updating arguments
 MainLaunchConfigurationTab_InvalidValueMessage={0} is not a valid value for variable {1} with type {2}
 MainLaunchConfigurationTab_InvalidValueTitle=Invalid Value
+
+FBLaunchConfigurationTab_ClockInterval=Clock interval:
+FBLaunchConfigurationTab_DebugTime=Debug Time
+FBLaunchConfigurationTab_ERROR_InvalidDebugTime=Invalid value for clock interval given. Must be number greater or equal to 0.
+FBLaunchConfigurationTab_ERROR_InvalidDebugTime=
```
mx990: I believe this is duplicate?

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
```diff
@@ -14,33 +14,52 @@
...
+import org.eclipse.fordiac.ide.model.eval.AbstractEvaluator;
+import org.eclipse.fordiac.ide.model.eval.EvaluatorThreadPoolExecutor;
 import org.eclipse.fordiac.ide.model.eval.fb.FBEvaluatorCountingEventQueue;
 import org.eclipse.fordiac.ide.model.eval.fb.FBEvaluatorExternalEventQueue;
 import org.eclipse.fordiac.ide.model.libraryElement.Event;
 
 public class FBLaunchEventQueue implements FBEvaluatorCountingEventQueue, FBEvaluatorExternalEventQueue {
 	private final AtomicBoolean repeat;
 	private final AtomicBoolean blocking;
+	private final AtomicReference<FBDebugClockMode> clockMode = new AtomicReference<>();
```
mx990: I would suggest initializing this with `FBDebugClockMode.SYSTEM` here instead of setting it in the constructor.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
mx990: I believe moving the `setDebugTime()` after the `incrementEventCount` should solve the problem for now.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
mx990: Yes, I understand. I believe we should definitely refactor this at some point in the near future, though.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBDebugClockMode.java
```diff
@@ -0,0 +1,32 @@
...
+ * Contributors:
+ *   Alois Zoitl - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.debug.fb;
+
+public enum FBDebugClockMode {
+
+	SYSTEM, INCREMENT, MANUAL;
+
+	public static FBDebugClockMode fromString(final String val) {
```
mx990, line 21: Why not use `FBDebugClockMode.valueOf(val)` instead?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: done

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: done

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBLaunchEventQueue.java
azoitl: totally agree

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/messages.properties
azoitl: ah missed that created by the eternalisation wizard.

plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/fb/FBDebugClockMode.java
azoitl: was not aware about it

mx990 (approved):

PR #70, opened 2024-04-19
Crossing reconnect finds proper sources

lindhuber:
Handles source collection the same way as target collection.
- finds the real sources through subapp hierarchies
- deletes unused connections and interface elements


oberlehner:
image

Dragging from the lower expanded Subapp does not delete the old labels.

image

oberlehner (approved):

PR #48, opened 2024-04-13
add function blocks for Revolution Pi

moniwe:

PR #80, opened 2024-04-23
Newline escaping

ernstblechaPT:
In #48 stray carriage-return characters were detected in the Description of FBs.
This pull request modifies the html escaping to remove Windows-style line endings so these do not show up again.


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -882,6 +882,7 @@ protected static String fullyUnEscapeValue(final String value) {
 		escapedValue = escapedValue.replace("&apos;", "\'"); //$NON-NLS-1$ //$NON-NLS-2$
 		escapedValue = escapedValue.replace("&#10;", "\n"); //$NON-NLS-1$ //$NON-NLS-2$
 		escapedValue = escapedValue.replace("&#9;", "\t"); //$NON-NLS-1$ //$NON-NLS-2$
+		escapedValue = escapedValue.replace("\r\n", "\n"); //$NON-NLS-1$ //$NON-NLS-2$
```
azoitl, line 885: I think it is better to only check for \r because the old escaping could resulted in `\r&#10;

ernstblechaPT: at this point in the unescape-sequence `&#10;` has already be replaced back to `\n` - everything should already be back to `\r\n` here

azoitl: ah missed that. Thx.

PR #84, opened 2024-04-24
Only download most recent library

pazi146:
Download only the most recent library when dealing with a version range


ernstblechaPT (approved):
Looks good to me!

azoitl (approved):

PR #85, opened 2024-04-24
Reconnecting connections from type pins not working

azoitl:
When reconnecting connections inside of CFB and Subapp types an exception occurred. Also the border crossing check was wrong.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/InterfaceElementEditPolicy.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/InterfaceElementEditPolicy.java
@@ -17,19 +17,15 @@
 package org.eclipse.fordiac.ide.application.policies;
 
 import org.eclipse.draw2d.ConnectionAnchor;
-import org.eclipse.emf.ecore.EObject;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/InterfaceElementEditPolicy.java
```diff
@@ -81,12 +77,12 @@ private Command createReconnectCommand(final ReconnectRequest request, final boo
 		final var newPin = (IInterfaceElement) getHost().getModel();
 
 		// border crossing source reconnect
-		if (isSourceReconnect && isBorderCrossing(targetPin, newPin)) {
+		if (isSourceReconnect && isBorderCrossing(newPin, targetPin, conn.getFBNetwork())) {
```
oberlehner, line 80: are you sure that this will never be null conn.getFBNetwork() ?

azoitl: if we are reconnecting an existing connection it is currently a valid connection in our model. so if conn.getFBNetwork is null we have a major model issues.

PR #100, opened 2024-04-29
Removed Ctrl+Click open Type in Editor Feature

### LINKED ISSUE #55: Remove Ctrl+click Jump to type HpyerLink
For a long time we had the feature that while pressing ctrl and clicking on an FB Type name or clicking on a plug or socket that the according type editor is opened. While this is a nice feature to quickly navigate to a type it has several problems:

- It leads to strange behavior when using the ctrl key to select several FBs
- For plugs and sockets it is not possible to create connections with the ctrl+click feature
- User feedback (i.e., showing link underline on type names) is not really always working

As I find the first two options rather important I would propose to remove this feature. 

We have as one solution the `Open type in  editor` context menu entry that provides this feature. I would propose to add this feature also to the `Navigate` main menu. With that the feature is IMHO more discoverable and at the same time does not lead to inconsistent behavior.

@bwiesmayr @moniwe  how you see that.

@jwalt thx for bringing this to my attention and triggering this idea.

azoitl:
With this commit the feature which allowed to ctrl+click on an FBType name or on an adapter pin is removed.

Furthermore some clean-ups in the touched files where performed.

Fixes #55


bwiesmayr (approved):
code looks good

PR #101, opened 2024-04-29
Add Margin Border to Subapp Content

azoitl:
In order to have space for any hidden connection labels ad the pins of the subapp a margin border was introduced to the subapp content.


lindhuber:
Works like a charm for the left side, whereas the right seems to have almost double the amount of padding. Expansion triggers from very far away even for small preference values.

lindhuber (approved):

azoitl:
> Works like a charm for the left side, whereas the right seems to have almost double the amount of padding. Expansion triggers from very far away even for small preference values.

I would expect this is coming from the incremental growing canvas which is always growing by a fixed amount of pixels. But this was there also bevore.

PR #107, opened 2024-05-03
group expanded interfaces in case of overflow

lindhuber:
This commit introduces a mechanism that reduces the number of target pins by grouping them by their source/destination subapps as well as a stacking mechanism that behaves like the default toolbar layout if pins would be drawn out of bounds.


azoitl:
If I have many grouped target pins that lead to an overflow only the most bottom ones are grouped together, however also the higher ones need to be grouped.

azoitl (approved):

PR #114, opened 2024-05-07
do not save visible children attribute unnecessarily

bwiesmayr:
With recent changes, there was a problem that the visible children attribute was saved even when the demux was not configured. The current change tries to fix this problem by checking first whether the attribute is the default attribute - in which case the attribute is not exported. Additionally, the visible children attribute is not stored for new FBs anymore (check for isIsConfigured). Hence, also the string comparison is only done for configured FBs

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
@@ -145,7 +145,7 @@ public final class ConfigurableFBManagement {
 			if (!isInDefaultConfiguration(fb, attr.getValue(), fb.getDataType())) {
 				return ECollections.asEList(structTypeAttr.get(0), attr);
 			}
-			// Until the
+			// Until the configured state is updated automatically in the commands, save result here:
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
```diff
@@ -136,19 +136,40 @@ private static EList<Attribute> getStructManipulatorAttributes(final StructManip
...
+		final EList<Attribute> structTypeAttr = getStructManipulatorAttributes(fb);
+		if (fb.isIsConfigured()) {
+			final Attribute attr = LibraryElementFactory.eINSTANCE.createAttribute();
+			attr.setName(LibraryElementTags.DEMUX_VISIBLE_CHILDREN);
+			attr.setType(ElementaryTypes.STRING);
+			attr.setValue(buildVisibleChildrenString(fb.getMemberVars()));
+			if (!isInDefaultConfiguration(fb, attr.getValue(), fb.getDataType())) {
+				return ECollections.asEList(structTypeAttr.get(0), attr);
+			}
+			// Until the
```
azoitl: I think something is missing in the comment

azoitl (changes requested):

azoitl (approved):

PR #117, opened 2024-05-07
TypeEditor put an EContentAdapter on the whole type

azoitl:
In several places the interface editor used an EContentAdapter either on the whole type or at least on the InterfaceList. This was partly not needed or costed a lot of performance. This fix reduces the number of adapters and uses less heavy adapters where possible.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2023 Profactor GmbH, TU Wien ACIN, fortiss GmbH
+ * Copyright (c) 2011, 2024 Profactor GmbH, TU Wien ACIN, fortiss GmbH
  * 				 			Johannes Kepler University,
  * 							Primetals Technologies Austria GmbH
  *
@@ -26,6 +26,7 @@ import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.emf.common.notify.Adapter;
 import org.eclipse.emf.common.notify.Notification;
+import org.eclipse.emf.common.notify.Notifier;
 import org.eclipse.emf.ecore.util.EContentAdapter;
 import org.eclipse.fordiac.ide.fbtypeeditor.figures.FBTypeFigure;
 import org.eclipse.fordiac.ide.gef.editparts.AbstractConnectableEditPart;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
```diff
@@ -38,38 +39,34 @@
...
-					if ((null != getParent()) && (null != getFigure()) && (getFigure().isShowing())) {
-						refresh();
-					}
-				});
+			final Object feature = notification.getFeature();
+			if (notification.getEventType() == Notification.ADD
+					&& LibraryElementPackage.eINSTANCE.getLibraryElement_VersionInfo().equals(feature)) {
+				((EObject) notification.getNewValue()).eAdapters().add(this);
+				getFigure().updateVersionInfoLabel();
+			} else if (Notification.REMOVING_ADAPTER != notification.getEventType()
```
mx990: I would suggest using `!Notification.isTouch()` instead to filter for actual changes.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
```diff
@@ -38,38 +39,34 @@
...
+					&& LibraryElementPackage.eINSTANCE.getVersionInfo_Version().equals(feature)) {
```
mx990: I believe this does not cover the case where the version in an existing `VersionInfo` is changed?

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
```diff
@@ -38,38 +39,34 @@
...
-					getFigure().updateVersionInfoLabel();
-				}
-
-				Display.getDefault().syncExec(() -> {
-					if ((null != getParent()) && (null != getFigure()) && (getFigure().isShowing())) {
-						refresh();
-					}
-				});
+			final Object feature = notification.getFeature();
+			if (notification.getEventType() == Notification.ADD
```
mx990: I believe this would also need to cover the cases where a `VersionInfo` was `SET` or `REMOVE`d, as well as `ADD_MANY` and `REMOVE_MANY`.

Instead of handling all this manually, how about simply overriding `EContentAdapter.addAdapter(Notifier)` to self-add only to `LibraryElement` and `VersionInfo`?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
azoitl: according to my tests it does.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
azoitl: good point changed.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
azoitl: The main reason is that with EContentAdapter I'm never sure what is simply. But you are right doing all the cases by hand is way to cumbersome. Changed accordingly. However LibraryElement is not needed as the adapter is added to it as start of the recursion.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/FBTypeEditPart.java
mx990: Ah, my mistake. I read that wrong.

PR #121, opened 2024-05-08
improve search and update for structmanipulators

bwiesmayr:
The search of struct manipulators caused NPEs when dealing with unconfigured mux/demux and also with error markers in mux/demux. Also not all instances were found, especially in the aforementioned cases: if an unconfigured mux was encountered, the search crashed and did not necessarily find all elements. This patch also fixes the update struct command for deleting structs, because the datatype entry is still present in the library in these cases. An introduced label ensures that the datatype entry is not reloaded, but the newly created error marker is used instead.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
@@ -21,7 +21,6 @@ package org.eclipse.fordiac.ide.model.commands.change;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.model.LibraryElementTags;
 import org.eclipse.fordiac.ide.model.data.DataType;
-import org.eclipse.fordiac.ide.model.data.StructuredType;
 import org.eclipse.fordiac.ide.model.datatype.helper.IecTypes;
 import org.eclipse.fordiac.ide.model.libraryElement.Demultiplexer;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
```diff
@@ -124,19 +129,19 @@ private DataType getDataTypeFromTypeEntry() {
...
+			final DataTypeLibrary datatypeLib = entry.getTypeLibrary().getDataTypeLibrary();
+			final TypeEntry reloadedTypeEntry = datatypeLib.getDerivedTypeEntry(newStructTypeEntry.getFullTypeName());
+			if (newStructTypeEntry instanceof ErrorDataTypeEntry) {
+				if (reloadedTypeEntry != null && reloadedTypeEntry != newStructTypeEntry) {
+					// type exists now
+					type = reloadedTypeEntry.getType();
+				}
+			} else if (reloadedTypeEntry == null) {
+				// type was deleted, create error marker
+				type = datatypeLib.getType(newStructTypeEntry.getFullTypeName());
```
azoitl, line 142: as you made `createErrorMarkerType` public you could use it here now as well. This would make for me the code better readable.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeStructCommand.java
azoitl: fixed with my update

azoitl:
Rebased and fixed merge conflict.

azoitl (approved):

azoitl:
Just noticed I forgot to add my change. It is now in PR #132.

PR #123, opened 2024-05-08
Added typed subapp support for move through hierarchy

azoitl:


sebHollersbacher (commented):
The TreeNodeLabelProvider in MoveThroughHierarchyHandler does not cover the case for showing images for SubAppTypes.
We maybe should move the LabelProvider inside the Dialog class

sebHollersbacher (approved):

PR #127, opened 2024-05-14
Fix type update on library import

pazi146:


oberlehner (approved):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/librarylinker/LibraryLinker.java
```diff
@@ -238,44 +237,55 @@ public boolean removeOldLibVersion(final IFolder libDirectory) {
...
+			container.accept(visitor);
 		} catch (final CoreException e) {
 			FordiacLogHelper.logError(e.getMessage(), e);
 		}
-
 	}
 
-	@Override
-	public void updateFBInstancesWithNewTypeVersion() {
+	private void cleanupOldLibraryVersion() {
```
azoitl, line 265: in this method you are doing two things:

1. cleaning up the deleted types
2. informing the user

I think the second should be moved to its own method. Because here I would never search for it.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/librarylinker/LibraryLinker.java
pazi146: split it into two methods

azoitl: thx

azoitl (approved):

azoitl:
Could you please cherry pick your changes to freeze. So we don't get the other commits in.

azoitl:
Thx!

PR #134, opened 2024-05-16
Change value watches view

azoitl:


oberlehner (approved):

oberlehner:
Looks good now. But for future we should think if it would make sense to create a WatchValueTreeArrayNode subclass.

azoitl:
> Looks good now. But for future we should think if it would make sense to create a WatchValueTreeArrayNode subclass.

After spending quite some time with the struct parser I can definitely agree that more work and clean-up is needed there.

PR #138, opened 2024-05-21
Fix follow connection

MandKastner:


azoitl:
Can you please split this PR into two independent PRs. As they should be independently discussed.

PR #145, opened 2024-05-24
Double Click on HiddenConnectionLabel jumps to opposite pin

azoitl:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
```diff
@@ -322,6 +328,36 @@ public void activate() {
...
+			}
+		}
+		super.performRequest(request);
+	}
+
+	private void followConnection(final EditPart targetEP) {
+		final EditPartViewer viewer = getViewer();
+		HandlerHelper.selectEditPart(viewer, targetEP);
+
+		viewer.select(targetEP);
```
MandKastner: do you need that here? i think the selectEditPart() before already does that

MandKastner (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
azoitl: thx for noticing. Removed it in my last commit

PR #150, opened 2024-05-28
Hiding InOut pin of the not connected interface element

PStmmr:
When an inout pin is connected with it's input/output, the opposite pin
will be visually removed.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
@@ -327,7 +327,6 @@ public abstract class AbstractFBNElementEditPart extends AbstractPositionableEle
 			return getTargetInputFigure(interfaceEditPart);
 		}
 		return getTargetOutputFigure(interfaceEditPart);
-
 	}
 
 	private IFigure getTargetInputFigure(final InterfaceEditPart interfaceEditPart) {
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
```diff
@@ -256,6 +256,8 @@ protected void createEditPolicies() {
 
 	private HiddenPinIndicator inputPinIndicator;
 	private HiddenPinIndicator outputPinIndicator;
+	private static List<VarDeclaration> inoutInConList;
+	private static List<VarDeclaration> inoutOutConList;
```
oberlehner: why is this static?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
```diff
@@ -388,7 +390,9 @@ private static int getInterfaceInputElementIndex(final InterfaceEditPart interfa
 		}
 		if (interfaceEditPart.isVariable()) {
 			if (((VarDeclaration) interfaceEditPart.getModel()).isInOutVar()) {
-				return interfaceList.getInOutVars().indexOf(interfaceEditPart.getModel());
+				final List<VarDeclaration> visInOutList = interfaceList.getInOutVars().stream()
```
oberlehner, line 406: It would be better to make a method for that. 
I think you are using that multiple times.

please don't name a variable x. 
it should have a meaningful name

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
```diff
@@ -460,8 +474,18 @@ protected List<Object> getModelChildren() {
 				.filter(it -> !it.isVisible()).toList();
 		final List<VarDeclaration> outputRemovalList = getModel().getInterface().getOutputVars().stream()
 				.filter(it -> !it.isVisible()).toList();
+
+		final List<VarDeclaration> inoutInConList = getRemovedInOutConnections(true);
+		final List<VarDeclaration> inoutOutConList = getRemovedInOutConnections(false);
+
+		inoutInConList.stream().forEach(vd -> vd.setValue(null));
```
oberlehner: add this to the method getRemovedInOutConnections. 
Furhtermore please rename getRemovedInOutConnections
it does not return connections, therefore it should be named getVarInOutPins

PR #168, opened 2024-06-02
Reworked StructRenameRefatoring to use new search

azoitl:
With this change the struct type rename refactoring uses the new search to find the correct usages of types. Furthermore, for dependent changes (i.e., struct in struct usage) the right type entry is used.


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/StructuredTypeMemberChange.java
```diff
@@ -62,8 +59,8 @@ public Change perform(final IProgressMonitor pm) throws CoreException {
 
 		final StructuredType structuredTypeEditable = (StructuredType) affectedStruct.getTypeEntry().getTypeEditable();
 		for (final VarDeclaration varDeclaration : structuredTypeEditable.getMemberVariables()) {
-			final String typeName = varDeclaration.getTypeName();
-			if (typeName.equals(oldName)) {
+			if ((varDeclaration.getType() != null && typeEntry == varDeclaration.getType().getTypeEntry())
+					|| (typeEntry.getFullTypeName().equals(varDeclaration.getFullTypeName()))) {
```
pazi146: varDeclaration.getFullTypeName() won't give the full name if its package is imported on that location - only TypeEntry.getFullTypeName() is guaranteed to return the fully qualified name

azoitl: ok then I think it is better to remove the or part. Will do that.

oberlehner:
Using the struct to be renamed in the interface of a simple FB, which is also used in  results in a deadlock after the rename.
before:
image


Preview (strangely the StructUser Simple has not be found in the search):
image


Afterwards everything hangs:

image

azoitl:
@oberlehner thx for finding this. I found a bug in the DataTypesearch (stupid copy paste error) and I think I solved the deadlock.

PR #184, opened 2024-06-06
Add test suite for CreateFBTType functionality

### LINKED ISSUE #154: Create FBT Test Module
Part of #153 

Test the creation of Function Block Types (FBTs) flow.

This flow ensures:
- User interactions for creating new FBTs.
- Verify the successful addition of FBTs.

ShantKhatri:
Fixes #154 

Tested functionalities:
- Create New FB Type
- Create New FB Type with existing name
- Delete FB Type


tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/CreateFBTTest/CreateFBTTest.java
bwiesmayr, line 1: In 4diac IDE, we have in each plugin a file called Messages.java, which keeps constants of UI visualized Strings. These Strings are found in the messages.properties files that are next to them. I would suggest to use these Strings wherever feasible so that small text changes do not stop the test functionality.

Andrearium:
There is a copyright header problem with the CreateFBTTest class.
As the code base has been copied, the original header cannot be changed. It is not allowed to delete names or years, only adding them is allowed.

Also, the wrong class was used as basis, in my opinion. The class New4diacProjectTest is only used for menu testing.
CreateFBTTest should inherit from Abstract4diacUITests, see documentation: https://wiki.eclipse.org/Eclipse_4diacWiki/Development/User_Interface_Quality_Improvement.
Using these methods save a lot of work with project creation, deletion, clean editing area,...

plugins/org.eclipse.fordiac.ide.product/org.eclipse.fordiac.ide.product.target
```diff
@@ -59,11 +59,13 @@
 			<unit id="org.eclipse.sirius.samples.feature.group" version="7.4.0.202403051613"/>
 			<unit id="org.eclipse.sirius.specifier.feature.group" version="7.4.0.202403051613"/>
 			<unit id="org.eclipse.sirius.specifier.properties.feature.feature.group" version="7.4.0.202403051613"/>
+
```
bwiesmayr: there appears to be a new empty line added

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/Abstract4diacUITests.java
bwiesmayr, line 1: please add your name to the copyright header, and e dit the year as 2023, 2024

plugins/org.eclipse.fordiac.ide.product/org.eclipse.fordiac.ide.product.target
```diff
@@ -59,11 +59,13 @@
...
 			<unit id="org.eclipse.swtbot.eclipse.feature.group" version="4.2.0.202405291350"/>
 			<unit id="org.eclipse.swtbot.eclipse.gef.feature.group" version="4.2.0.202405291350"/>
 			<unit id="org.eclipse.swtbot.feature.group" version="4.2.0.202405291350"/>
 			<unit id="org.eclipse.swtbot.ide.feature.group" version="4.2.0.202405291350"/>
 			<unit id="org.eclipse.xtext.sdk.feature.group" version="2.35.0.v20240528-1523"/>
+			<unit id="org.eclipse.swtbot.generator.feature.feature.group" version="0.0.0"/>
```
bwiesmayr: for the production code, we need to review whether we can add this to the repository. I'd be in favor of it because it could promote creating more tests in the future. What do you think @azoitl ?

bwiesmayr (changes requested):

plugins/org.eclipse.fordiac.ide.product/org.eclipse.fordiac.ide.product.target
azoitl: After quickly reading the intro of swtbot generator I agree that this is something very useful. As long as it is only in the target platform but not referenced in any product related feature or plugin it should not get included into the 4diac IDE package. But a test would be good to ensure that. 

Please also add the version to be used. While we are not using versions in our manifests (please correct the manifest below in that respect) we have the dedicated versions for nearly everything in our product target file. Currently the only exception are the GEF Classic features. The reason for that is that we are beta testers of the latest GEF Classic developments and with providing 0.0. we get always the latest. For things from simrel like swtbot generator I find that not as good.

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/CreateFBTTest/CreateFBTTest.java
```diff
@@ -0,0 +1,236 @@
...
+	 * are enabled correctly, if you can assign a select a parent folder, enter type
+	 * name, packages and can select type template and at each stage check the state
+	 * of buttons and the UI responses. Then it checks if the newly created type can
+	 * be found in the hierarchy tree of the system explorer.
+	 *
+	 */
+	@SuppressWarnings("static-method")
+	@Test
+	public void createNewFBTypeWithoutPackage() {
+		final SWTBotMenu fileMenu = bot.menu(FILE).menu(NEW).menu(TYPE_PROJECT).click();
```
bwiesmayr, line 50: can we use here the createFBType method that you itnroduced in Abstract4diacUITests?

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/Abstract4diacUITests.java
```diff
@@ -353,6 +373,51 @@ protected static SWTBotGefEditor goToCompositeInstanceViewer(final String fb) {
 		return editor;
 	}
 
+	/**
+	 * Creates a New Type with given name.
+	 *
+	 * @param parentName Name of the parent project.
+	 * @param typeName   Name of the new type.
+	 */
+	protected static void createFBType(final String parentName, final String typeName) {
```
bwiesmayr, line 382: what if you add an additional String parameter here so that we can select the kind of FB type that we want to create? e.g. Adapter, Composite, Struct etc. Maybe even an enum that helps us manage these Strings?

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/CreateFBTTest/CreateFBTTest.java
ShantKhatri: CreateFBType is for creating FB type where we need to, and this test method will test the behavior of UI components while creating a new FB type without a package name.

bwiesmayr (approved):

PR #188, opened 2024-06-10
fix checkTypeLibrary ant task

pazi146:
ensure error markers are generated properly


azoitl (approved):
LGTM

pazi146:
PR turns out to be unnecessary - simply use importProject beforehand in the same target

PR #195, opened 2024-06-14
Deployments show additional information

PStmmr:
Deployments show the number of resource elements and their number of connections.


azoitl (changes requested):
For larger deployments this has a very poor performance. As you only need the numbers you could simple count for the different resources in the postCommand.
if destination is not null and not empty it contains the name of the resource. The command string should be an information on what is done. With this you can increment an counter per resource. Which you can use to build the string in the connection closed command. I would also recommend to use a string builder there.

azoitl (approved):

PR #200, opened 2024-06-16
DeploymenFBEvaluator syncs time

azoitl:
With this change the DeploymentFBEvaluator tries to set the faketime of the evaluator device so that evaluated fbs in 4diac IDE and fbs in the device share the same time base.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBEvaluator.java
+++ b/plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBEvaluator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Martin Erich Jobst
+ * Copyright (c) 2023, 2024 Martin Erich Jobst
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBEvaluator.java
```diff
@@ -118,6 +120,7 @@ public void evaluate(final Event event) throws EvaluatorException {
 		try {
 			writeVariables(getType().getInterfaceList().getInputVars());
 			writeVariables(getType().getInterfaceList().getInOutVars());
+			sharedState.writeDeviceParameter(FAKE_TIME_DEV_PARAM_NAME, StandardFunctions.NOW_MONOTONIC().toString());
```
mx990, line 123: Should we do this only if the device actually has a parameter `FakeTime`?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBEvaluator.java
azoitl: 4diac FORTE as well as 4diac IDE handles this gracefully. When we finally know that this is the way to got we can still add a check but I don't see that as very important. 

@ernstblechaPT what do you think?

ernstblechaPT: I expect the writeVariables call to take the majority of the time, so this should for now be a negliable influence on network performance - for now I would leave this as is

ernstblechaPT (approved):

plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBEvaluator.java
azoitl: Then I leave it as is and we will see how it behaves.

PR #207, opened 2024-06-18
Do not close WatchesView Cell editor on watch update

azoitl:
When watch values are updated (even with no value change) the cell editor is closed in the viewer refresh. This fix prevents this which is the only way to sensible edit values.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.monitoring/src/org/eclipse/fordiac/ide/monitoring/views/WatchesView.java
+++ b/plugins/org.eclipse.fordiac.ide.monitoring/src/org/eclipse/fordiac/ide/monitoring/views/WatchesView.java
@@ -116,7 +116,17 @@ public class WatchesView extends ViewPart implements ISelectionListener {
 		root = new Composite(parent, SWT.NONE);
 		root.setLayout(new GridLayout());
 		filteredTree = new FilteredTree(root, SWT.H_SCROLL | SWT.V_SCROLL | SWT.FULL_SELECTION, getPatternFilter(),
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.monitoring/src/org/eclipse/fordiac/ide/monitoring/views/WatchesView.java
```diff
@@ -80,6 +80,7 @@ public class WatchesView extends ViewPart implements ISelectionListener {
 	private final WatchesContentProvider provider = new WatchesContentProvider();
 	private boolean visible = false;
 	private WatchesViewSelectionFilterAction toggleSelection;
+	private boolean refreshingValues = false;
```
ernstblechaPT: This variable seems to be never used

ernstblechaPT (changes requested):
Works as expected!

plugins/org.eclipse.fordiac.ide.monitoring/src/org/eclipse/fordiac/ide/monitoring/views/WatchesView.java
azoitl: ah yes this is a left over from an experiment that didn't work will remove it

azoitl: done

bwiesmayr (approved):

PR #210, opened 2024-06-19
Correctly show varinout in typed and untyped subapps

azoitl:
With this change the interface bar of typed or untyped subapps shows on both sides var inouts.

Furthermore the filtering of varinouts now reflects the internal connection state of subapps and only for subapps.


PStmmr (commented):
I see some possible positive changes in performance, but I would like to know why getRemovedVarInOutPins is replaced into SubAppForFBNetworkEditPart.java

azoitl:
> I see some possible positive changes in performance, but I would like to know why getRemovedVarInOutPins is replaced into SubAppForFBNetworkEditPart.java

As I tried to to convey in my commit message. VarInOut Pins should only be hidden for subapp instances if they are not connected on the inside. However all the code for showing hidden connection labels is in the `AbstractFBNetworkEditPart` class. therefore haveing a basic implementation in `AbstractFBNetworkEditPart` and a subapp specific refinement in the `SubAppForFBNetworkEditPart` seemed for me the best solution.

PStmmr (approved):

PR #216, opened 2024-06-21
autolayout imrovements for expanded subapps

lindhuber:
Contains some changes to improve code and layout quality as well as the possibility to layout expanded subapps by selecting everything(ctrl+a) and calling the new menu entry.


oberlehner:
After layouting this :
image


image

I get this

image

oberlehner:
Furthermore the undo behaviour of "layout nested subapps" feels strange. pressing undo does not undo the whole layout, it just "undoes" one connection. this means i have to press multiple times undo until i can undo the layout.

azoitl:
@oberlehner is the sub-app size locked?

oberlehner:
> @oberlehner is the sub-app size locked?

no

oberlehner:
Functionilayt is given, UNDO needs to be fixed in a follow up commit

PR #217, opened 2024-06-21
Deletion of InternalFBs didn't correctly use table selection

azoitl:
When more then one FB was selected the InternalFBsSection always used the last selected element. This lead to exceptions as the same element was deleted multiple times.

Furthermore the DeleteInternalFBCommand was cleaned.


bwiesmayr (approved):
reviewed the code, looks great (didn't test it yet)

sebHollersbacher (approved):

PR #220, opened 2024-06-23
Removed SystemModel purging on dependency changes

azoitl:
The SystemModel is currently still used in to many places as identifier. Therefore purging it on any dependency change can lead to inconsistent states. This leads in the best case to not working monitoring or deployment and in the worst case to data loss.


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/SystemEntryImpl.java
```diff
@@ -71,4 +71,5 @@ protected AbstractTypeExporter getTypeExporter(final AutomationSystem type) {
 	public EClass getTypeEClass() {
 		return LibraryElementPackage.Literals.AUTOMATION_SYSTEM;
 	}
+
```
oberlehner: dont add empty lines with no change

azoitl: done

oberlehner (approved):

bwiesmayr (approved):

Issue #221, opened 2024-06-23
Add function blocks


MoritzO02:
Add FBT files that are not yet in the repo, that have already been developed. 

For example QE.fbt,

diplfranzhoepfinger:
@MartinMelikMerkumians 

this way i like it. 

eclipse-4diac/4diac-forte#174 is i my Eye not necessary, i like to have ONE Issue which connects the PR of Forte and IDE. 

@MoritzO02 pls read. 

and pls reference the eclipse-4diac/4diac-forte#175 with this Issue.

franz-hoepfinger-4diac:
@MoritzO02 this can be closed as soon as QE is finished.

franz-hoepfinger-4diac:
ah, and the C++ Stuff is also added.

azoitl:
All PRs are merged so I'm closing this. Feel free to reopen if more blocks are missing.

PR #222, opened 2024-06-23
Add QE.fbt to io-1.0.0 typelib folder

MoritzO02:
Add the QE functionblock to the io-1.0.0 typelib folder.
Opens in 4diac IDE Version: 2.0.1.202406201157

for issue #221


azoitl (changes requested):
Can you please remove the service sequences, Although they are in the template they do not make much sense for this example.

diplfranzhoepfinger:
> Can you please remove the service sequences, Although they are in the template they do not make much sense for this example.

totally Agree !

data/typelibrary/io-1.0.0/typelib/QE.fbt
```diff
@@ -0,0 +1,107 @@
...
+		<InputVars>
+			<VarDeclaration Name="QI" Type="BOOL" Comment="Event Input Qualifier"/>
+			<VarDeclaration Name="PARAMS" Type="STRING" Comment="Service Parameters"/>
+		</InputVars>
+		<OutputVars>
+			<VarDeclaration Name="QO" Type="BOOL" Comment="Event Output Qualifier"/>
+			<VarDeclaration Name="STATUS" Type="STRING" Comment="Service Status"/>
+		</OutputVars>
+	</InterfaceList>
+	<Service RightInterface="RESOURCE" LeftInterface="APPLICATION" Comment="Output service interface function block for event output data">
```
diplfranzhoepfinger: von hier

data/typelibrary/io-1.0.0/typelib/QE.fbt
```diff
@@ -0,0 +1,107 @@
...
+				<InputPrimitive Interface="APPLICATION" Event="INIT-"/>
+				<OutputPrimitive Interface="APPLICATION" Event="INITO-" Parameters="STATUS"/>
+			</ServiceTransaction>
+		</ServiceSequence>
+		<ServiceSequence Name="resource_initiated_termination" Comment="">
+			<ServiceTransaction>
+				<OutputPrimitive Interface="APPLICATION" Event="INITO-" Parameters="STATUS"/>
+			</ServiceTransaction>
+		</ServiceSequence>
+	</Service>
```
diplfranzhoepfinger: bis hier löschen!

azoitl:
@MoritzO02 @diplfranzhoepfinger I may be stupid but I don't understand what this block is even doing. What is the purpose of a QE block?

rinstrum-franz:
\> @MoritzO02 @diplfranzhoepfinger I may be stupid but I don't understand what this block is even doing. What is the purpose of a QE block?

we made it for @darrenpearson and Marek Swierzy from Rinstrum [https://www.rinstrum.com/product/r320-abs-waegeelektronik/](https://www.rinstrum.com/product/r320-abs-waegeelektronik/) , but at General it might be useful for others as well.

QE is opposite of IE.

while IE is a Event input (e.g. a Double-Click on a Button)   
QE is a Output which can tell another Device a Event.

in our case the Device was a SCALE.



you see the Buttons here: 

*   ZERO
*   TARE

these buttons attatched to 4diac would be IE. 

But we had the other way round. 

4diac was the Button. so the Scale was connected to 4diac as a Slave, and 4diac told the scale to Zero. 

this is a QE then. 

Understood ? 

i can imagine many other uscases. like trimming a Temperature Sensor to 0 and 100°C

like shooting a Photo from a Camera, 

like anything where you need to Output just the Information of a Event.

diplfranzhoepfinger:
eclipse-4diac/4diac-forte#175

MoritzO02 (commented):
@azoitl service removed from QE

azoitl:
@diplfranzhoepfinger thx for the explanation. Yes this makes a bit sense. But I'm not sure if this should then be handled with communication. As in the the end a communication is going on and no process data, for which the process interface was designed. If it makes sense as QE, then we definitely need a better implementation on 4diac FORTE side.

diplfranzhoepfinger:
> @diplfranzhoepfinger thx for the explanation. Yes this makes a bit sense. But I'm not sure if this should then be handled with communication. As in the the end a communication is going on and no process data, for which the process interface was designed. If it makes sense as QE, then we definitely need a better implementation on 4diac FORTE side.

true, The Communication which is going on is the Reading (Cyclic Reading) of the Weight (which is a Analog Input in the End, so it is a Process Data) 

the QE is not going on, it is a one-Time Event. 

and you are right, there is 1 clash in the Implementaition, i will mark it in a second.

azoitl:
Maybe a `PUBLISH_0` is the cleaner way to go then?

diplfranzhoepfinger:
> Maybe a `PUBLISH_0` is the cleaner way to go then?

just because it is a Communication ? would a ADC which is on a SPI Bus then also be a PUBLISH and SUBSCRIBE ? 

and a ADC which is CPU internal a Processinterface ? 

where can we read more basics about the 2 Processinterface and Publish/Subscribe ? What to use when ?

what i want definitely to avoid is to have the Reading of the ADC in Processinterface ( ID + mapping) and the ZERO and CAL ob Publish/Subscribe

azoitl:
I fear I don't have yet a good and clear answer for you and I also think the boundaries are fuzzy. The process interface is for me anything that is delivering process values to the application or takes it from the application. If this is done via directly connected IOs or remotely via some communication system is for me logically equal and should be transparent for the application. 

I think the reason for the discussion here is that currently I have a knot in my brain how an event would fit into this. Because so far events related to process interfaces are related to changes of their values. With your proposal this is now the first time different.

diplfranzhoepfinger:
> With your proposal this is now the first time different.

Which is not true. 

we have already the IE FB, where we interface the 
Espressif Button: [button](https://components.espressif.com/components/espressif/button/versions/3.2.2)
/ [github](https://github.com/espressif/esp-iot-solution/tree/master/components/button)

and so a Event can be 

https://github.com/espressif/esp-iot-solution/blob/dd15770d0f8c5fedea78809737f04b63d3765c83/components/button/include/iot_button.h#L28-L37

```
    BUTTON_PRESS_DOWN = 0,
    BUTTON_PRESS_UP,
    BUTTON_PRESS_REPEAT,
    BUTTON_PRESS_REPEAT_DONE,
    BUTTON_SINGLE_CLICK,
    BUTTON_DOUBLE_CLICK,
    BUTTON_MULTIPLE_CLICK,
    BUTTON_LONG_PRESS_START,
    BUTTON_LONG_PRESS_HOLD,
    BUTTON_LONG_PRESS_UP,
```

for these (except 2) we have already IO Mappings. 


so a Event can be a Change of a Input signal (when there is a IO Changing from TRUE to FALSE, but a Event can also be like a Click, a Double Click, a Long-Press, a Press, a Release and so on ... 

so you can also Map 1 Button to say 5 Events, and whatever the Customer does it releases Events on different IE Blocks. (and still you can map it to a IX Block) so in Summary you could have 5 IE and 1 IX mapped to 1 physical Button. 

this is Quite common in Building Automation, e.g. the Shelly Buttons support 4 Actions: press, long Press, double and triple Click. so a 2x2 Keypad can support 16 Actions. 

now, coming Back to QE, when the Partner Device has a Button as Input, and a Single Click there is wanted, we have to have a QE as Output, otherwise this Story would be unround. 

right ?

azoitl:
For me this is different. For the IE you have a boolean input and based on a certain pattern on raising and falling edges you derive events. Theoretically you could do this events also in your IEC 61499 application. Moving that to the IO driver is a convenience function and for me makes sense.  

The output side is now different. You explain you would mimic with the QE a button press. But a button press is a rising edged followed by a falling edge of a button. So if you want to do this your QE you would need to send a true followed shortly after by a false. Everything else is something different.

diplfranzhoepfinger:
> For me this is different. For the IE you have a boolean input and based on a certain pattern on raising and falling edges you derive events. 

true. 

> Theoretically you could do this events also in your IEC 61499 application.

also true. Only Theoretically. Practically this would be a wast of Computation Power. 

>  Moving that to the IO driver is a convenience function and for me makes sense.
> 
> The output side is now different. You explain you would mimic with the QE a button press. But a button press is a rising edged followed by a falling edge of a button. So if you want to do this your QE you would need to send a true followed shortly after by a false. Everything else is something different.

yes, you are right. 

but this controller only takes one Event. not a raising and a falling Edge. 

it needs a Command, and then does the Action.

diplfranzhoepfinger:
so @azoitl reason to invent QE is User Friendly interface. 

getting the Analog input by ID, and making the Zero and CAL by QE makes sense. 

otherwise a Man who want to interface it must learn not only the Mapping which is already quite complex but also the PUBLISH, and this will screw him up.

azoitl:
> so @azoitl reason to invent QE is User Friendly interface. 

@diplfranzhoepfinger I get that and the longer we discuss I get the reason and also how it can be implemented. 

In your case it would be this event message. In others it could be to really generate the pulse, I mentioned above, in the driver so that I as application engineer takes care of it. 

So we should think how we get that represented in a good way in 4diac FORTE. Should we discuss that there?

@diplfranzhoepfinger thx for your patience with me. But this is a major API change and I didn't want to talk this light-hearted.

diplfranzhoepfinger:
> > so @azoitl reason to invent QE is User Friendly interface.
> 
> @diplfranzhoepfinger I get that and the longer we discuss I get the reason and also how it can be implemented.

yes, needs to be wisely decided. 


> 
> In your case it would be this event message. 

> In others it could be to really generate the pulse, I mentioned above, in the driver so that I as application engineer takes care of it.


true. it could generate finde Pulse, Double Pulse etc.. same as Button, but opposite-wise. 


> 
> So we should think how we get that represented in a good way in 4diac FORTE. Should we discuss that there?
> 

we can discuss that here ? 
or do you preffer matrix for this ? 

> @diplfranzhoepfinger thx for your patience with me. But this is a major API change and I didn't want to talk this light-hearted.

definitely true. 
this is the Reason we employed @MoritzO02 as a Pull Request Manager

azoitl:
I think it is better to discuss at one of the PRs. Because then the discussion is better located and stored at the right place.

diplfranzhoepfinger:
> I think it is better to discuss at one of the PRs. Because then the discussion is better located and stored at the right place.

OK, i agree. 

so 

> So we should think how we get that represented in a good way in 4diac FORTE. Should we discuss that there?

YES

franz-hoepfinger-4diac (approved):

PR #228, opened 2024-06-25
Add copying of connections using drag&drop on split editor

sebHollersbacher:


azoitl:
You always have to take the opposite of the selected connection handle. For data connections you are not doing that and this leads to strange connected connections if I select the source handle.

azoitl (approved):

Issue #230, opened 2024-06-26
Remove service sequences from fbt files


MoritzO02:
Remove the `<Service> </Service>` block from fbt files

diplfranzhoepfinger:
@azoitl this issue can be closed, as all PRs are merged or closed.

azoitl:
@diplfranzhoepfinger thx for the feedback. Will close it.

PR #231, opened 2024-06-26
removed service from two convert FBs

MoritzO02:
for issue #230


azoitl:
this is an empty service sequence that is not wrong. I would like to keep it as it is.

PR #232, opened 2024-06-26
removed service sequences for two event FBs

MoritzO02:
for issue #230


azoitl:
This is a valid service sequence that I explicitly added these should stay.

PR #234, opened 2024-06-27
Inverted coloring of hidden labels for selections

PStmmr:
Selected labels become white in the background and colored in the foreground


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/FBNConnectionEndpointPolicy.java
```diff
@@ -120,14 +122,34 @@ public void layout() {
 			}
 		};
 		if (con.getSourceDecoration() != null) {
+			figure.add(selectonOverlayFigure(con.getSourceDecoration()));
```
azoitl, line 127: you may need to give the `figure` a stack layout so that both subfigures are drawn on top of each other.

azoitl (approved):

PR #235, opened 2024-06-29
removed the service sequences from five io files

MoritzO02:
removed for QB, QD, QL, QW, QX
for issue #230


MoritzO02:
There are also some more FBs with service sequences in the subdirectories of the io directory. And around 40 in other directories.
Should i create pull requests for those too?

azoitl (approved):

franz-hoepfinger-4diac:
> There are also some more FBs with service sequences in the subdirectories of the io directory. And around 40 in other directories. Should i create pull requests for those too?

@azoitl here was a Question from @MoritzO02  ? 

Thanks, 
Franz

azoitl:
@franz-hoepfinger-4diac: The same question was asked at PR #236. I hope answering once is sufficient.

diplfranzhoepfinger:
> @franz-hoepfinger-4diac: The same question was asked at PR #236. I hope answering once is sufficient.

of course. @MoritzO02 is quite eager in committing, i need to speed up to follow.

PR #254, opened 2024-07-04
Added package names to Dialogs

sebHollersbacher:

PR #267, opened 2024-07-10
Revert "Added package names to Dialogs"

oberlehner:
Reverts eclipse-4diac/4diac-ide#254


oberlehner:
Revert does not make any sense at this point .

PR #269, opened 2024-07-11
Fixed Manifest File for SWTBot Tests

Andrearium:


Andrearium:
The changes in the manifest file are obsolete and therefore also this PR, as a simpler solution was found in PR #270.

PR #270, opened 2024-07-11
Test suit for FB Type Basic functionality

ShantKhatri:
Fixes #157 
Tested functionalities:

- Add a connection between two pins in FB
- Remove a connection between two pins in FB
- Change data pin comment
- Change event pin comment 
- Change pin data type
- Create a new data Input pin
- Create a new data output pin
- Create a new event input pin
- Create a new event output pin

PR #274, opened 2024-07-11
Inital commit for supapplication SWTBot Tests

Andrearium:
Created class SubapplicationTests for SWTBot subapplication tests and added the first test createSubappViaMenu(). 

Also, new required variables were added to class Abstract4diacUITests class and recently added static variable names of the class were sorted alphabetically.


tests/org.eclipse.fordiac.ide.test.ui/META-INF/MANIFEST.MF
Andrearium, line 1: The changes in the manifest file are obsolete and therefore also this PR, as a simpler solution was found in PR #270.

azoitl (approved):

PR #287, opened 2024-07-15
Add call to prepare before setting inputs and inouts

ernstblechaPT:
In some cases the input values were lost when running internal FBs via the DeploymentEvaluator


ernstblechaPT:
@mx990 Calling prepre just before and then again inside evaluate feels rather heavy handed - do you have any suggestions?

mx990:
> @mx990 Calling prepre just before and then again inside evaluate feels rather heavy handed - do you have any suggestions?

The call to `prepare()` is normally performed in `BaseFBEvaluator.prepare()` for internal FB evaluators and in `StructuredTextEvaluator.createEvaluator(FBVariable)` if the evaluator (for some reason) does not exist yet. This should normally cover all cases. Could you please verify if `BaseFBEvaluator.prepare()` is called properly for the "outer" FB in your case?

PR #288, opened 2024-07-15
 Refactor ST grammar to limit syntactic predicates

mx990:
Refactor ST grammar to avoid syntactic predicates in literals and calls. This limits the necessary prediction to signed
literals only, providing better recovery and error messages. It also simplifies the model when distinguishing between single and repeat array init elements.


MartinMelikMerkumians (approved):

MartinMelikMerkumians:
Less syntactic predicates is always a good thing 👍

azoitl (approved):

PR #290, opened 2024-07-16
Added tests for Subapplication

Andrearium:


tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/SubapplicationTests.java
```diff
@@ -59,4 +64,113 @@ public void createSubappViaMenu() {
...
+		assertNotNull(createConnection(EO, EI));
+
+		// drag rectangle over to FB, therefore FB should be selected
+		final SWTBot4diacGefEditor editor = (SWTBot4diacGefEditor) bot.gefEditor(PROJECT_NAME);
+		editor.drag(250, 50, 600, 200);
+		assertDoesNotThrow(() -> editor.waitForSelectedFBEditPart());
+		List<SWTBotGefEditPart> selectedEditParts = editor.selectedEditParts();
+		assertFalse(selectedEditParts.isEmpty());
+		assertTrue(selectedEditParts.stream().filter(p -> p.part() instanceof FBEditPart)
+				.map(p -> (FBEditPart) p.part()).anyMatch(fb -> E_SWITCH_FB.equals(fb.getModel().getName())));
```
oberlehner: please make a method for a this stream. it looks similar in several occurrences.

Furthermore, i think it would then make sense to use assertEqual here or any other assertion that might fit better: 
https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html

Andrearium: Thanks for the tip. 
There is already a method in the abstract parent class that I overlooked and is now used. There is now also a method for checking whether the subapplication is selected.

Regarding assertEquals: it checks whether the FB is selected, so I think a boolean return value makes sense. However, I have extended the test by first checking how many elements are selected and for this assertEquals fits perfectly.

bwiesmayr (approved):

PR #296, opened 2024-07-17
Do not used grouped target pins for only one target

azoitl:
A grouped target pin was used even when there was only one target. This resulted in strange ui behavior.


MandKastner (approved):
you could replace get(0) with getFirst()

azoitl:
> you could replace get(0) with getFirst()

This is changed in the follow up PR #299.  So I think it does not make sense to do this in this PR.

PR #303, opened 2024-07-18
Type lib lookup case insensitive

ernstblechaPT:
TypeNames in IEC 61131-3 should be handled without considering the casing. The TypeLibrary did only allow names with identical casing.


ernstblechaPT:
Not sure about the performance implications this has...

mx990 (changes requested):
I am not so much worried about the performance impact, since this is simply the price we have to pay for having a case-insensitive language/standard. I am more worried about the potential implications of changing the keys to lowercase in those maps.

I also believe this fix is quite dangerous, since it covers only a small part of the `Map` interface, while other parts, such as `keySet()`, `values()`, their various methods, and iterators, remain untouched. Since these maps are currently publicly accessible, this opens the door for unexpected and problematic behavior when used elsewhere.

I believe we first have to restrict access to these maps, before we can change the keys to lowercase. In that case, I would also explicitly use `toLowerCase()` each time when accessing the map to make it more clear what is happening.

I am currently working on a fix that addresses those issues.

mx990:
I have created an alternate PR https://github.com/eclipse-4diac/4diac-ide/pull/304 to address the issues I mentioned above.

PR #304, opened 2024-07-18
Make type library case insensitive

mx990:
Make the type library case insensitive and also limit direct access to internal type maps.


ernstblechaPT (approved):
Yes, this is a better solution than adding a wrapper around ConcurrentHashMap

azoitl (approved):

PR #312, opened 2024-07-19
Manifest file fixes for SWTBot Tests

ShantKhatri:
- Bundle version of org.hamcrest.library is fixed in MANIFEST.MF file in test.ui for SWTBot Tests.

### DIFF
--- a/tests/org.eclipse.fordiac.ide.test.ui/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.fordiac.ide.test.ui/META-INF/MANIFEST.MF
@@ -33,4 +33,4 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.swt,
  org.hamcrest.core,
  org.eclipse.draw2d,
- org.hamcrest.library;bundle-version="[1.3.0,2.0.0)"
+ org.hamcrest.library;bundle-version="[1.3.0,2.2.1)"
--- a/tests/org.eclipse.fordiac.ide.test.ui/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.fordiac.ide.test.ui/META-INF/MANIFEST.MF
@@ -33,4 +33,4 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.swt,
  org.hamcrest.core,
  org.eclipse.draw2d,
- org.hamcrest.library;bundle-version="[1.3.0,2.2.1)"
+ org.hamcrest.library


azoitl:
@ShantKhatri I fear now I'm confused. In your original commit you had:
image

I hoped that this change, which is much less restrictive then before would be enough.

ShantKhatri:
> @ShantKhatri I fear now I'm confused. In your original commit you had: image
> 
> I hoped that this change, which is much less restrictive then before would be enough.

After considering it further, I realized that we don't need to replace org.hamcrest.core with org.hamcrest.library. Instead, it should be sufficient to simply add org.hamcrest.library without any version restrictions.

bwiesmayr (approved):

ShantKhatri:
> @ShantKhatri I fear now I'm confused. In your original commit you had: image
> 
> I hoped that this change, which is much less restrictive then before would be enough.


@azoitl I'm still confused here.
Which change is less restrictive?
The change which is in the image or the change which we have done on this(#312 ) PR.

azoitl:
> > @ShantKhatri I fear now I'm confused. In your original commit you had: image
> > I hoped that this change, which is much less restrictive then before would be enough.
> 
> @azoitl I'm still confused here. Which change is less restrictive? The change which is in the image or the change which we have done on this(#312 ) PR.

I find the one in my screenshot, which is from one of your PRs, the simplest.

But you also write this:

> After considering it further, I realized that we don't need to replace org.hamcrest.core with org.hamcrest.library. Instead, it should be sufficient to simply add org.hamcrest.library without any version restrictions.

So I think this solution seems to be best. Could you change this PR to this proposal?

ShantKhatri:
> > > @ShantKhatri I fear now I'm confused. In your original commit you had: image
> > > I hoped that this change, which is much less restrictive then before would be enough.
> > 
> > 
> > @azoitl I'm still confused here. Which change is less restrictive? The change which is in the image or the change which we have done on this(#312 ) PR.
> 
> I find the one in my screenshot, which is from one of your PRs, the simplest.
> 
> But you also write this:
> 
> > After considering it further, I realized that we don't need to replace org.hamcrest.core with org.hamcrest.library. Instead, it should be sufficient to simply add org.hamcrest.library without any version restrictions.
> 
> So I think this solution seems to be best. Could you change this PR to this proposal?

Thanks! I'll change it.

azoitl (approved):
thanks for changing it

PR #313, opened 2024-07-19
Added more Zoom Levels

azoitl:
The preconfigured zoom levels are also used for scroll wheelezooming. In order to give users more control more levels where introduced between 45% and 75% as well as 100% and 200%. Furthermore the levels above 200% where removed as they where not adding any values.


bwiesmayr (approved):

oberlehner:
Why do you removed the values above 200% ? They are working and maybe some people are using it. Or do they have any negative impact on performance if they are activated ?

azoitl:
> Why do you removed the values above 200% ? They are working and maybe some people are using it. Or do they have any negative impact on performance if they are activated ?

A just saw you allready merged the PR. One of the main reasons is that larger then 200% adds no value and on most screen even leads to confusing drawings. Where people have to scroll out first to orient themselves. Reducing maximum zoom can help here to reduce the number of user interactions needed.

PR #319, opened 2024-07-22
Extend test application generation framework

WinterMelanie:
In the past, a test application framework was developed that generates a set of FBs based on a test specification that is provided as a service sequence. The initial implementation had a few problems, which are fixed in this PR, e.g.:

- reduced code duplication by introducing helper methods
- fixed a bug where the number of states was not counted correctly, leading to incorrect ECCs
- fixed bugs for matching the expected result with the actual result of a test execution
- fixed bugs when running all tests with a single event by adding additional timers


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/testappgen/CompositeTestFBGenerator.java
```diff
@@ -281,6 +281,13 @@ private void createConnToTestFB(final int index) {
 		compositeFB.getFBNetwork().getEventConnections()
 				.add(createEventConnection(runAllFB.getInterface().getEventOutputs().get(index),
```
bwiesmayr: this is a code duplication making it hard to read. better create an EList of connections and then addAll, then also you can tell for each event connection what they are for

final EList<Connection> conns = new BasicEList<>();
Connection whatIsItAbout = ...
compositeFB.getFBNetwork().getEventConnections().addAll(conns);


also put things you need very often into helper variables, such as compositeFB.getInterfaceList().getEventInputs(), so that the code is better readable. This can be done in many many places.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/testappgen/TestsignalFBGenerator.java
```diff
@@ -41,7 +44,15 @@ public BasicFBType generateTestFb() {
 	// each testcase needs an event that starts it
 	@Override
 	protected List<Event> createInputEventList() {
-		return testSuite.getTestCases().stream().map(n -> createEvent(n.getName() + "_TEST", true)).toList(); //$NON-NLS-1$
+		// return testSuite.getTestCases().stream().map(n -> createEvent(n.getName() +
```
bwiesmayr: there is a comment here

bwiesmayr (approved):

PR #321, opened 2024-07-23
Add ISOBUS TECU communication layer and related functionality. Add First Block. 

franz-hoepfinger-4diac:
add the Module,
and add the First of many Blocks.

belongs to https://github.com/eclipse-4diac/4diac-forte/issues/188


franz-hoepfinger-4diac:
to support this: 
here a Example of how to use: 
image

azoitl:
given the current situation I think this should be an external library.

franz-ms-muc:
> given the current situation 

what Situation ? 

> I think this should be an external library.

I agree ! 
the only missing Link is Project Specific Setting for the "External Libraries" so we can distribute Projects over Github as Open Source and still use External Library.

azoitl:
> > given the current situation
> 
> what Situation ?

The license situation

PR #323, opened 2024-07-23
Fixed NPE by executing repair Operation

oberlehner:
Repairing a missing type error marker from a struct member produced a NPE

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/RepairCommandHandler.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/RepairCommandHandler.java
@@ -141,7 +141,9 @@ public class RepairCommandHandler extends AbstractHandler {
 					handleChangeDataType(choice);
 					break;
 				case CREATE_MISSING_TYPE:
-					ErrorMarkerResolver.repairMissingStructuredDataType(type);
+					if (type instanceof final VarDeclaration varDeclaration) {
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/RepairCommandHandler.java
```diff
@@ -141,7 +141,9 @@ public boolean performFinish() {
 					handleChangeDataType(choice);
 					break;
 				case CREATE_MISSING_TYPE:
-					ErrorMarkerResolver.repairMissingStructuredDataType(type);
+					if (type instanceof final VarDeclaration varDeclaration) {
+						ErrorMarkerResolver.repairMissingStructuredDataType(varDeclaration);
+					}
```
azoitl, line 146: should there be some info to the user in the else?

Gatsch (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/RepairCommandHandler.java
oberlehner: i think we should not even be able to open the dialog at this point if the user selects an error marker which is not a "Missing DT/FBT".  
This is verified in the repairProperty tester.
But in general the repairCommandHandler needs a clean-up.

PR #324, opened 2024-07-23
Fixed detaching of SubApp from Type with no open editor

sebHollersbacher:


azoitl (changes requested):
I fear this fixes the wrong problem. It should not be allowed that the SubApp is detached from the type in the system explorer. Can please change that such that the handler is not active when the subapp is selected in the system explorer tree.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/UntypeSubApplicationHandler.java
```diff
@@ -55,22 +56,25 @@ public void setEnabled(final Object evaluationContext) {
...
 		}
 		return null;
 	}
 
 	private static SubApp getSubApp(final Object currentElement) {
-		if (currentElement instanceof SubApp) {
-			return (SubApp) currentElement;
+		if (currentElement instanceof final SubApp subApp) {
+			return subApp;
+		}
```
azoitl: should we remove the subapp check? because currently an editor selection always involves an editpart

azoitl (approved):

PR #325, opened 2024-07-23
Avoid possible NPE when accessing e.getName()

ernstblechaPT:
FordiacErrorMarkerInterfaceHelper assumed that every ErrorMarker will have a non-null value returned via getName()


mx990:
> FordiacErrorMarkerInterfaceHelper assumed that every ErrorMarker will have a non-null value returned via getName()

I believe that should normally be a reasonable assumption. It seems to me the problem is that we create an error marker interface pin with a `null` name in the first place.

I believe we should rather catch such cases where the error marker is created and deal with the problem there properly (e.g., an error during import). Otherwise, we likely risk similar problems in other places down the line.

oberlehner:
I think we should find the root cause for the NPE . I don't think there should be a null name anywhere.

PR #326, opened 2024-07-23
Remove Double Escaping of Quotes in CFB Export #320

### LINKED ISSUE #320: Invalid WSTRING when exporting a CFB
The exporter adds `\$"`  at the beginning and at the end of a WSTRING FB data port when the FB is located inside a Composite FB.

- expected WSTRING ID:

image

- actual WSTRING ID:

image

- cpp file of the exported CFB:

image

azoitl:
`"` in values of CFB member fbs where double escaped during the C++ export. This led to wrong values being used in the 4diac FORTE.

Fixes #320


m-meingast (approved):
Works as expected

PR #328, opened 2024-07-24
Improve logic for project restoration

pazi146:
also start using Version and VersionRange instead of String

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
@@ -25,7 +25,6 @@ import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
 import java.text.MessageFormat;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -596,72 +599,84 @@ public void checkManifestFile(final IProject project, final TypeLibrary typeLibr
...
+	private void checkDependency(final Required req, final IProject project, final TypeLibrary typeLibrary) {
 		checkLibChanges();
+		String preferred = null;
+
 		// check already linked lib
-		if (projectLibs.containsKey(req.getSymbolicName()) && (projectLibs.get(req.getSymbolicName()).stream()
-				.filter(p -> versionComparator.contains(req.getVersion(), p)).count() > 0)) {
-			return;
+		final IFolder libFolder = project.getFolder(TYPE_LIB_FOLDER_NAME).getFolder(req.getSymbolicName());
+		if (libFolder.exists()) {
```
oberlehner, line 627: please try to flatten the if/else hierachy. This is hard to read

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -596,72 +599,84 @@ public void checkManifestFile(final IProject project, final TypeLibrary typeLibr
...
+			if (checkLibraries(req, project, typeLibrary, stdlibraries)) {
+				return;
+			}
 
-			startLocalLinkJob(req, project, libraries, typeLibrary);
-			return;
+			// check local libs
+			if (checkLibraries(req, project, typeLibrary, libraries)) {
+				return;
+			}
```
oberlehner, line 667: This code is hard to read.
For example you can make methods, checkStandardLib, checkLocalLib, as you use that multiple times

PR #329, opened 2024-07-24
Instance comments are now saved for Event pins

PStmmr:
A check was changed in the ElementExporter, which now also allows event comments

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
@@ -52,7 +52,6 @@ import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
 import org.eclipse.fordiac.ide.model.libraryElement.AttributeDeclaration;
 import org.eclipse.fordiac.ide.model.libraryElement.ColorizableElement;
 import org.eclipse.fordiac.ide.model.libraryElement.ErrorMarkerInterface;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
```diff
@@ -464,8 +465,9 @@ private void addParam(final IInterfaceElement ie) throws XMLStreamException {
 		final boolean hasAttributes = hasNonTrivialAttributes(ie);
 		final boolean hasInitalValue = (ie instanceof final VarDeclaration varDecl) && (varDecl.getValue() != null
 				&& varDecl.getValue().getValue() != null && !varDecl.getValue().getValue().isBlank());
-		final boolean hasComment = ie instanceof final VarDeclaration varDecl && varDecl.getComment() != null
-				&& !varDecl.getComment().equals(""); //$NON-NLS-1$
+		final boolean hasComment = (ie instanceof final VarDeclaration varDecl && varDecl.getComment() != null
```
bwiesmayr: is it enough here for you to ask for IInterfaceElement and not duplicate the check? this also includes e.g. adapters but they also have comments so that should not be a problem?

azoitl: As all interface elements may have comments I find that a good idea.

bwiesmayr (approved):
technically, this could be simplified to "".equals(ie.getComment()), as there is no null check required anymore, but I'd take it

PR #331, opened 2024-07-25
Fixed setting copy/cut/paste actions in FBNetworkEditor

sebHollersbacher:
When multiple editors were open, switching between them cause the old editor to delete the actions of the new one


azoitl (approved):
It looks like that this fix, also fixes an issue that I'm experiencing under Linux when ending 4diac IDE with open editors.

PR #339, opened 2024-07-29
Add new event-driven bistable FBType E_T_FF_SR.

franz-hoepfinger-4diac:
E_T_FF_SR is a Mix between E_SR and E_T_FF

sometimes a E_T_FF lacks a Reset. 
so in some Application you want a T_FF, but need a definitive OFF Input. 

this Block fullfills this.

### DIFF
--- a/data/typelibrary/events-1.0.0/typelib/E_RS.fbt
+++ b/data/typelibrary/events-1.0.0/typelib/E_RS.fbt
@@ -1,41 +1,44 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!DOCTYPE FBType SYSTEM "http://www.holobloc.com/xml/LibraryElement.dtd">
-<FBType Comment="Event-driven bistable" Name="E_RS">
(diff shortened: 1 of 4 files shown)


franz-hoepfinger-4diac:
Structure similar to:
https://eclipse.dev/4diac/en_help.php?helppage=html/4diacIDE/use4diacLocally.html

azoitl (commented):
Wouldn't an implementation as Basic FB be cleaner?

diplfranzhoepfinger:
> Wouldn't an implementation as Basic FB be cleaner?

I agree, yes. 

would be cleaner ! 

in the Documentation we can still say this is same as the Tutorial, for interested people a nice mind-task. 

for non-interested people it must just work, so does not matter.

azoitl:
> > Wouldn't an implementation as Basic FB be cleaner?
> 
> I agree, yes. 
> 
> would be cleaner ! 
> 
> in the Documentation we can still say this is same as the Tutorial, for interested people a nice mind-task. 
> 
> for non-interested people it must just work, so does not matter.

so?

franz-hoepfinger-4diac:
> > I agree, yes.

> so?

i will design and make a Block.

azoitl:
Great!

franz-hoepfinger-4diac:
see now ...

azoitl (approved):

PR #342, opened 2024-07-30
Make version range conflict message more explicit

pazi146:


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/messages.properties
```diff
@@ -16,7 +16,7 @@ Confirm=Confirm
 
 ErrorMarkerLibNotAvailable=Library not available: {0} Version: {1}
```
oberlehner: can you add the path to the manifest that caused the error.
this is confusing otherwise, especially when the error is caused by a transitive import

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/messages.properties
```diff
@@ -16,7 +16,7 @@ Confirm=Confirm
...
 ErrorMarkerStandardLibNotAvailable=Standard library not available: {0} Version: {1}
```
oberlehner: same here

oberlehner:
For example:

This error marker:
image

has the root cause in lib08

image

oberlehner:
So a i would expect a message like  :
The depeendcy  **xyz** 1.0.0  declared in  lib08/manifest.mf is not available

pazi146:
added a used by clause - multiple libraries might have the same dependency
(also the linked manifest files of libraries aren't visible in the IDE)

PR #360, opened 2024-08-06
fix behavior of inner typed subapps

MandKastner:


diplfranzhoepfinger:
wow !

diplfranzhoepfinger:
there is a Problem with Composite FBs. 

`    forte::core::CInternalFB<FORTE_IX> fb_IX;`


it then does not find the 

`#include "IX_fbt.h"`

and if you specifiy the Path in CMakeList.txt, 

then it does not find the 

`#include "processinterface.h"`

and if you specifiy this Path  as well in CMakeList.txt, 

then it does not link. 

maybe it is a bit special to have a IX in a Composite FB, but maybe this points to a special Error, which should be solved anyhow. 

thanks, 
Franz

MandKastner:
@diplfranzhoepfinger thank you for mentioning, i will take a look at it

plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/data/ResourceDeploymentData.java
```diff
@@ -137,18 +139,72 @@ public static String getMappedParentName(final FBNetworkElement fbnElement) {
...
-			final Function<VarDeclaration, String> dataInputValue = DeploymentHelper
-					.getVariableValueRetargetable(dataInput, subApp);
-			if (dataInputValue != null) {
-				for (final ConDeploymentDest destData : getSubappInterfaceconnections(subAppHierarchy, prefix,
-						dataInput)) {
-					final VarDeclaration destVar = (VarDeclaration) destData.destination;
-					final String destDataValue = dataInputValue.apply(destVar);
-					params.add(new ParameterData(destDataValue, destData.prefix, destVar));
+
+			if (useSubAppInitialValue(subAppHierarchy, subApp, dataInput)) {
```
azoitl: as the useSubAppInitialValue is an expensive operation I would only call it when necessary. Therefore I would first perform the check `if(dataInputValue != null && useSubappInitialValue(..))` 
Otherwise deployment may be tremendously slower.

plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/data/ResourceDeploymentData.java
```diff
@@ -137,18 +139,72 @@ public static String getMappedParentName(final FBNetworkElement fbnElement) {
...
+				}
+			}
+		}
+	}
+
+	private static boolean useSubAppInitialValue(final Deque<SubApp> subAppHierachy, final SubApp subApp,
+			final VarDeclaration varDec) {
+		final IInterfaceElement ie = subApp.getInterfaceElement(varDec.getName());
+		if (!hasInputConnections(ie)) {
+			return DeploymentHelper.hasTypeOrInstanceInitialValue(subApp, varDec);
```
azoitl: I'm not sure if hsyTypeirInstanceInitalValue is doing what it should do here. We need to check if either the instance of the type has an value. The method checks if the pin has a value or if the subapp is typed. This is something different.

azoitl (changes requested):

azoitl (approved):

azoitl:
@diplfranzhoepfinger can you please test this also for your cases. It will be in the nightly build in app 1h.

franz-hoepfinger-4diac:
ah, there is a Change in IDE too ? Forte Side it is compiling, 
but i dit not Test-Run it yet. 

i will do so ASAP.

PR #361, opened 2024-08-06
Correct Request IDs in bootfile

MandKastner:
I think the bootfile containes wrong IDs: before this change they started with ID=2 and the last ID appeared twice. 

image


azoitl:
Unfortunately this his not fixing the problem. It just hides that `DeplyomentExecutor` is using in several places `id` directly instead of using `getNextId()`.

kumajaya:
> Unfortunately this his not fixing the problem. It just hides that `DeplyomentExecutor` is using in several places `id` directly instead of using `getNextId()`.

Sorry, I can't confirm this PR as I didn't build the 4diac IDE from source but the issue does exist.

oberlehner:
> > Unfortunately this his not fixing the problem. It just hides that `DeplyomentExecutor` is using in several places `id` directly instead of using `getNextId()`.
> 
> Sorry, I can't confirm this PR as I didn't build the 4diac IDE from source but the issue does exist.

If you want to test it, you first have to run 4diac IDE from source -> https://github.com/eclipse-4diac/4diac-documentation/blob/main/src/development/building4diac.adoc

After that you need to check out this PR and start 4diac in eclipse.

azoitl (approved):

PR #379, opened 2024-08-12
make multiplexer configurable

bwiesmayr:
Multiplexers are now configurable. This means that pins can be hidden. Import from XML does not yet work, but edits within the IDE are possible.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/META-INF/MANIFEST.MF
+++ b/plugins/org.eclipse.fordiac.ide.application/META-INF/MANIFEST.MF
@@ -26,7 +26,8 @@ Require-Bundle: org.eclipse.emf.ecore.editor,
  org.eclipse.fordiac.ide.model.ui,
  org.eclipse.core.expressions,
  org.eclipse.fordiac.ide.ui.errormessages,
- org.eclipse.fordiac.ide.model.search
+ org.eclipse.fordiac.ide.model.search,
+ org.eclipse.jface
 Bundle-Vendor: Eclipse 4diac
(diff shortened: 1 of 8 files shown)


azoitl:
With your change pins would be completely gone. I'm not sure if this is what we would like to have for struct muxers. Maybe I would like to provide constants to some struct values and hide that from the ui. This would not be possible with your change or?

diplfranzhoepfinger:
crazy. I might be interested in this, but cannot imagine how it looks.

azoitl (approved):

diplfranzhoepfinger:
will be in the Next nightly ?

azoitl:
As always when it is merged into develop it will be in the next nightly build.

diplfranzhoepfinger:
4diac IDE

Version: 3.0.0

Build id: 2024-08-27\_0544

testing: 

Backward compatibility is not given. 

other than that it works great.

bwiesmayr:
> 4diac IDE
> 
> Version: 3.0.0
> 
> Build id: 2024-08-27_0544
> 
> testing:
> 
> Backward compatibility is not given. 
> 
> other than that it works great.

May I ask what exactly was not backwards compatible?

diplfranzhoepfinger:
a SUB with a Multiplexer which was configured shows up red. 

then you need to select another Struct, and switch back to the original struct, then the red is gone. 

i guess the XML Read-in is not converting it to the new Format.

bwiesmayr:
Hm, I did not change the format how to store the reference to a structured type itself, so I am a bit surprised. Thanks for reporting

Issue #162, opened 2024-05-29
Interaction Tab Test Module


ShantKhatri:
Part of #153 

Test the UI changes within the "Interaction" tab actions.

This flow includes sub-modules/tasks:

- [ ]  Type Info
- [ ]  Event In- & Outputs
- [ ]  Var In- & Outputs
- [ ]  Var InOuts
- [ ]  Plugs & Sockets
- [ ]  Var internals
- [ ]  Constants
- [ ]  Functional Blocks
- [ ]  Attributes tabs

ShantKhatri:
fixed by #381

PR #355, opened 2024-08-04
Added Test Suit for DataType Editor

ShantKhatri:
part of #153 

Test suites for Data Type Editor, these changes test the different scenarios of the data type table. 
Scenarios Include:

- Rename variable
- Change variable name with button
- Change name of variable with existing name
- Set invalid name
- Reorder variable
- Change datatype of variable
- Set invalid datatype
- Change comment of the variable
- Change initial value of the variable
- Set invalid initial value

PR #381, opened 2024-08-13
Properties Sheet's Tabs Tests

ShantKhatri:
part of #162 

This includes Test suites for 

- Attributes Tab
- Function Blocks Tab
- Constants Tab
- Event In-Outputs Tab
- Var In&Outputs Tab
- Var Internal Tab


ShantKhatri:
It has a dependency on #355.

ShantKhatri:
Hey  @bwiesmayr , whenever you get a chance, could you please take a look at my PR? Thanks a lot!

bwiesmayr:
> Hey @bwiesmayr , whenever you get a chance, could you please take a look at my PR? Thanks a lot!

I'm already half-way through it, but the PRs are quite big, that takes its time

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/AttributesTabTests.java
```diff
@@ -0,0 +1,71 @@
...
+
+	@Override
+	public void tryToSetInValidName() {
+		// Here Name column has same behavior as Data Type column or Invalid value
+		NatTableHelper.setInvalidDataType(natTableBot, 1, 1, UITestNamesHelper.IF);
+	}
+
+	@Override
+	public void tryToChangeNameOfVariableWithExistingName() {
+		// Here Name column has same behavior as Data Type column or Invalid value
```
bwiesmayr, line 67: shouldn't we still test that this is the case? We might change the code in the future, tests should not be based on the current implementation

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/NatTableHelper.java
```diff
@@ -0,0 +1,241 @@
...
+
+	/**
+	 * Changes the value of a specific cell in the NatTable.
+	 *
+	 * @param dataTypeTableBot NatTable containing the targeted cell.
+	 * @param newValue         The new value to set in the cell.
+	 * @param row              The row number of the cell.
+	 * @param col              The column number of the cell.
+	 */
+	protected static void changeCellValueInNatTbale(final SWTBot4diacNatTable dataTypeTableBot, final String newValue,
```
bwiesmayr, line 50: Table

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/NatTableWithEditorBehaviorTests.java
```diff
@@ -0,0 +1,213 @@
...
+	 *
+	 * This method verifies that the event name can be changed successfully and
+	 * checks that the old name is no longer present in the editor, while the new
+	 * name is.
+	 */
+	@SuppressWarnings("static-method")
+	@Test
+	public void changeEventName() {
+		assertNotNull(editor.getEditPart(TESTVAR1));
+		NatTableHelper.changeCellValueInNatTbale(natTableBot, UITestNamesHelper.TESTVAR, 3, 1);
```
bwiesmayr, line 107: NatTable

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/VarInternalsTabTests.java
```diff
@@ -0,0 +1,61 @@
...
+	 * Tests changing the retain tag of a variable in the DataType Editor Table.
+	 *
+	 * This method checks whether the retain tag of a variable can be modified
+	 * successfully. It changes the retain tag of the specified variable and
+	 * verifies the operation.
+	 */
+	@SuppressWarnings("static-method")
+	@Test
+	public void changeRetainTag() {
+		NatTableHelper.setRetainValue(natTableBot, 1, 5, RetainTag.RETAIN);
```
bwiesmayr, line 58: shouldn't we also try unsetting a value. Because maybe setting works, but resetting does not

bwiesmayr: Also I question a bit whether we are testing the right things here. We only know at this point that the NAT table operation works correctly, which tests also the correct setup of the NAT table. Are we able to also check that the value was correctly set up in the underlying model? One idea could be to inbetween switch the tab before checking the value. This would reload the information from the internal model and make sure that the values are actually saved in the background.

bwiesmayr (approved):
I'd have some questions and comments, but we can also fix those in a separate commit

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/VarInternalsTabTests.java
ShantKhatri: > shouldn't we also try unsetting a value. Because maybe setting works, but resetting does not

Thanks for pointing, will add the resetting tests in the next PR.

ShantKhatri: > Also I question a bit whether we are testing the right things here. We only know at this point that the NAT table operation works correctly, which tests also the correct setup of the NAT table. Are we able to also check that the value was correctly set up in the underlying model? One idea could be to inbetween switch the tab before checking the value. This would reload the information from the internal model and make sure that the values are actually saved in the background.

Sure, will try to update these tests as per your suggestion, it is a good idea to switch the tab in between so that we can have confirmation for the value are actually stored or not.

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/NatTableWithEditorBehaviorTests.java
ShantKhatri: I'll update in new PR.

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/NatTableHelper.java
ShantKhatri: I'll update in new PR.

tests/org.eclipse.fordiac.ide.test.ui/src/org/eclipse/fordiac/ide/test/ui/fbtype/AttributesTabTests.java
ShantKhatri: Here, I found the behavior of both the columns as the same, I'll update this by separate implementation for testing the scenario when a user tries to assign the existing name.

PR #405, opened 2024-08-21
Added Quick-Fix option for missing FB type

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/AbstractFBErrorMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Sebastian Hollersbacher
+ *    - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.application.error_resolution;
```
mx990: Please follow the Eclipse Naming Convention for package names, particularly https://wiki.eclipse.org/Naming_Conventions#Additional_rules:
> Package names should contain only lowercase ASCII alphanumerics, and **avoid underscore** _ or dollar sign $ characters.

```suggestion
package org.eclipse.fordiac.ide.application.marker.resolution;
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,67 @@
...
+	private TypeEntry newEntry;
+
+	public ChangeFBMarkerResolution(final ErrorMarkerFBNElement errorType) {
+		super(FordiacMessages.Repair_Dialog_ChangeFBType, errorType);
+	}
+
+	@Override
+	public void run(final IMarker[] markers, final IProgressMonitor monitor) {
+		final DataTypeTreeSelectionDialog dialog = new DataTypeTreeSelectionDialog(PlatformUI.getWorkbench()
+				.getActiveWorkbenchWindow().getActivePage().getActivePart().getSite().getShell(),
```
mx990:
```suggestion
		final DataTypeTreeSelectionDialog dialog = new DataTypeTreeSelectionDialog(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/CustomMarkerResolutionGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		return new IMarkerResolution[0];
+	}
+
+	@Override
+	public boolean hasResolutions(final IMarker marker) {
+		try {
+			final EObject target = FordiacErrorMarker.getTarget(marker);
+			return target instanceof ErrorMarkerFBNElement;
```
mx990: This will potentially load all files with markers in it when updating the *Problems* view. I would again suggest to rely on the target type attribute.
```suggestion
			final EClass targetType = FordiacErrorMarker.getTargetType(marker);
			return LibraryElementPackage.Literals.ERROR_MARKER_FBN_ELEMENT.equals(targetType);
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/CustomMarkerResolutionGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+public class CustomMarkerResolutionGenerator implements IMarkerResolutionGenerator2 {
+
+	@Override
+	public IMarkerResolution[] getResolutions(final IMarker marker) {
+		try {
+			final EObject obj = FordiacErrorMarker.getTarget(marker);
+			if (obj instanceof final ErrorMarkerFBNElement errorType) {
+				return new IMarkerResolution[] { new CreateMissingFBMarkerResolution(errorType),
+						new ChangeFBMarkerResolution(errorType) };
+			}
```
mx990: The `FordiacErrorMarker.getTarget(marker)` will load the file if necessary, which can be a considerable draw of resources and impact overall performance.

I would suggest to rely on the target type attribute instead, pass the marker to the resolutions and use `FordiacErrorMarker.getTarget(marker)` only when actually fixing the marker (or looking for other markers).
```suggestion
			final EClass targetType = FordiacErrorMarker.getTargetType(marker);
			if (LibraryElementPackage.Literals.ERROR_MARKER_FBN_ELEMENT.equals(targetType)) {
				return new IMarkerResolution[] { new CreateMissingFBMarkerResolution(marker),
						new ChangeFBMarkerResolution(marker) };
			}
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,67 @@
...
+public class ChangeFBMarkerResolution extends AbstractFBErrorMarkerResolution {
+
+	private TypeEntry newEntry;
+
+	public ChangeFBMarkerResolution(final ErrorMarkerFBNElement errorType) {
+		super(FordiacMessages.Repair_Dialog_ChangeFBType, errorType);
+	}
+
+	@Override
+	public void run(final IMarker[] markers, final IProgressMonitor monitor) {
```
mx990: There may be cases when only `run(IMarker)` is called. I would suggest to perform any preparatory steps on the first call of `run(IMarker)` (e.g., when `newEntry` is `null`) instead of overriding `run(IMarker[], IProgressMonitor)`.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/CreateMissingFBMarkerResolution.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.fordiac.ide.ui.FordiacMessages;
+
+public class CreateMissingFBMarkerResolution extends AbstractFBErrorMarkerResolution {
+
+	public CreateMissingFBMarkerResolution(final ErrorMarkerFBNElement errorType) {
+		super(FordiacMessages.Repair_Dialog_New_FB, errorType);
+	}
+
+	@Override
+	public void run(final IMarker[] markers, final IProgressMonitor monitor) {
```
mx990: same here regarding `run(IMarker)` vs. `run(IMarker[], IProgressMonitor)`

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/error_resolution/CustomMarkerResolutionGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.eclipse.core.resources.IMarker;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
+import org.eclipse.fordiac.ide.model.libraryElement.ErrorMarkerFBNElement;
+import org.eclipse.ui.IMarkerResolution;
+import org.eclipse.ui.IMarkerResolutionGenerator2;
+
+public class CustomMarkerResolutionGenerator implements IMarkerResolutionGenerator2 {
```
mx990: Please use a more descriptive name, such as `FordiacMarkerResolutionGenerator`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/wizards/RestrictedNewTypeWizard.java
```diff
@@ -0,0 +1,104 @@
...
+	private RestrictedNewTypeWizard() {
+	}
+
+	@Override
+	protected NewFBTypeWizardPage createNewFBTypeWizardPage() {
+		return new NewFBTypeWizardPage(new StructuredSelection());
+	}
+
+	public static TypeEntry showRestrictedNewTypeWizard(final String name, final String fileEnding,
+			final String projectName) {
```
mx990: I would suggest to pass the `Shell` as a parameter to facilitate future cases where a `Shell` is already available.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/wizards/RestrictedNewTypeWizard.java
```diff
@@ -0,0 +1,104 @@
...
+		}
+		for (final Control c : root.getChildren()) {
+			if (c instanceof final Composite com) {
+				flattenedSet.addAll(getFlattenedList(com));
+			} else {
+				flattenedSet.add(c);
+			}
+		}
+		return flattenedSet;
+	}
```
mx990, line 104: This reaches *very* far into the internals of the wizard page up into the `TreeViewer` itself and thus seems quite messy and fragile. It looks like there is no straightforward way to filter containers in the existing wizard page base class `WizardNewFileCreationPage`.

I believe we may either have to rely on validation here or write an entirely custom wizard page, similar to `org.eclipse.jdt.ui.wizards.NewClassWizardPage`. What do you think @oberlehner @azoitl?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractDataTypeErrorMarkerResolution.java
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.emf.ecore.EClass;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.ui.views.markers.WorkbenchMarkerResolution;
+
+public abstract class AbstractDataTypeErrorMarkerResolution extends WorkbenchMarkerResolution {
```
mx990: I would suggest extracting the shared functionality with `AbstractFBErrorMarkerResolution` into a common base class.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractFBErrorMarkerResolution.java
```diff
@@ -0,0 +1,104 @@
...
+	public TypeLibrary getTypeLibrary() {
+		try {
+			final EObject target = FordiacErrorMarker.getTarget(marker);
+			if (EcoreUtil.getRootContainer(target) instanceof final LibraryElement le) {
+				return le.getTypeLibrary();
+			}
+		} catch (final IllegalArgumentException | CoreException e) {
+			// target not found
+		}
+		return null;
```
mx990: I would suggest getting the type library via the resource of the marker, which avoids having to potentially load the target object:
```suggestion
		return TypeLibraryManager.INSTANCE.getTypeLibrary(marker.getResource().getProject());
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractFBErrorMarkerResolution.java
```diff
@@ -0,0 +1,104 @@
...
+			if (target instanceof final ErrorMarkerFBNElement ref) {
+				return findOtherTypeMarker(markers, ref.getType());
+			}
+			if (target instanceof final FB fb && fb.getTypeEntry() instanceof final ErrorFBTypeEntryImpl ref) {
+				return findOtherTypeMarker(markers, ref.getType());
+			}
+		} catch (final IllegalArgumentException | CoreException e) {
+			// target not found
+		}
+		return new IMarker[0];
```
mx990: I have created a PR https://github.com/eclipse-4diac/4diac-ide/pull/421 that adds additional information to error markers, which should make it easier to find other markers, particularly without relying on `FordiacErrorMarker.getTarget(...)`:
```suggestion
				return Stream.of(markers)
				.filter(other -> LibraryElementValidator.DIAGNOSTIC_SOURCE.equals(FordiacErrorMarker.getSource(other))
						&& LibraryElementValidator.TYPED_CONFIGUREABLE_OBJECT__VALIDATE_TYPE == FordiacErrorMarker
								.getCode(other)
						&& Arrays.equals(FordiacErrorMarker.getData(other), FordiacErrorMarker.getData(marker)))
				.toArray(IMarker[]::new);
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,77 @@
...
+	private TypeEntry newEntry;
+
+	public ChangeFBMarkerResolution(final IMarker marker) {
+		super(FordiacMessages.Repair_Dialog_ChangeFBType, marker);
+	}
+
+	@Override
+	public void run(final IMarker marker) {
+		if (newEntry == null) {
+			createNewEntry(); // TODO fix cancel selection
```
mx990: I would suggest to add a `canceled` flag that skips all future `run(IMarker)`s.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/FordiacMarkerResolutionGenerator.java
```diff
@@ -0,0 +1,66 @@
...
+				return new IMarkerResolution[] { new CreateMissingFBMarkerResolution(marker),
+						new ChangeFBMarkerResolution(marker) };
+			}
+			if (LibraryElementPackage.Literals.VAR_DECLARATION.equals(targetType)) {
+				return new IMarkerResolution[] { new ChangeDataTypeMarkerResolution(marker),
+						new CreateDataTypeMarkerResolution(marker) };
+			}
+		} catch (final IllegalArgumentException | CoreException e) {
+			e.printStackTrace();
+		}
```
mx990: I would suggest to filter using the new `FordiacErrorMarker.getSource/Code(IMarker)` from PR https://github.com/eclipse-4diac/4diac-ide/pull/421 and avoid `getTarget(IMarker)` altogether.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/FordiacMarkerResolutionGenerator.java
```diff
@@ -0,0 +1,66 @@
...
+					&& FordiacErrorMarker.getTarget(marker) instanceof final FB fb
+					&& fb.getTypeEntry() instanceof ErrorFBTypeEntryImpl) {
+				return true;
+			}
+
+			return LibraryElementPackage.Literals.ERROR_MARKER_FBN_ELEMENT.equals(targetType)
+					|| LibraryElementPackage.Literals.VAR_DECLARATION.equals(targetType);
+		} catch (final IllegalArgumentException | CoreException e) {
+			return false;
+		}
```
mx990: I would suggest to filter using the new `FordiacErrorMarker.getSource/Code(IMarker)` from PR https://github.com/eclipse-4diac/4diac-ide/pull/421 and avoid `getTarget(IMarker)` altogether.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/wizards/RestrictedNewTypeWizard.java
mx990: Done

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,83 @@
...
+import org.eclipse.fordiac.ide.model.ui.editors.DataTypeTreeSelectionDialog;
+import org.eclipse.fordiac.ide.model.ui.nat.DataTypeSelectionTreeContentProvider;
+import org.eclipse.fordiac.ide.model.ui.nat.TypeNode;
+import org.eclipse.fordiac.ide.ui.FordiacMessages;
+import org.eclipse.jface.window.Window;
+import org.eclipse.ui.PlatformUI;
+
+public class ChangeDataTypeMarkerResolution extends AbstractErrorMarkerResolution {
+
+	private LibraryElement newEntry;
```
mx990: Please rename this field to reflect that this is a data type and not a type entry. I would also suggest to change the type to `DataType` and perform the instanceof check early when choosing the new type.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,83 @@
...
+						interfaceElement, new NullProgressMonitor());
+			} else if (errorType instanceof final StructManipulator fb && newEntry instanceof final DataType d) {
+				AbstractLiveSearchContext.executeAndSave(new ChangeStructCommand(fb, d), fb, new NullProgressMonitor());
+			} else if (errorType instanceof final ConfigurableFB fb && newEntry instanceof final DataType d) {
+				AbstractLiveSearchContext.executeAndSave(new ConfigureFBCommand(fb, d), fb, new NullProgressMonitor());
+			}
+		}
+	}
+
+	private void createNewEntry() {
```
mx990: Please rename this method to accurately reflect that this is for choosing an (existing) data type instead of creating a new type entry.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	private void createNewEntry() {
+		final DataTypeTreeSelectionDialog dialog = new DataTypeTreeSelectionDialog(
+				PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
+				FBTypeSelectionTreeContentProvider.INSTANCE);
+		dialog.setInput(getTypeLibrary());
+		if (dialog.open() == Window.OK && dialog.getFirstResult() instanceof final TypeNode node
+				&& !node.isDirectory()) {
+			newEntry = node.getType().getTypeEntry();
```
mx990: I would suggest to create a `getTypeEntry()` method in `TypeNode` to handle this case without the indirection over the type.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,79 @@
...
+			return;
+		}
+
+		if (newEntry == null) {
+			createNewEntry();
+		}
+
+		if (newEntry != null) {
+			final EObject target = FordiacErrorMarker.getTarget(marker);
+			if (target instanceof final ErrorMarkerFBNElement err) {
```
mx990: Is there a reason to restrict this to only `ErrorMarkerFBNElement`s? If not, I would suggest to handle all `FBNetworkElement`s here. Of course, this would require to reorder and chain both `if`s.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -0,0 +1,79 @@
...
+			}
+			if (target instanceof final FB fb && fb.eContainer() instanceof final BaseFBType base
+					&& base.getInternalFbs().contains(fb)) {
+				AbstractLiveSearchContext.executeAndSave(new UpdateInternalFBCommand(fb, newEntry), fb,
+						new NullProgressMonitor());
+			}
+		}
+	}
+
+	private void createNewEntry() {
```
mx990: Again, please rename this method to accurately reflect that this is for choosing an (existing) FB type entry instead of creating a new one.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+
+	@Override
+	public void run(final IMarker marker) {
+		if (newEntry == null) {
+			createNewEntry();
+		}
+
+		if (newEntry != null) {
+			final EObject errorType = FordiacErrorMarker.getTarget(marker);
+			if (errorType instanceof final VarDeclaration target) {
```
mx990: Is there a reason to only handle `VarDeclaration` here instead of `IInterfaceElement`?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+				AbstractLiveSearchContext.executeAndSave(ChangeDataTypeCommand.forDataType(target, newEntry.getType()),
+						target, new NullProgressMonitor());
+			} else if (errorType instanceof final StructManipulator fb && newEntry instanceof final DataType d) {
```
mx990: I believe `newEntry instanceof final DataType d` is always false (same below)?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+				marker.getResource().getProject().getFullPath());
+		final TypeFromTemplateCreator creator = new TypeFromTemplateCreator(targetFile, template, ""); //$NON-NLS-1$
+		creator.createTypeFromTemplate(new NullProgressMonitor());
+
+		newEntry = (DataTypeEntry) creator.getTypeEntry();
+	}
+
+	private static IFile getTargetFile(final String typeName, final IPath path) {
+		return ResourcesPlugin.getWorkspace().getRoot()
+				.getFile(new Path(path + File.separator + "Type Library" + File.separator + typeName + ".dtp")); //$NON-NLS-1$ //$NON-NLS-2$
```
mx990: The `typeName` may contain `::` if it has a package (e.g., `pkgtest::test1::StructTest`), in which case this will produce an invalid path. I would also suggest to use the appropriate constants and helper methods from `IPath` to construct the path based on the project:
```suggestion
		return project.getFile(Path.fromOSString(SystemManager.TYPE_LIB_FOLDER_NAME).append(typeName.replace(PackageNameHelper.PACKAGE_NAME_DELIMITER, String.valueOf(IPath.SEPARATOR))).addFileExtension(TypeLibraryTags.DATA_TYPE_FILE_ENDING));
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+				AbstractLiveSearchContext.executeAndSave(new ConfigureFBCommand(fb, d), fb, new NullProgressMonitor());
+			}
+		}
+	}
+
+	private void createNewEntry() {
+		final File template = new File(TEMPLATE_PATH);
+		final IFile targetFile = getTargetFile(FordiacErrorMarker.getData(this.marker)[0],
+				marker.getResource().getProject().getFullPath());
+		final TypeFromTemplateCreator creator = new TypeFromTemplateCreator(targetFile, template, ""); //$NON-NLS-1$
```
mx990: Please use the correct package name (obtainable via `PackageNameHelper.extractPackageName(String)`) instead of an empty string.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateMissingFBMarkerResolution.java
```diff
@@ -0,0 +1,91 @@
...
+
+		if (target instanceof final FB fb && fb.getTypeEntry() instanceof ErrorFBTypeEntryImpl
+				&& fb.eContainer() instanceof final BaseFBType base && base.getInternalFbs().contains(fb)) {
+			AbstractLiveSearchContext.executeAndSave(new UpdateInternalFBCommand(fb, newEntry), fb,
+					new NullProgressMonitor());
+		}
+	}
+
+	@Override
+	public void run(final IMarker[] markers, final IProgressMonitor monitor) {
```
mx990, line 67: Looking at the code, it seems we have no choice but to process all markers at once here. In that case, I would suggest to delegate from `run(IMarker)` to this method instead to handle the case where that method is called directly. Please have a look at `WorkbenchMarkerResolutionAdapter.run(IMarker)` for an example.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/CreateDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,82 @@
...
+		creator.createTypeFromTemplate(new NullProgressMonitor());
```
mx990, line 81: I believe we should open the newly created type afterwards, see also `NewTypeWizard.openTypeEditor(IFile)`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeDataTypeMarkerResolution.java
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		if (selectedType == null) {
+			selectDataType();
+		}
+
+		if (selectedType != null) {
+			final EObject errorType = FordiacErrorMarker.getTarget(marker);
+			if (errorType instanceof final IInterfaceElement interfaceElement
+					&& interfaceElement.getType().getTypeEntry() instanceof ErrorDataTypeEntry) {
```
mx990: This may cause an NPE in case the reason for the error is that the type is null. Also, why do you need to explicitly check if it is an `ErrorDataTypeEntry` here in the first place?

mx990 (changes requested):

mx990 (commented):
Looks good to me now, aside from my earlier comment https://github.com/eclipse-4diac/4diac-ide/pull/405#discussion_r1725055464.

PR #411, opened 2024-08-21
Add support for generic internal FBs in forte_ng export

mx990:
Add support for generic internal FBs in forte_ng export using a new attribute for the generic class name.


azoitl:
shouldn't the manifest files for the typelibs that now use the attribute get a dependency to the new core lib?

mx990:
> shouldn't the manifest files for the typelibs that now use the attribute get a dependency to the new core lib?

Yes, I have added the missing dependencies.

azoitl (approved):

PR #419, opened 2024-08-25
DataTypeEditorTests class extends from NatTableWithoutEditorBehaviorTests

ShantKhatri:
As DataTypeEditorTests has same test methods as NatTableWithoutEditorBehaviorTests and this editor also doesn't have any editor behavior while operating with NatTable. So extended it from NatTableWithoutEditorBehaviorTests to reduce the code duplicacy.


ShantKhatri:
This PR has a dependency on #381.

bwiesmayr (approved):

PR #426, opened 2024-08-27
Load expanded subapp for closed editors

bwiesmayr:
Add async execution to make sure the editor is loaded before the position of the editpart of an expanded subapp is asked. 

This patch also ensures that subapp is correctly opened with go to child.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/AbstractBreadCrumbEditor.java
+++ b/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/AbstractBreadCrumbEditor.java
@@ -156,7 +156,7 @@ public abstract class AbstractBreadCrumbEditor extends AbstractCloseAbleFormEdit
 		final EObject container = getFBNetworkContainer(subapp);
 		breadcrumb.setInput(container);
 		openEditor(container);
-		Display.getCurrent().asyncExec(() -> HandlerHelper.showExpandedSubapp(subapp, getActiveEditor()));
+		HandlerHelper.showExpandedSubapp(subapp, this);
 	}
 
 	private void openEditor(final Object element) {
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/AbstractBreadCrumbEditor.java
```diff
@@ -141,16 +142,28 @@ protected Composite createPageContainer(final Composite parent) {
...
-				setActivePage(pagenum);
-				getSite().getPage().getNavigationHistory().markLocation(getNavigationLocationProvider());
-			}
+			openEditor(element);
+		}
+	}
+
+	private void showAndSelectExpandedSubapp(final SubApp subapp) {
+		final EObject container = getFBNetworkContainer(subapp);
+		breadcrumb.setInput(container);
```
azoitl: isn't setInput resulting in the end such that openEditor is called and therefore you now are opening the editor twice?

bwiesmayr: after the user selects the expanded subapp in the breadcrumb, the breadcrumb takes the item. This is not what we want because we opened actually the parent editor. So it could be that the openEditor() method (else-branch) is called again, but nothing should happen because the editor is already open. Calling setInput() ensures that the breadcrumb shows the correct contents and layouts the breadcrumb. What I could maybe do now is calling setInput() with SelectionChangedEvent = null, as long as I add null checks everywhere. Do you think that's better?

azoitl: I don't think the null checks would make it better. I know that setInput on the breadcrump is important. I just was not sure if the openEditor afterwards is double and may cause problems later

azoitl (approved):

PR #427, opened 2024-08-27
STOPWATCH: Add new events ET and RESET, update ECState positions, and implement reset functionality. Improve algorithm readability.

franz-hoepfinger-4diac:
the STOPWATCH is not really a Stopwatch. 

Time output is only possible at Stop, think on a Sports Event: you want to get like Round Times, 
also there is no Reset to 0.

### DIFF
--- a/data/typelibrary/utils-1.0.0/typelib/E_STOPWATCH.fbt
+++ b/data/typelibrary/utils-1.0.0/typelib/E_STOPWATCH.fbt
@@ -18,12 +18,10 @@
 			</Event>
 		</EventInputs>
 		<EventOutputs>
-			<Event Name="STARTO" Type="Event" Comment="Start the stopwatch confirm">
-			</Event>
-			<Event Name="ETO" Type="Event" Comment="Trigger the Time output  confirm">
+			<Event Name="EO" Type="Event" Comment="Event Output">
 				<With Var="TD"/>
 			</Event>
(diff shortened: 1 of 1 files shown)


azoitl:
Is it really a good idea to have only one output event and send always EO on all the new different input events?

franz-hoepfinger-4diac:
> Is it really a good idea to have only one output event and send always EO on all the new different input events?

no, that is a BAD idea. 

i will improve !

franz-hoepfinger-4diac:
@azoitl now, this looks nicer right ?

azoitl (commented):
I don't know for me are this a bit to much output events now. Especially they are now confusing my how to normally and correctly use that block. The original functionality was to measure the time between a start and a stop event. I understand the ET event and having for that an ETo event. Also the reset event makes sense with a dedicated RESETO but why a STARTO?

franz-hoepfinger-4diac:
so @azoitl we make a EO, which is triggered on START and on RESET ? 

so make 
ETO, RESETO, and EO, 
OK ?

azoitl:
> so @azoitl we make a EO, which is triggered on START and on RESET ?
> 
> so make ETO, RESETO, and EO, OK ?

why an output event on START?

franz-hoepfinger-4diac:
> why an output event on START?

i did remove it: 

image
image

azoitl (approved):

PR #434, opened 2024-08-28
Fix validation of identifiers in model

mx990:
Fix validation of identifiers in model to disallow consecutive or trailing underscores and also add tests.


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/CreateGroupCommand.java
```diff
@@ -27,7 +27,7 @@ public class CreateGroupCommand extends AbstractCreateFBNetworkElementCommand {
 
 	private static final int GROUP_TOP_BORDER = 20;
 	private static final int GROUP_BORDER = 2;
-	private static final String INITIAL_GROUP_NAME = "__Group01"; //$NON-NLS-1$
```
ernstblechaPT, line 30: I'm not sure if this should go straight into freeze with the old default Group Name being invalid

mx990: That is a good point. I added a separate commit that produces a warning instead of an error for now, while keeping the strict validation in place for user edits.

This should allow users to gracefully transition away from invalid names, while still preventing any invalid names from being used in new code.

ernstblechaPT (approved):

azoitl (approved):

PR #435, opened 2024-08-29
Show Internal FB Interface Values for SamplingEvaluators

azoitl:
When an FB is debugged with a SamplingEvaluator the debug view shows now both the outside and the internal values.

@mx990 I made the sampling evaluator the default. Is this ok?


mx990 (approved):
> I made the sampling evaluator the default. Is this ok?

Yes, now that we have a proper view for it.

bwiesmayr (approved):

PR #438, opened 2024-08-30
Add option to use system includes in forte_ng export

mx990:
Add an experimental argument `-exportSystemIncludes` to use system includes (i.e., `#include <path>`) instead of user includes (i.e., `#include "path"`) for dependencies in forte_ng export. This allows to avoid header file name collisions between identically named types with and without package.

The FB types `Test` and `pkgtest::test1::Test`, for example, currently result in two header files:
```
Test_fbt.h
pkgtest/test1/Test_fbt.h
```
If the second FB type uses the first internally, this would result in an include directive in `pkgtest/test1/Test_fbt.h`:
```cpp
#include "Test_fbt.h"
```
which means it will include itself, since the current folder is searched first for user includes.

Using a system include:
```cpp
#include <Test_fbt.h>
```
instead refers to the correct file, since that only searches the system include paths without considering the current folder.

At the same time, using system includes may break certain build environments, particularly when cross-compiling. Therefore, this is added as an experimental option only for now.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
@@ -75,8 +75,6 @@ abstract class ForteFBTemplate<T extends FBType> extends ForteLibraryElementTemp
 		#include "«type.generateTypeGenIncludePath»"
 		#endif
 		
-		#include "criticalregion.h"
-		#include "resource.h"
 		«getDependencies(emptyMap).generateDependencyIncludes»
 		«type.compilerInfo?.header»
 	'''
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
@@ -79,18 +79,6 @@ abstract class ForteFBTemplate<T extends FBType> extends ForteLibraryElementTemp
 		«type.compilerInfo?.header»
 	'''
 
-	def protected generateImplTypeIncludes(Iterable<DataType> vars) '''
-		«IF !vars.empty»
-			«vars.generateTypeIncludes»
-		«ENDIF»
-	'''
-
-	def protected generateAdapterIncludes(Iterable<AdapterDeclaration> vars) '''
-		«FOR include : vars.map[type.generateDefiningInclude].toSet»
-			#include "«include»"
-		«ENDFOR»
-	'''
-
 	def protected generateFBDeclaration() '''
 		DECLARE_FIRMWARE_FB(«FBClassName»)
 	'''
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
@@ -28,20 +28,22 @@ abstract class ForteNgExportTemplate extends ExportTemplate {
 	def protected generateDependencyIncludes(Iterable<? extends INamedElement> dependencies) '''
 		«dependencies.filter(DataType).generateTypeIncludes»
 		«FOR include : dependencies.reject(DataType).map[generateDefiningInclude].toSet.sort»
-			#include "«include»"
+			«include.generateDependencyInclude»
 		«ENDFOR»
 	'''
 
 	def protected generateTypeIncludes(Iterable<DataType> types) '''
 		«FOR include : types.map[generateDefiningInclude].toSet.sort»
-			#include "«include»"
+			«include.generateDependencyInclude»
 		«ENDFOR»
-		#include "iec61131_functions.h"
-		#include "forte_array_common.h"
-		#include "forte_array.h"
-		#include "forte_array_fixed.h"
-		#include "forte_array_variable.h"
+		«generateDependencyInclude("iec61131_functions.h")»
+		«generateDependencyInclude("forte_array_common.h")»
+		«generateDependencyInclude("forte_array.h")»
+		«generateDependencyInclude("forte_array_fixed.h")»
+		«generateDependencyInclude("forte_array_variable.h")»
 	'''
+
+	def protected generateDependencyInclude(String path) '''#include "«path»"'''
 
 	def getFileBasename() { name.replaceAll("\\.[^.]+$", "") }
 }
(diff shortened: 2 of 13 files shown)


ernstblechaPT (approved):

azoitl:
@jwalt as our cross compiling expert do you see any issues here?

jwalt:
I think this is a difficult situation. `#include "Test_fbt.h"` (as it is now) is unconditionally wrong. Double quoted includes should always refer to relative filenames, so I think the exporter's behaviour should change unconditionally. A proper fix depends on the exact scenario:

1. If there is only one source tree that refers to each other's files, then export as `#include "../../Test_fbh.t"`. I believe that is not a valid assumption, people will have more than one source tree (4diac-forte git and external modules, for example)
2. Do it like in the proposed patch, always, no command line option needed, and have a single `-I path/to/root/of/modules/dir` for each source tree, but no `-I` options for any subdirectories of that (because that would break this fix again)

Additionally, you might consider enforcing a common prefix directory on export, like `fbt/` so that include lines look like `#include <fbt/Test_fbt.h>`. This would reduce the chance of path name conflicts but impose a certain directory structure on users.

mx990:
> Double quoted includes should always refer to relative filenames, [...]

I have to *strongly* disagree with that statement. The `" "` form is normally used for header files from the current library or program, regardless of whether it is relative to the current directory or a (user-defined) include path. The `< >` form, on the other hand, is normally used for system headers and also often for *external* dependencies, such as boost or Qt.

The C++ standard gives a clear recommendation in section *19.2 Source file inclusion*:
> Although an implementation may provide a mechanism for making arbitrary source files available to
the < > search, in general programmers should use the < > form for headers provided with the implementation,
and the " " form for sources outside the control of the implementation. For instance:
>  ```cpp
>  #include <stdio.h>
>  #include <unistd.h>
>  #include "usefullib.h"
>  #include "myprog.h"
>  ```

The [gcc manual](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html) is a bit more direct and also has additional information on its implementation-specific behavior:
> `#include "file"`
>     This variant is used for header files of your own program. It searches for a file named file first in the directory containing the current file, then in the quote directories and then the same directories used for `<file>`.

There are also options in gcc besides `-I` to influence each set of include directories separately, that is, `-iquote` and `-isystem`. That is also why I was originally referring to them as *user* and *system* includes, respectively.

> 1. If there is only one source tree that refers to each other's files, then export as `#include "../../Test_fbh.t"`. I believe that is not a valid assumption, people will have more than one source tree (4diac-forte git and external modules, for example)

I agree that relative paths using `..` should be avoided. I would like to point out that the source tree of 4diac-forte itself is also by no means uniform, with several independent modules, tests, and so on.

> 2. Do it like in the proposed patch, always, no command line option needed, and have a single `-I path/to/root/of/modules/dir` for each source tree, but no `-I` options for any subdirectories of that (because that would break this fix again)

This is already the current behavior in the forte_ng export. That is why the problem only occurrs when using types without a package, which will currently end up in the top-level folder (and include path) without any prefix, see also my earlier example.

> Additionally, you might consider enforcing a common prefix directory on export, like `fbt/` so that include lines look like `#include <fbt/Test_fbt.h>`. This would reduce the chance of path name conflicts but impose a certain directory structure on users.

I would definitely prefer to enforce a specific directory structure on export in the future. We should also be more strict with include directories and header file structure in 4diac-forte itself. I am particularly concerned about potential name conflicts with system headers when using generic names, such as `device.h`.

jwalt:
> > Double quoted includes should always refer to relative filenames, [...]
> 
> I have to _strongly_ disagree with that statement. The `" "` form is normally used for header files from the current library or program, regardless of whether it is relative to the current directory or a (user-defined) include path. The `< >` form, on the other hand, is normally used for system headers and also often for _external_ dependencies, such as boost or Qt.

I generally agree with you, but I think in our case the generic recommendations from the standard are not helpful.
In complex projects, and 4diac-forte in particular, management of compiler include directories is a mess. There is much potential for ambiguity and name clashes, especially in our code base and with our multitude of target platforms (think the exotic embedded OSes we also support). The trouble is that external libraries can and do add compiler options like `-I` which affect the `"..."` search path as well. This is a mess, but we need to live with that -- forcing all possible dependency libraries to improve their interoperability is unrealistic. For that we have too many exotic components we may want to use.

Add to that the fact that people build custom code external to the 4diac-forte source. In this case the notion of which files belong to the implementation is vague: 4diac-forte has no control over external modules, and to external modules the 4diac code base also looks like external code. Thus I don't see a decent use for the `"..."` vs. `<...>` distinction except using `"..."` to enforce relative path name lookup (creating a strongly-coupled dependency).

In my opinion, a clean way of dealing with all that would be something like this:

4diac-forte should not refer to external modules at all. External modules should treat 4diac-forte as an external library. So including 4diac-forte headers should ideally look something like this: `#include <4diac-forte/core/device.h>`. 4diac-forte would have a single `-I` option, nothing more. All headers from within 4diac-forte is referred through the path prefix `4diac-forte`, which would of course need a slight adaptation of the source tree.

Generated code remains problematic. Modules from 4diac-forte live in directories that are not derivable from the type name, and for external module directories all bets are off. Therefore, we should either enforce module path names to be derivable from the type name/package, and/or we introduce a different prefix for fbt includes (`#include <4diac-fbt/...>`), and each module (including external ones) can then safely add `-I` flag as needed. 

> > 1. If there is only one source tree that refers to each other's files, then export as `#include "../../Test_fbh.t"`. I believe that is not a valid assumption, people will have more than one source tree (4diac-forte git and external modules, for example)
> 
> I agree that relative paths using `..` should be avoided. I would like to point out that the source tree of 4diac-forte itself is also by no means uniform, with several independent modules, tests, and so on.

I actually think relative paths with back references are fine, as long as the strongly-coupled dependency they create reflects the spirit of the code, for example inside the 4diac core. I fully agree that the current state of 4diac is an utter mess. I'd love to refactor that.

> > 2. Do it like in the proposed patch, always, no command line option needed, and have a single `-I path/to/root/of/modules/dir` for each source tree, but no `-I` options for any subdirectories of that (because that would break this fix again)
> 
> This is already the current behavior in the forte_ng export. That is why the problem only occurrs when using types without a package, which will currently end up in the top-level folder (and include path) without any prefix, see also my earlier example.

I thought the problem was because `#include "..."` fails because of a file with the same name reachable using relative path name. `#include <...>` would fix that. But as we seem to agree, this is really just a band-aid that is fragile in other ways.

> I would definitely prefer to enforce a specific directory structure on export in the future. We should also be more strict with include directories and header file structure in 4diac-forte itself. I am particularly concerned about potential name conflicts with system headers when using generic names, such as `device.h`.

Absolutely, thus my proposal above.

mx990:
> I thought the problem was because `#include "..."` fails because of a file with the same name reachable using relative path name. `#include <...>` would fix that. But as we seem to agree, this is really just a band-aid that is fragile in other ways.

Yes, this PR was never intended as anything but a temporary, but unfortunately necessary, band-aid for the problem at hand. That is why it is merely added as a command line argument for specific use cases instead of changing the default behavior. I plan to revert the relevant commit once the proper solution is in place.

azoitl (approved):

azoitl:
As it is an intermediate fix I would merge that for now to get our users going. I understood some major 4diac FORTE restructuring is upcoming.

PR #446, opened 2024-09-03
Show full typename in With-Table to also include Arrays

sebHollersbacher:


oberlehner:
It is still missing in the data section:
image

oberlehner:
Please make a new PR for the Data property section.

PR #447, opened 2024-09-03
Rework Library dependency resolve

pazi146:
Resolves entire dependency tree before importing all libraries into project
(old version sequentially imported one library after the other)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -834,125 +712,300 @@ public IStatus runInWorkspace(final IProgressMonitor monitor) throws CoreExcepti
...
+		linked.forEach(l -> {
+			try {
+				project.getFolder(TYPE_LIB_FOLDER_NAME).getFolder(l).delete(true, null);
+			} catch (final CoreException e) {
+				// empty
 			}
+		}); // remove still linked libraries
+
+		FordiacMarkerHelper.updateMarkers(project.getFile(MANIFEST), FordiacErrorMarker.LIBRARY_MARKER, markerList,
+				true);
```
mx990, line 830: I would suggest to create the markers directly, since we are already running in a workspace job:
```suggestion
		final IFile manifestFile = project.getFile(MANIFEST);
		if (manifestFile.isAccessible()) {
			try {
				manifestFile.deleteMarkers(FordiacErrorMarker.LIBRARY_MARKER, true, IResource.DEPTH_INFINITE);
				for (final var builder : markerList) {
					builder.createMarker(manifestFile);
				}
			} catch (final CoreException e) {
				FordiacLogHelper.logError("Could not update error markers on resource: " + manifestFile.getName(), //$NON-NLS-1$
						e);
			}
		}
```

tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java
mx990, line 1: I believe we may need to wait for the job to complete after `FordiacMarkerHelper.updateMarkers(...)` in `cleanupAfterEach()`.

mx990 (changes requested):
I was unable to reproduce the failing test on my machine, but I noticed two potential problems, which may be causing the test to fail.

mx990 (changes requested):
I was unable to reproduce the failing test on my machine, but I noticed two potential problems, which may be causing the test to fail.

PR #452, opened 2024-09-04
Added the AttributeDeclaration to the create attribute command

oberlehner:
Therefore the attDecl is not null and the search works again after saving an application.


azoitl:
Please do not change this command. Use the `ChangeAttributeDeclarationCommand` for updating attribute declarations.

oberlehner:
> Please do not change this command. Use the `ChangeAttributeDeclarationCommand` for updating attribute declarations.

changed that

azoitl:
@mx990 I noticed that we didn't add the attribute declarations to our dependency tracking. I tried to do that now. Can you please check if you used the api correct. At least in my tests it worked.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -439,7 +439,7 @@ protected void parseGenericAttributeNode(final ConfigurableObject confObject)
 						name -> getTypeLibrary().getAttributeTypeEntry(name), name -> null);
 				final AttributeTypeEntry attributeTypeEntry = addDependency(entry);
 				if (attributeTypeEntry != null && attributeTypeEntry.getType() != null) {
-					attribute.setAttributeDeclaration(attributeTypeEntry.getType());
+					attribute.setAttributeDeclaration(addDependency(attributeTypeEntry.getType()));
```
mx990: There is already an `addDependency(...)` in line 440.

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -146,18 +147,16 @@ private void handleDependencyUpdate(final TypeEntry typeEntry) {
...
+		final AttributeTypeInstanceSearch search = new AttributeTypeInstanceSearch(editedElement, atEntry);
 		final List<? extends EObject> result = search.performSearch();
 
 		result.forEach(at -> {
 			// update attribute here
 			if (at instanceof final ConfigurableObject co) {
-				final UpdateAttributeDeclarationCommand cmd = new UpdateAttributeDeclarationCommand(co, atEntry);
-				cmd.execute();
+				final Attribute attribute = co.getAttribute(atEntry.getTypeName());
+				ChangeAttributeDeclarationCommand.forName(attribute, atEntry.getTypeName()).execute();
```
mx990: This may break depending on how the type name is used, that is, with or without package name. I would suggest extracting a `forTypeEntry(AttributeTypeEntry)` method from: https://github.com/eclipse-4diac/4diac-ide/blob/9aa7a522ec56eebea9c4629d9857d6e0726ef8e4/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeAttributeDeclarationCommand.java#L67-L73

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
azoitl: Damn missed that. Fixed.

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: Good point. Done.

mx990 (approved):

ernstblechaPT (approved):

PR #465, opened 2024-09-10
Ant task check dependencies

pazi146:
- checks for unused dependencies


plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui.tests/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,6 @@ Bundle-ActivationPolicy: lazy
 Require-Bundle: org.eclipse.fordiac.ide.globalconstantseditor.ui,
  org.eclipse.xtext.testing,
  org.eclipse.xtext.xbase.testing,
- org.eclipse.xtext.xbase.junit,
```
ernstblechaPT: I think this should be at least a seperate commit

pazi146: thanks for catching that

ernstblechaPT:
I have tested the ant-task with a new, empty project (no instances in appliction, no system configured):

` Unused dependencies detected: convert, signalprocessing, net, utils, math, devices, segments, reconfiguration, resources`

I don't see why the rest of the linked libraries are not listed as well?

oberlehner:
` iec61131-3` is also missing.

pazi146:
Some standard libraries depend on others:

- `convert` depends on `core`
- `events` depends on `iec61131-3`
- `iec61131-3` depends on `core`
- `net` depends on `core`
- `rtevents` depends on `core` and `events`
- `utils` depends on `core`

Result: `core`, `events` and `iec61131-3` will not show as unused when including all standard libraries

ernstblechaPT (approved):

PR #481, opened 2024-09-13
Avoid removing and re-adding annotations during live validation

mx990:
Avoid removing and re-adding annotations during live validation, including resulting superfluous notifications to listeners.


ernstblechaPT (approved):

ernstblechaPT (commented):
Only realised too late that I am needed to test this...

 - Works as soon as the editor is dirty
 - When you edit a block where the existing error marker comes from a saved state it still jumps back to the beginning

ernstblechaPT (approved):
Cursor now stays in place

azoitl (approved):

PR #485, opened 2024-09-13
Add INI File Storage FB

franz-hoepfinger-4diac:
Add INI File Storage FB


azoitl (commented):
why not just one block with any in and outputs for the values? It is anyhow an sifb.

diplfranzhoepfinger:
> why not just one block with any in and outputs for the values? It is anyhow an sifb.

this is the reason i am posting it that way. 

YES, makes Sense. I change it.

diplfranzhoepfinger:
chnaged to ANY

diplfranzhoepfinger:
better with ANY\_ELEMENTARY  
[https://www.fernhillsoftware.com/help/iec-61131/common-elements/datatypes-generic.html](https://www.fernhillsoftware.com/help/iec-61131/common-elements/datatypes-generic.html)

Struct and Array Types might be difficult,  
while INT, Real and STRING is possible.

azoitl (approved):

PR #486, opened 2024-09-13
add ESP32 Storage on NVS

franz-hoepfinger-4diac:
add ESP32 Storage on NVS


diplfranzhoepfinger:
chnaged to ANY
@azoitl to make it consistent with INI

franz-hoepfinger-4diac:
better with ANY_ELEMENTARY
https://www.fernhillsoftware.com/help/iec-61131/common-elements/datatypes-generic.html

Struct and Array Types might be difficult, 
while INT, Real and STRING is possible.

franz-hoepfinger-4diac:
better with ANY_ELEMENTARY
https://www.fernhillsoftware.com/help/iec-61131/common-elements/datatypes-generic.html

Struct and Array Types might be difficult, 
while INT, Real and STRING is possible.

Issue #489, opened 2024-09-14
adapters in FBs are shown in lower case


moniwe:
using 4diac-ide Version: 3.0.0 Build id: 2024-09-14_0544 shows adapters in lower case within the interface, selecting there within the drop down also shows it in lower case:
![grafik](https://github.com/user-attachments/assets/5f8cfc77-28a0-41af-aa8e-90d5b3b5b190)

when entering it manually within the tabbed properties it accepts the capital letters:
![grafik](https://github.com/user-attachments/assets/344fa439-69b0-4008-9a99-876767f84af2)

it seems that it became an issue due to package name changes.

mx990:
Thank you for the hint about the package name change. This enabled me to reproduce the problem. The name of the created error type entry used the lowercase name instead of the original name. I have created a PR https://github.com/eclipse-4diac/4diac-ide/pull/512 to fix the problem.

moniwe:
thanx

PR #397, opened 2024-08-20
Do not overwrite unchanged export

ernstblechaPT:
During export of function blocks already existing files should not be overwritten by files of identical content.

Also contains changes that modernize the surrounding code (e.g. records instead of handcrafted data-classes)

PR #490, opened 2024-09-14
Merge dialog needs to be run in display thread  #489

azoitl:
As exporting is running in a separate thread all ui interactions triggered by the exporter need to be run in the display thread. This patch fixes that for the merge question dialog.

Furthermore the invocation was cleaned a bit.

https://github.com/eclipse-4diac/4diac-ide/issues/489


ernstblechaPT (changes requested):
This still fails with a "ThreadAccessException" when opening the CompareEditor (approximatly line 200) - with the following change this seems to be solved:

```
					Display.getDefault().syncExec(() -> {
						opener.openCompareEditor();
					});
```

azoitl:
> This still fails with a "ThreadAccessException" when opening the CompareEditor (approximatly line 200) - with the following change this seems to be solved:
> 
> ```
> 					Display.getDefault().syncExec(() -> {
> 						opener.openCompareEditor();
> 					});
> ```

thanks for noting. I think here an async call is fine. A few lines below I found a further MessageDialog that I had to wrap in an display thread call. Hope this is now all.

ernstblechaPT (approved):
Yes, syncExec should be fine here - this will only change when the compareEditor opens;

Regarding that other piece of code: This should be mostly dead code as the "no changes" case is handled outside due to my earlier fix from #397

azoitl:
> Yes, syncExec should be fine here - this will only change when the compareEditor opens;
> 
> Regarding that other piece of code: This should be mostly dead code as the "no changes" case is handled outside due to my earlier fix from #397

should we then remove that part. The idea was that when the user selected merge to inform them one nothing was to merge. So it should not be dead code, or?

ernstblechaPT:
> should we then remove that part. The idea was that when the user selected merge to inform them one nothing was to merge. So it should not be dead code, or?

Yes, this is now dead code - the user no longer has the option to select "merge" when there were no changes

ernstblechaPT (approved):

PR #504, opened 2024-09-17
Fix editor updates for transitive notifications in type entry

mx990:
For transitive notifications of changed types, either the old or new value is usually null. This resulted in no notification being sent to the editor. The editor update should also get the changed type entry from the notification instead of indirectly via the type entry of the new value.


oberlehner (commented):
this might throw an exception in org.eclipse.fordiac.ide.model.edit.TypeEntryAdapter.notifyChanged(Notification) in Assert.isTrue(notification.getNewValue() instanceof TypeEntry); when it is null.
Are you sure the newValue is never null ?

mx990:
> this might throw an exception in org.eclipse.fordiac.ide.model.edit.TypeEntryAdapter.notifyChanged(Notification) in Assert.isTrue(notification.getNewValue() instanceof TypeEntry); when it is null. Are you sure the newValue is never null ?

The `newValue` of the notification is now initialized from `notification.getNotifier()` (i.e., where the original notification came from). If that was `null`, there would be a serious problem with the implementation of whoever sent the notification.

PR #506, opened 2024-09-17
Make the import-ant-task wait for the library download

ernstblechaPT:


plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/Import4diacProject.java
```diff
@@ -144,7 +144,8 @@ private IProgressMonitor getMonitor() {
 
 	public static void waitBuilderJobsComplete() {
 		waitJobsComplete(job -> !(job.getName().startsWith("Compacting") //$NON-NLS-1$
-				|| job.getName().startsWith("Periodic workspace save."))); //$NON-NLS-1$
+				|| job.getName().startsWith("Periodic workspace save.") //$NON-NLS-1$
```
oberlehner: can you make  a constant for that ? 
when somechanges the tet in the job, it will fail.
this will not happen if waitBuilderJobsComplete and the workspace job share a constant.
Would be great, if it does not lead to cyclic dependency conflicts .

ernstblechaPT: Sure, I can use the message-strings of the respective packages

PR #510, opened 2024-09-17
[Fix] Type Initial Values Where not Shown at Block Pins

azoitl:
If a FB or SubApp type has type initial values these values where not shown in grey at instances of these types.


plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
```diff
@@ -40,6 +42,12 @@ public static String getInitialOrDefaultValue(final Object element) {
 			if (hasInitalValue(element)) {
 				return varDec.getValue().getValue();
 			}
+
+			final String typeInitialValue = getTypeInitialValue(varDec);
+			if (!typeInitialValue.isBlank()) {
+				return typeInitialValue;
+			}
+
```
mx990, line 50: This is supposed to be already handled by `VariableOperations.newVariable(...)` in `getDefaultValue(...)` below. I have submitted a PR https://github.com/eclipse-4diac/4diac-ide/pull/513 to fix the underlying problem.

azoitl: Thx. This was the first place I could find a fix for that problem.

PR #514, opened 2024-09-18
Workaround for issue : Copying structs to other folder creates 2 identical structs

PStmmr:
This workaround creates a separate window, when a struct is copied and pasted with the same name. If the window gets closed without changes, the copied file will get deleted.


PStmmr:
The only problems are then:
1) If you copy a struct into the same folder, the window pops up twice because there is already a check for it.

2) If you delete a struct and then execute undo (ctrl+z), then the reset is seen as an error and you have to rename the file (the window pops up)

diplfranzhoepfinger:
> The only problems are then:
> 
> 1.  If you copy a struct into the same folder, the window pops up twice because there is already a check for it.
> 2.  If you delete a struct and then execute undo (ctrl+z), then the reset is seen as an error and you have to rename the file (the window pops up)

this is same for EVERY FB. 

e.g. if you copy a Folder, but rename the Folder, all FBs are duplicated.

PR #521, opened 2024-09-19
Add new SCALE function block for signal processing

franz-hoepfinger-4diac:
Created a new XML file defining the SCALE function block for scaling input values to desired output range in the signal processing library.


azoitl (approved):

franz-hoepfinger-4diac:
image

PR #522, opened 2024-09-19
Add new scaling function SCALE_LIM.fct with limit feature. Define input/output variables and logic for scaling calculation.

franz-hoepfinger-4diac:
Add new scaling function SCALE_LIM.fct with limit feature. Define input/output variables and logic for scaling calculation.


azoitl (commented):
I find excape a strange name for the upper and lower bounds. Normally I would call them limit. Also the four different escape values can not be understood just from the names.

franz-hoepfinger-4diac:
> I find excape a strange name for the upper and lower bounds. Normally I would call them limit. Also the four different escape values can not be understood just from the names.

you are right, i will propose better names.

franz-hoepfinger-4diac:
@azoitl see now. 
i also changed SCALE, to be consistent. 
see https://github.com/eclipse-4diac/4diac-ide/pull/527

franz-hoepfinger-4diac:
image

azoitl (approved):

PR #525, opened 2024-09-19
Enable seamless drag-and-drop between graphical editors

mx990:
This is currently a rough proof of concept. It is based on the idea of always performing drag-and-drop and forwarding `DragOver` events as `MouseEvent`s to GEF to enable the GEF tools to work correctly when inside the current editor.


mx990:
The approach in https://github.com/eclipse-4diac/4diac-ide/pull/530 is more promising.

PR #528, opened 2024-09-20
Enable seamless connections between graphical editors

mx990:
This is currently a rough proof of concept. It attempts to find the target in another editor based on the current mouse position. There are some missing checks and only normal connection dragging (without modifiers) works so far. The connection feedback also does not work at the moment.


mx990:
The approach in https://github.com/eclipse-4diac/4diac-ide/pull/530 is cleaner.

PR #530, opened 2024-09-20
Enable seamless connections between graphical editors

mx990:
This is currently a rough proof of concept. It attempts to find the target in another editor based on the current mouse position and then feed that as a converted mouse event to the tool implementation. There are some missing checks and connection feedback does not work at the moment.


azoitl:
Based on this work PR #2027 was done. Therefore I'm closing this one.

PR #532, opened 2024-09-22
update Wago FBs that they use STRING instead of WSTRING

moniwe:
this is related to [Wago updates on forte PR 238](https://github.com/eclipse-4diac/4diac-forte/pull/238)


azoitl (changes requested):
can you please update the copy right headers

azoitl (approved):

PR #542, opened 2024-09-24
Fixed creation of pins with array-datatype when creating new SubApp

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/CreateInterfaceElementCommand.java
```diff
@@ -85,6 +85,11 @@ public CreateInterfaceElementCommand(final DataType dataType, final String name,
 		this(dataType, name, interfaceList, isInput, false, index);
 	}
 
+	public CreateInterfaceElementCommand(final DataType dataType, final String name, final InterfaceList interfaceList,
```
oberlehner, line 88: Please make static create helper methods with meaningfull names. If someone wants to reuse this command it is really hard to find the right constructor.

In general this class needs a cleanup, the code looks awful. I know this is not your fault, but we should use the chance to make it even look a little bit nicer here

oberlehner: But please make a new PR on develop for that.

PR #543, opened 2024-09-24
Improvement of Layout of Table in LibrarySelection page

azoitl:
The table showing available libraries in the LibrarySelection page is now using the full available space and adjusts its size on resizing the dialog.


pazi146 (commented):
Resize doesn't work - table columns have fixed width

azoitl:
> Resize doesn't work - table columns have fixed width

This is an SWT Bug/Features. would require more work that I see in a separate PR. 

For more details see here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=13467

pazi146 (approved):

PR #545, opened 2024-09-24
Refactor locking implementation in type entry

mx990:
Refactor locking implementation in type entry to improve performance and avoid deadlocks due to chained notifications while holding a lock.


oberlehner (approved):
Thx for the comments, this makes the code clearer.

azoitl (approved):

PR #551, opened 2024-09-25
Bugfix in remove watch handler

oberlehner:
right clicking on an error fbnetworkelement inside an editor of a baseFbType caused an type cast exception in the setEnabled method of the watches handler


oberlehner:
duplicate of #552

PR #552, opened 2024-09-25
Added check to prevent ClassCastException

sebHollersbacher:
ClassCastException could be thrown when opening the ContextMenu on an errorMarker


oberlehner:
i think this is a duplicate of https://github.com/eclipse-4diac/4diac-ide/pull/551

PR #560, opened 2024-09-26
Extracted basic type editor handling into own AbstractTypeEditor class

azoitl:
As all type editors (FBType, Attribute, and DataType) have a lot of common code this commit extracted the base functionality of the FBTypeEditor into a new plugin.

As part of this work also the overall infrastructure was cleaned and improved.


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeXtextEditor.java
mx990, line 1: Should we then also move this to `typeeditor` and generalize it to `TypeEditorXtextPage`?

mx990 (commented):
After removing `schema/fBTEditorTabs.exsd` from `plugins/org.eclipse.fordiac.ide.fbtypeeditor/build.properties`, the build worked for me.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeXtextEditor.java
azoitl: @mx990 thx for pointing that out. Yesterday I didn't have time to check  the maven output.

Regarding `FBTypeExtextEditor`: I would say this depends. I only would see this valuable if pure XTextEditors (e..g, GlobalConsts) would later also need more editors tabs (e.g., an Description tab added). 

Anyhow this is only the first commit in a longer chain of refactoring. The next step is to migrate type saving to the `AbstractTypeEditor` and then migrate either DataTypeEditor or AttributeEditor to the new infra to see if it works and what is still missing in `AbstractTypeEditor`.

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/TypeEditorInput.java
```diff
@@ -11,28 +11,31 @@
...
 
 /**
- * An EditorInput for opening FBType files with specified content. The equals
+ * An EditorInput for opening any type file with specified content. The equals
  * method is adapted that EditorInput is equal to another EditorInput if the
  * content is equal.
  */
-public class FBTypeEditorInput implements IFileEditorInput {
-	private FBType fbType;
+public class TypeEditorInput implements IFileEditorInput, IPersistableElement {
```
mx990, line 33: Should we also implement `IPathEditorInput` and `IURIEditorInput`, similar to `FileEditorInput`?

plugins/org.eclipse.fordiac.ide.attributetypeeditor/src/org/eclipse/fordiac/ide/attributetypeeditor/properties/AttributeTargetSection.java
```diff
@@ -103,8 +104,8 @@ private static StructuredType getTarget(final AttributeDeclaration attributeDecl
 	@Override
 	public void setInput(final IWorkbenchPart part, final ISelection selection) {
 		super.setInput(part, selection);
-		if (part instanceof final AttributeTypeEditor editor) {
-			setType(editor.getEditedElement());
+		if (part instanceof final AttributeTypeEditorPage editor) {
```
mx990: The part is still an `AttributeTypeEditor` and not an `AttributeTypeEditorPage` here. Also, selecting one of the checkboxes currently results in a NPE.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/AbstractTypeEditor.java
```diff
@@ -0,0 +1,458 @@
...
+
+	@Override
+	public void init(final IEditorSite site, final IEditorInput editorInput) throws PartInitException {
+		final TypeEditorInput typeEditorInput = checkEditorInput(editorInput);
+
+		if (isValidTypeEditorInput(typeEditorInput)) {
+			typeEditorInput.getTypeEntry().eAdapters().add(adapter);
+			annotationModel = new FordiacMarkerGraphicalAnnotationModel(typeEditorInput.getTypeEntry().getFile());
+			validationJob = new ValidationJob(getPartName(), getCommandStack(), annotationModel);
+		}
```
mx990: I believe this should probably be moved to `setInput` below (also dispose old model and job, if present), in order to also handle later changes to the input?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.attributetypeeditor/src/org/eclipse/fordiac/ide/attributetypeeditor/properties/AttributeTargetSection.java
azoitl: Thx for noting. Eclipse refactoring shadowed this change a bit. Fixed now.

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/TypeEditorInput.java
azoitl: I'm not sure. The uri implementation is straight forward. The path not so much and I'm at that point not sure if we need it.

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/AbstractTypeEditor.java
azoitl: setInput is only called by us or by the TypeEntryAdapter in case of a new file. But I guess we need to do something in reload file, or?

mx990: I am concerned about the case when we get a different file via the new input, precisely coming from the `TypeEntryAdapter` after changes to the file of the type entry. In this case, the annotation model will still point to the old file.

I believe `reloadFile()` should be fine as is. As long as the file stays the same, the flushing of the command stack will reload the markers in the annotation model via the command stack listener in the validation job.

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/TypeEditorInput.java
mx990: The path implementation should also be straightforward (i.e., `file.getLocation()`) if we discount the use of non-local files. However, I guess we probably don't actually need either at the moment.

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/AbstractTypeEditor.java
azoitl: @mx990 added code that will update the file of the annotation model. I hope this meets the goals.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/ResourceMarkerGraphicalAnnotationModel.java
```diff
@@ -137,6 +137,13 @@ public long getModificationStamp() {
 		return resource.getModificationStamp();
 	}
 
+	public final void setResource(final IResource resource) {
+		if (this.resource != resource) {
+			this.resource = resource;
+			reload();
```
mx990: I considered this, but I am a bit worried that this might cause race conditions when there is a concurrent update from the `resourceChangeListener`, which is quite likely since this is also called when moving resources around. There may also be stale non-marker annotations left over afterwards, since a `reload()` intentionally only clears marker annotations.

Our textual relative `org.eclipse.ui.texteditor.ResourceMarkerAnnotationModel` also does not change the resource and instead uses a new annotation model for a when the resource changes.

However, I can also do that in a separate PR. Sorry for not getting around to this sooner.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/ResourceMarkerGraphicalAnnotationModel.java
azoitl: I mostly wanted to learn more about the annotation model. so no problem at all. I already expected taht my solution was to simple.

mx990 (approved):

Issue #562, opened 2024-09-27
Adapter broken in 4diac-ide_3.0.0-win32.win32.x86_64_nightly_2024-09-27_0756


diplfranzhoepfinger:
Adapter broken in 4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-27\_0756

Composite FB with Adapter: Broken

Basic FB with Adapter: broken.

see my Training Project here:

[https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac_training1/releases/tag/v0.12](https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac_training1/releases/tag/v0.12)

open it with 

*   4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-24\_0544
    *   no issue
*   4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-27\_0756
    *   get Errors
*   4diac-ide\_3.0.0-win32.win32.x86\_64\_milestone\_2024-09-28\_0008
    *   no issue
*   4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-29\_0545
    *   no issue

Error Log: (shrinked)

```
Description    Resource    Path    Location    Type
The variable DT is undefined    sequence_ET_04.fbt    /test/Type Library/utils_plus/sequence-control/combi    line 8    4diac IDE Structured Text Algorithm Problem
```

azoitl:
This is a duplicated to #564. And in #564 the description of the problem is such that we could trace the issue down.

diplfranzhoepfinger:
completed.

Issue #564, opened 2024-09-28
plugs do not seam to be added to the CFB network


moniwe:
I have a project with lots of errors, concerning connections, double clicking on any of these errors did not move me to any screen, I therefore could not figure out how to resolve them and just accepted them since it deployed and worked. I migrated 4diac-ide to the newest version (Version: 3.0.0, Build id: 2024-09-27_0756) and reopened the project, than double clicking on any of the errors (there were other error now, former network connections now some else) lead me to this

![grafik](https://github.com/user-attachments/assets/d0ea54c9-96e9-4805-a096-0b53ef9d3beb)

I than opened this corresponding function block, and found this:

![grafik](https://github.com/user-attachments/assets/fac842e2-4bfd-4632-8e16-c1b7add755d5)

unfortunately I was not able to fix it. I tried removing the existing interface of this function block, which worked, but adding a new one resulted in a red marked interface without adding the corresponding plug to the network. I also tried doing it directly within the XML and restarting 4diac-ide, but it did not accept it.

I have a similar function block with a socket, which does not have this problem.

maybe related to #562

mx990:
The *Type not set!* on the FB instance (to the right in both screenshots) indicates that the type either was not properly set or could not be loaded. This, of course, means that any changes made to the original type are not reflected on the instance. I am currently working on a fix to also show an error on the FB instance in such cases.

Could you please check if the FB instance has a valid `Type` attribute in the XML or if there are any exceptions in the log? In the meantime, I would suggest to set the type manually using *right click->Change Type* on the FB instance to see if that resolves the problem.

azoitl:
I just did test on freeze and develop, it seems that it is a regression introduced in develop.

diplfranzhoepfinger:
> maybe related to #562

yes, very same what i see. 

only on Develop. 

only with the named Version. i will check with latest nightly as well. (seems 4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-27\_0756 is latest currently)

*   4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-24\_0544
    *   no issue
*   4diac-ide\_3.0.0-win32.win32.x86\_64\_nightly\_2024-09-27\_0756
    *   **get Errors**
*   4diac-ide\_3.0.0-win32.win32.x86\_64\_milestone\_2024-09-28\_0008
    *   no issue

azoitl:
[...]

PR #566, opened 2024-09-28
Fix type entry subclasses

mx990:
Fix type entry subclasses after https://github.com/eclipse-4diac/4diac-ide/pull/545:
- plug type creation in adapter entry,
- type class caching in FB type entry,
- remove extra synchronized in subclasses that could still cause deadlocks.

Fixes https://github.com/eclipse-4diac/4diac-ide/issues/562, https://github.com/eclipse-4diac/4diac-ide/issues/564

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AdapterTypeEntryImpl.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AdapterTypeEntryImpl.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Stream;
 
+import org.eclipse.emf.common.notify.NotificationChain;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.model.dataexport.AbstractTypeExporter;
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AdapterTypeEntryImpl.java
```diff
@@ -54,12 +55,23 @@ public EClass getTypeEClass() {
...
-			// wenn we get a new type ensure that the plug is the mirror of the socket
+			// ensure that the plug is the mirror of the socket
 			adpType.setPlugType(createPlugType(adpType));
 		}
-		super.setType(type);
+		return super.basicSetType(type, notifications);
+	}
+
+	@Override
+	protected synchronized NotificationChain basicSetTypeEditable(final LibraryElement type,
```
azoitl: For the editable type we don't need the plug and as we want to get rid of it I would prefer to not add it anymore.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AdapterTypeEntryImpl.java
mx990: Makes sense, removed.

azoitl (approved):

PR #567, opened 2024-09-28
Adjusted Adapter Type Parsing to new TypeLib Synchronization

azoitl:
On setType the AdpaterType entry generated also the Plug representation. With the update this has to be done in the basicSetType.


azoitl:
This fixes: #564

azoitl:
Better fix is in #566.

PR #570, opened 2024-09-29
adds a Timestamp SIFB

moniwe:


franz-hoepfinger-4diac:
we need such things yes. 
but is it really good to have only ns ? maybe others need ms ? 
just thinking: 
http://www.oscat.de/de/component/jdownloads/send/2-oscat-basic/7-oscat-basic333-de.html?Itemid=0

here is a T_PLC_MS and a T_PLC_US

but first **good** to have somebody to start this.

franz-hoepfinger-4diac:
other Topic: 

normally we have 2 Times: 

a Tick, which is counting from System Start, 
and a RTC-Clock, which can be set during Startup by GPS, by a NTP Server, by a Mobile Station. 

different Timers. 
the FB should be clearly marked which one is it. 

for the second one we will also need conversion Funtions like in this Example: https://forge.codesys.com/forge/talk/Runtime/thread/02c4655a4d/ (ST Code)

or here: https://github.com/espressif/esp-idf/blob/46acfdce969f03c02b001fe4d24fa9e98f6adc5e/examples/protocols/sntp/main/sntp_example_main.c#L102 (C-Code)

moniwe:
basically I wrote this FB to write IO changes into a InfluxDB 2.0. This DB wants timestamps in ns.

azoitl:
@franz-hoepfinger-4diac  what you want is already available with the `NOW` and `NOW_MONOTONIC` functions. Use these for your use cases.

diplfranzhoepfinger:
> @franz-hoepfinger-4diac what you want is already available with the `NOW` and `NOW_MONOTONIC` functions. Use these for your use cases.

where are they ? seems they are not in the typelibrary Folder ? 

only available in ST ? need to make avalialbe as Function ?

diplfranzhoepfinger:
> @franz-hoepfinger-4diac what you want is already available with the `NOW` and `NOW_MONOTONIC` functions. Use these for your use cases.

where are they ? seems they are not in the typelibrary Folder ? 

only available in ST ? need to make avalialbe as Function ?

diplfranzhoepfinger:
> basically I wrote this FB to write IO changes into a InfluxDB 2.0. This DB wants timestamps in ns.

how about name it Timestamp\_ns then it is clear for any first-time users ? 

and it gives space for later Timestamp\_ms and whatever ?

moniwe:
yap Timestamp_ns might be a better name, will rename it ...

MartinMelikMerkumians:
> yap Timestamp_ns might be a better name, will rename it ...

I would suggest Unix_Epoch_Timestamp_in_NS to make the functionality clear in the type name, as there are several start dates for timestamps. Currently this information is only visible in the description

mx990:
I believe you could also add an additional input to specify the start date, the default then being the Unix Epoch.

mx990 (approved):

MartinMelikMerkumians (approved):

azoitl (approved):

PR #578, opened 2024-10-01
Changed UpdateFBInstanceChange to AbstractCommandChange

oberlehner:
Not using the abstract command change in the udpate FB instance change resulted in eContainer == null errors, since it was not guaranteed that the correct eObject has been used during the refacotoring.
Furthermore did some clean up.


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/UpdateFBInstanceChange.java
```diff
@@ -28,75 +27,57 @@
...
-		return new UpdateFBInstanceChange(instance, typeEntry);
-	}
 
-	@Override
-	public Object getModifiedElement() {
-		return null;
+		if (element instanceof final StructManipulator muxer
+				&& typeEntry.getType() instanceof final StructuredType structuredType) {
+			return new ChangeStructCommand(muxer, structuredType);
+		}
```
azoitl, line 77: I think it could also be a configureable function block (e.g., F_MOVE) then you have to perform the update command on this. As structManipulator is a child of configureableFB the check has to be performed afterwards.

oberlehner: Should we make a generic command for that ?
I think it is very error-prone when we have to do that on various places.
I think of something like making the constructor private and make a factory method which creates the correct command.
What do you think ?

pazi146 (approved):

PR #579, opened 2024-10-01
FB Network Grid Now Follows IEC 61499 Coordinate System Resolution

azoitl:
With this change the grid resolution can not be changed anymore but is the same size as the difference between to pins of a FB. This results in cleaner drawings and with snap to grid positions of FBs are always in 100 IEC 61499-2 units. To get that also in expanded subapps and groups the top part of them is adjust to use a multiple of the grid size.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/GroupFigure.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/GroupFigure.java
@@ -20,8 +20,10 @@ import org.eclipse.draw2d.Graphics;
 import org.eclipse.draw2d.GridData;
 import org.eclipse.draw2d.GridLayout;
 import org.eclipse.draw2d.IFigure;
+import org.eclipse.draw2d.MarginBorder;
 import org.eclipse.draw2d.PositionConstants;
 import org.eclipse.draw2d.RoundedRectangle;
+import org.eclipse.draw2d.ToolbarLayout;
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.geometry.Insets;
 import org.eclipse.draw2d.geometry.Rectangle;
@@ -29,6 +31,7 @@ import org.eclipse.fordiac.ide.gef.draw2d.AdvancedLineBorder;
 import org.eclipse.fordiac.ide.gef.figures.AbstractShadowBorder;
 import org.eclipse.fordiac.ide.gef.figures.BorderedRoundedRectangle;
 import org.eclipse.fordiac.ide.gef.preferences.DiagramPreferences;
+import org.eclipse.fordiac.ide.model.CoordinateConverter;
 
 public class GroupFigure extends Figure {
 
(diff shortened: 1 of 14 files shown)


franz-hoepfinger-4diac:
> With this change the grid resolution can not be changed anymore but is the same size as the difference between to pins of a FB. This results in cleaner drawings and with snap to grid positions of FBs are always in 100 IEC 61499-2 units. To get that also in expanded subapps and groups the top part of them is adjust to use a multiple of the grid size.

VERY GOOD !!! long awaited ! Big support from my side.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/SubAppForFbNetworkFigure.java
```diff
@@ -247,15 +249,22 @@ private void addComment() {
...
 
 		commentFigure = new InstanceCommentFigure();
-		commentFigure.setBorder(new AdvancedLineBorder(PositionConstants.SOUTH));
+		final AdvancedLineBorder commentLineSeperator = new AdvancedLineBorder(PositionConstants.SOUTH);
+		commentFigure.setBorder(commentLineSeperator);
 		commentFigure.setCursor(Cursors.SIZEALL);
 		commentContainer.add(commentFigure);
 		refreshComment();
+
+		final int lineHeight = (int) CoordinateConverter.INSTANCE.getLineHeight();
```
oberlehner, line 273: This code looks similar to GroupFigure

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/GroupFigure.java
```diff
@@ -68,16 +70,32 @@ private void createMainFigure() {
...
+	private IFigure createCommentFigure() {
 		commentFigure = new InstanceCommentFigure();
 		commentFigure.setCursor(Cursors.SIZEALL);
 		final AdvancedLineBorder commentBorder = new AdvancedLineBorder(PositionConstants.SOUTH);
 		commentFigure.setBorder(commentBorder);
+
+		final Figure commentContainer = new Figure();
+		commentContainer.setLayoutManager(new ToolbarLayout());
+
+		final int lineHeight = (int) CoordinateConverter.INSTANCE.getLineHeight();
```
oberlehner, line 89: what is this doing ? 
Why dont you use top = Math.floor(lineHeight) ?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/SubAppForFbNetworkFigure.java
azoitl: Yes it is similar, but that code need to be adjusted to the individual figure. As the group looks different to subapp and may also change differently I would like to keep that separated.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/figures/GroupFigure.java
azoitl: This code is applying half of the line height above and half of it below the comment text. As we may have an uneven line height we need to compensate for rounding errors. Putting all above the comment would result in a strange look.

oberlehner (approved):

franz-hoepfinger-4diac:
VERY VERY VERY GOOD !!!

this is so cool !!!

image

franz-hoepfinger-4diac:
no more decimals, but fine integer Numbers. So so so good.

franz-hoepfinger-4diac:
i tested, it is VERY COOL ! !!!!!!

PR #582, opened 2024-10-01
Update GitHub Workflow Actions to Their Latest Versions

azoitl:
Dependabot is rather picky with version names, also using the commit hash instead of the version is better for reducing supply-chain attacks. See https://github.com/eclipse/gef-classic/issues/471 for more details.


azoitl:
I think this is not building because release is currently in itself not building. But @MartinMelikMerkumians I think it should still be merged to have better and newer versions of our github actions.

oberlehner (approved):

PR #588, opened 2024-10-03
Fixed issue where Quick-Fix option did not show up in dirty Editor

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -94,11 +96,26 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
-			if (!annotation.isEmpty()
-					&& annotation.iterator().next() instanceof final GraphicalMarkerAnnotation markerAnnotation) {
-				return markerAnnotation.getMarker();
+
+			final IMarker marker = getMarkerFromAnnotationModel(annotationModel, editPart.getModel());
+			if (marker != null) {
+				return marker;
 			}
+
+			annotationModel.reload();
```
mx990: I believe the main problem with this approach is that this may bring back annotations for stale markers that have already been (manually) fixed in the editor compared to the saved state. If the quickfix is canceled, for example, these stale marker annotations may then remain visible in the editor. It also does not allow to use quickfixes for new problems only present in the editor, which consequently do not have a marker yet.

I still believe the only way to handle this situation properly with the present infrastructure is to create temporary markers for the quickfix.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
sebHollersbacher: Correct me if I'm wrong, but as long as we don't save the editor or manually reload the Annotation model, it is not possible to get all Markers of the current state, right?
Or do you mean we create temporary markers for the whole editor using its typeEntry? In which case we would not use the Annotation Model at all.

mx990: > Correct me if I'm wrong, but as long as we don't save the editor or manually reload the Annotation model, it is not possible to get all Markers of the current state, right?

The only way to get up-to-date markers for the current state in the editor would be to save and then wait for the builder to complete its run. A reload of the annotation model would only bring back the stale markers from the last save.

> Or do you mean we create temporary markers for the whole editor using its typeEntry? In which case we would not use the Annotation Model at all.

My initial idea was to create temporary markers based on the annotation model for the selected elements in the editor. However, this would limit quickfixes to the current selection when performed from inside the editor.

After thinking about this some more, maybe it also does not really make sense to perform bulk quickfixes from inside the editor in the first place. The Java and Xtext editors also do not seem support this.

Maybe we should instead focus on offering a simpler workflow in the editor, where users can select individual fixes quickly from the context menu. This would leave the bulk quickfixes, which are more powerful but also more slow due to the extra dialog, to the *Problems* view.

sebHollersbacher: I've changed it now, so that only the markers of the selected editparts are used. Maybe we can find a better solution in the future.
@mx990 I'm not sure if the creation/deletion of the temporary markers is done correctly.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/errormarker/FordiacErrorMarker.java
```diff
@@ -48,9 +48,11 @@ public final class FordiacErrorMarker {
 	public static final String IMPORT_MARKER = "org.eclipse.fordiac.ide.model.import"; //$NON-NLS-1$
 	public static final String TYPE_LIBRARY_MARKER = "org.eclipse.fordiac.ide.model.typeLibrary"; //$NON-NLS-1$
 	public static final String LIBRARY_MARKER = "org.eclipse.fordiac.ide.model.library"; //$NON-NLS-1$
+	public static final String TEMPORARY_MARKER = "org.eclipse.fordiac.ide.model.temporary"; //$NON-NLS-1$
```
mx990, line 51: Please also define a marker type in the `plugin.xml`:
```xml
<extension
	id="temporary"
	name="4diac IDE Temporary Marker"
	point="org.eclipse.core.resources.markers">
	<super
		type="org.eclipse.core.resources.marker">
	</super>
	<persistent value="false" />
</extension>
```
Making it inherit directly from `org.eclipse.core.resources.marker` means it will not show up in the *Problems* view and making it non-persistent means any stale temporary markers will not survive a restart.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
 				.map(FordiacQuickFixHandler::getMarker).toArray(IMarker[]::new);
 	}
 
+	public static boolean hasMarker(final EditPart editPart) {
+		if (FordiacAnnotationUtil
+				.getAnnotationModel(editPart) instanceof final FordiacMarkerGraphicalAnnotationModel annotationModel) {
+			final var annotation = annotationModel.getAnnotations(editPart.getModel());
+			return !annotation.isEmpty();
+		}
```
mx990:
```suggestion
		final GraphicalAnnotationModel annotationModel = FordiacAnnotationUtil.getAnnotationModel(editPart);
		if (annotationModel != null) {
			return annotationModel.hasAnnotation(editPart.getModel(),
					annotation -> annotation instanceof GraphicalMarkerAnnotation
							|| annotation instanceof GraphicalValidationAnnotation);
		}
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+		return false;
+	}
+
 	public static IMarker getMarker(final EditPart editPart) {
 		if (FordiacAnnotationUtil
 				.getAnnotationModel(editPart) instanceof final FordiacMarkerGraphicalAnnotationModel annotationModel) {
```
mx990: There should be no need for the instanceof check here.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
 			final var annotation = annotationModel.getAnnotations(editPart.getModel());
+
 			if (!annotation.isEmpty()
 					&& annotation.iterator().next() instanceof final GraphicalMarkerAnnotation markerAnnotation) {
```
mx990: What if the first annotation is not a marker annotation? I would suggest using https://github.com/eclipse-4diac/4diac-ide/blob/c4b326463d87c6a02d05fdeeee0cbef7d0070bce/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java#L123-L126 from below to find a marker or validation annotation to fix.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+				@Override
+				protected void execute(final IProgressMonitor monitor)
+						throws CoreException, InvocationTargetException, InterruptedException {
+					ErrorMarkerBuilder.createErrorMarkerBuilder(validationAnnotation.getText())
+							.addAdditionalAttributes(validationAnnotation.getAttributes())
+							.setType(FordiacErrorMarker.TEMPORARY_MARKER).createMarker(resource);
+				}
+			}.run(new NullProgressMonitor());
+		} catch (final InvocationTargetException | InterruptedException e) {
+			e.printStackTrace();
```
mx990: Please use `FordiacLogHelper.logError(...)` instead.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+								marker.delete();
+							}
+						} catch (final CoreException e) {
+							e.printStackTrace();
+						}
+					});
+				}
+			}.run(new NullProgressMonitor());
+		} catch (final InvocationTargetException | InterruptedException e) {
+			e.printStackTrace();
```
mx990: Please use `FordiacLogHelper.logError(...)` instead.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+				@Override
+				protected void execute(final IProgressMonitor monitor)
+						throws CoreException, InvocationTargetException, InterruptedException {
+					Arrays.stream(selectedMarkers).forEach(marker -> {
+						try {
+							if (marker.getType().equals(FordiacErrorMarker.TEMPORARY_MARKER)) {
+								marker.delete();
+							}
+						} catch (final CoreException e) {
+							e.printStackTrace();
```
mx990: Please use `FordiacLogHelper.logError(...)` instead.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+							.addAdditionalAttributes(validationAnnotation.getAttributes())
+							.setType(FordiacErrorMarker.TEMPORARY_MARKER).createMarker(resource);
+				}
+			}.run(new NullProgressMonitor());
+		} catch (final InvocationTargetException | InterruptedException e) {
+			e.printStackTrace();
+		}
+	}
+
+	private static void deleteTemporaryMarker(final IMarker[] selectedMarkers) {
```
mx990, line 158: I would suggest filtering for temporary markers at the beginning. If there are none, you can just return early.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
+		try {
+			new WorkspaceModifyOperation() {
```
mx990: Please only lock the current file by passing it to the `WorkspaceModifyOperation(ISchedulingRule)` constructor.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
```diff
@@ -91,23 +99,76 @@ public static IMarker[] getMarkersFromEditorSelection(final IStructuredSelection
...
 	}
 
-	private static IMarker[] getAllMarker(final IResource resource) {
+	private static void createTemporaryMarker(final GraphicalValidationAnnotation validationAnnotation,
+			final IResource resource) {
 		try {
-			return resource.findMarkers(IMarker.MARKER, true, IResource.DEPTH_INFINITE);
-		} catch (final CoreException e) {
-			return new IMarker[0];
+			new WorkspaceModifyOperation() {
```
mx990: Please only lock the current file by passing it to the `WorkspaceModifyOperation(ISchedulingRule)` constructor.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/errormarker/FordiacErrorMarker.java
sebHollersbacher: I thought this was not necessary, as the temporary markers are created if the quick-fix option selected and deleted after the quick-fix is done/cancelled

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FordiacQuickFixHandler.java
sebHollersbacher: We need the resource of the AnnotationModel for the temporary markers, so I changed it to ResourceMarkerGraphicalAnnotationModel

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/errormarker/FordiacErrorMarker.java
mx990: I checked and they are non-persistent by default, so I guess this is fine.

mx990 (approved):

azoitl (approved):

PR #589, opened 2024-10-03
streamlined layout infrastructure

lindhuber:
The block and connection layout now trigger the layout based on where the mouse is located at the moment of activation.
Expanded subapp    -> layout subapp and all nested expanded subapps
Editor or Application -> layout application only


oberlehner (approved):
works fine !

azoitl (approved):

PR #597, opened 2024-10-07
Draft: Implementation for LibraryElementHasher

FRoith:
This change will add hashing functionality to the IDE, and automatically adds computed hash values to the output of the Forte NG exporter.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
@@ -1,0 +1,524 @@
+package org.eclipse.fordiac.ide.export.forte_ng.st;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.emf.common.util.EList;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.export.ExportException;
+import org.eclipse.fordiac.ide.export.forte_ng.util.IHasher;
+import org.eclipse.fordiac.ide.model.data.DataType;
+import org.eclipse.fordiac.ide.model.libraryElement.AdapterConnection;
+import org.eclipse.fordiac.ide.model.libraryElement.AdapterDeclaration;
+import org.eclipse.fordiac.ide.model.libraryElement.Algorithm;
+import org.eclipse.fordiac.ide.model.libraryElement.BaseFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.BasicFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.CompositeFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.DataConnection;
+import org.eclipse.fordiac.ide.model.libraryElement.ECC;
+import org.eclipse.fordiac.ide.model.libraryElement.ECState;
+import org.eclipse.fordiac.ide.model.libraryElement.ECTransition;
+import org.eclipse.fordiac.ide.model.libraryElement.ErrorMarkerFBNElement;
+import org.eclipse.fordiac.ide.model.libraryElement.ErrorMarkerInterface;
+import org.eclipse.fordiac.ide.model.libraryElement.Event;
+import org.eclipse.fordiac.ide.model.libraryElement.EventConnection;
+import org.eclipse.fordiac.ide.model.libraryElement.FB;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetwork;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetworkElement;
+import org.eclipse.fordiac.ide.model.libraryElement.FBType;
+import org.eclipse.fordiac.ide.model.libraryElement.FunctionBody;
+import org.eclipse.fordiac.ide.model.libraryElement.FunctionFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.Group;
+import org.eclipse.fordiac.ide.model.libraryElement.IInterfaceElement;
+import org.eclipse.fordiac.ide.model.libraryElement.INamedElement;
+import org.eclipse.fordiac.ide.model.libraryElement.InterfaceList;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.model.libraryElement.Method;
+import org.eclipse.fordiac.ide.model.libraryElement.OtherAlgorithm;
+import org.eclipse.fordiac.ide.model.libraryElement.OtherMethod;
+import org.eclipse.fordiac.ide.model.libraryElement.STAlgorithm;
+import org.eclipse.fordiac.ide.model.libraryElement.STMethod;
+import org.eclipse.fordiac.ide.model.libraryElement.ServiceInterfaceFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.SimpleFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.SubApp;
+import org.eclipse.fordiac.ide.model.libraryElement.SubAppType;
+import org.eclipse.fordiac.ide.model.libraryElement.Value;
+import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
+import org.eclipse.fordiac.ide.structuredtextalgorithm.util.StructuredTextParseUtil;
+
+public final class LibraryElementHasher implements IHasher {
+	String DEFAULT_HASH_ALG = "SHA3-512"; //$NON-NLS-1$
+	String HASH_VERSION = "v1"; //$NON-NLS-1$ // increment this whenever anything is changed in the hashing
+								// algorithm
+	String FULL_HASH_STRING = HASH_VERSION + ":" + DEFAULT_HASH_ALG; //$NON-NLS-1$
+
+	@Override
+	public String getHashType() {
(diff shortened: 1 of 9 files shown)


plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+			parts.add("DATAINPUTS{"); //$NON-NLS-1$
+			for (final VarDeclaration vdec : il.getVisibleInputVars()) { // TODO VisibleInputVars == DataInputs?
+				parts.add(String.format("%s:%s", vdec.getName(), vdec.getTypeName())); //$NON-NLS-1$
+			}
+			parts.add("}DATAINPUTS"); //$NON-NLS-1$
+		}
+		if (il.getVisibleOutputVars().size() > 0) {
+			parts.add("DATAOUTPUTS{"); //$NON-NLS-1$
+			for (final VarDeclaration vdec : il.getVisibleOutputVars()) { // TODO getVisibleOutputVars == DataOutputs?
+				parts.add(String.format("%s:%s", vdec.getName(), vdec.getTypeName())); //$NON-NLS-1$
```
azoitl, line 91: Why are you not using your vardecl method here?

plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+			parts.add("OUTPUTEVENTS{"); //$NON-NLS-1$
+			for (final Event event : il.getEventOutputs()) {
+				parts.add(String.format("%s:%s", event.getName(), event.getTypeName())); //$NON-NLS-1$
+			}
+			parts.add("}OUTPUTEVENTS"); //$NON-NLS-1$
+		}
+		if (il.getVisibleInputVars().size() > 0) {
+			parts.add("DATAINPUTS{"); //$NON-NLS-1$
+			for (final VarDeclaration vdec : il.getVisibleInputVars()) { // TODO VisibleInputVars == DataInputs?
+				parts.add(String.format("%s:%s", vdec.getName(), vdec.getTypeName())); //$NON-NLS-1$
```
azoitl, line 84: Why are you not using your vardecl method here?

plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+			parts.add("INPUTEVENTS{"); //$NON-NLS-1$
+			for (final Event event : il.getEventInputs()) {
+				parts.add(String.format("%s:%s", event.getName(), event.getTypeName())); //$NON-NLS-1$
+			}
+			parts.add("}INPUTEVENTS"); //$NON-NLS-1$
+		}
+		if (il.getEventOutputs().size() > 0) {
+			parts.add("OUTPUTEVENTS{"); //$NON-NLS-1$
+			for (final Event event : il.getEventOutputs()) {
+				parts.add(String.format("%s:%s", event.getName(), event.getTypeName())); //$NON-NLS-1$
```
azoitl, line 77: Why are you not using your event method here?

plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+	public String getHashType() {
+		return FULL_HASH_STRING;
+	}
+
+	private void process(final InterfaceList il, final List<String> parts, final String version) {
+		parts.add("IINTERFACES{"); //$NON-NLS-1$
+		if (il.getEventInputs().size() > 0) {
+			parts.add("INPUTEVENTS{"); //$NON-NLS-1$
+			for (final Event event : il.getEventInputs()) {
+				parts.add(String.format("%s:%s", event.getName(), event.getTypeName())); //$NON-NLS-1$
```
azoitl, line 70: Why are you not using your event method here?

plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+		}
+	}
+
+	private void process(final Event event, final List<String> parts, final String version) { // TODO add width
+		parts.add(String.format("EVENT:%s", event.getQualifiedName())); //$NON-NLS-1$
+	}
+
+	private void process(final VarDeclaration vdec, final List<String> parts, final String version) {
+		// TODO handle LocalVariable or MemberVarDeclaration
+		parts.add(String.format("VDEC:%s:%s", vdec.getName(), //$NON-NLS-1$
```
azoitl, line 373: data type name is missing

plugins/org.eclipse.fordiac.ide.export.forte_ng.st/src/org/eclipse/fordiac/ide/export/forte_ng/st/LibraryElementHasher.java
```diff
@@ -0,0 +1,524 @@
...
+	private void process(final AdapterConnection aconn, final List<String> parts, final String version) {
+		parts.add(String.format("%s--AdapterConn-->%s", aconn.getSource().getQualifiedName(), //$NON-NLS-1$
+				aconn.getDestination().getQualifiedName()));
+	}
+
+	private void process(final ECC ecc, final List<String> parts, final String version) {
+		parts.add(String.format("ECC{")); //$NON-NLS-1$
+		parts.add(String.format("ECStates{")); //$NON-NLS-1$
+		for (final ECState ecs : ecc.getECState()) {
+			parts.add(ecs.getName());
```
azoitl, line 396: actions are missing

azoitl (changes requested):
The IHasherInterface, the extension point, and the helper for getting a hasher should be moved to the model plugin, best to the typelibrary. Because we will needed it for different use cases. I would even expect that the TypeEntry will cash the hash values as they are rather expensive to create.

Please also add the copyright headers to the new files. This is the reason the build didn't run through.

mx990:
Aside from what @azoitl has already mentioned, I have a few general remarks regarding your current approach:

- You are hashing ST algorithms and methods by generating C++ code and using that for your hash. This means your hash:
    - is specific to the forte_ng export,
    - is sensitive to any future changes in the exported C++ code,
    - is sensitive to values of global constants in array ranges, which also makes it difficult to cache,
    - requires you to fully parse the ST code *and* generate C++ code, which makes it particularly expensive.
    
    I believe your hash is also currently sensitive to any added or removed lines in the ST code, since the generated C++ code contains `#line` directives (with the line number from the ST code). You would need to suppress them, for example, by overriding `generateLineDirective(EObject)` to return an empty string. Of course, that is easily done in this specific case, but something similar may come up again in the future and break your hash.
- You are explicitly iterating over all relevant model elements for your hash. I believe this is quite fragile and error-prone, since this makes it easy to accidentially omit relevant information. This is particularly problematic when someone else may extend the model in the future, without realising the need to extend the hash alongside it.

I would thus like to propose a different approach:
- use an XMI or binary EMF resource to generate the input for hashing,
- use the plain ST code, stripped of any extra whitespace, pragmas, and comments, for the hash,
- use meta-model annotations to specify features to be *excluded* from the hash.

I believe this has the following advantages over the current approach:
- all features are included in the hash by default,
- based on a known format instead of running our own,
- exceptions can be specified declaratively on the meta-model,
- follows any future changes to the model without significant additional work,
- does not require parsing of the ST code.

azoitl:
@mx990 how is your xmi based approach not breaking any existing XMI export?

mx990:
> @mx990 how is your xmi based approach not breaking any existing XMI export?

I believe the hashing implementation should be entirely independent of the XMI export, although it would certainly be quite similar in how it works. I would propose the following:
- create custom annotations for the EMF meta-model [1] to:
  - exclude a feature from hashing (e.g., for comments, etc.),
  - specify a class for pre-processing elements (e.g., for ST algorithms to strip comments, etc.)
- create a custom `EcoreUtil.Copier` that:
  - skips excluded features,
  - calls the pre-processor on copied elements if specified [2],
- create a generic hasher that works similar to the XMI export [3]:
  - copy the type (with the custom `Copier`),
  - add the result into a plain `org.eclipse.emf.ecore.xmi.impl.XMIResourceImpl`,
  - save the resource into an appropriate `OutputStream`,
  - use the result for hashing.

[1] see `org.eclipse.fordiac.ide.model.emf.FordiacMetaData`
[2] see also `org.eclipse.fordiac.ide.structuredtextcore.resource.LibraryElementXtextResource.ShallowCopier`
[3] see `org.eclipse.fordiac.ide.export.xmi.XMIExportFilter`

FRoith:
I do really like the approach, I'll look into implementing it that way, and see if I hit any snags on the way. The only potential issue I see, is that I don't yet have an idea of how I would introduce detecting "old versions" of the hash with this approach, as that was one of the requested features, but I'll see what I can find.

oberlehner:
Please add unit tests.
This is really important, because we want to detect early if any implementation inside the IDE (e.g change in the parser) might potentially changes the hash values.

So please add test cases for every type that can hashed:
-SubappType
-SimpleType
-CFB
-SIFB

Are you hashing Structured Data Type ? i cannot find that in the code

azoitl:
This has been superseeded by other work.

PR #603, opened 2024-10-08
Change system explorer navigation contents

pazi146:
split off content providers for automation systems


plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/providers/AutomationSystemContentProvider.java
```diff
@@ -0,0 +1,67 @@
...
+	public boolean hasChildren(final Object element) {
+		return SystemManager.isSystemFile(element);
+	}
+
+	@Override
+	public Object getParent(final Object object) {
+		if (object instanceof final IResource ires) {
+			return ires.getParent();
+		}
+		return super.getParent(object);
```
azoitl, line 47: not sure which of the provider is winning but we could get here EObjects as they are possible children. So to be on the save side we should check if the return value here is a Library element and return then the according IFile.

azoitl (changes requested):
To be consistent (and how I understood you also to be correct) the FBTypeContentProvider should now also only deliver the first level or and delegate the rest to the EObjectContentProvider or did I miss anything?

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/providers/EObjectContentProvider.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.fordiac.ide.model.libraryElement.Application;
+import org.eclipse.fordiac.ide.model.libraryElement.AutomationSystem;
+import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableObject;
+import org.eclipse.fordiac.ide.model.libraryElement.FBType;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementPackage;
+import org.eclipse.fordiac.ide.model.libraryElement.SystemConfiguration;
+import org.eclipse.fordiac.ide.model.libraryElement.TypedConfigureableObject;
+import org.eclipse.fordiac.ide.systemmanagement.SystemManager;
+
+public class EObjectContentProvider extends AdapterFactoryContentProvider {
```
azoitl: As this content provider is dedicated for system model content should it then have a better name. The content it can provide is defined by the adapter factories you are using.

pazi146: Removed this class as it turned out to be unnecessary

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/providers/AutomationSystemContentProvider.java
pazi146: contained in the new base class

azoitl (approved):

PR #608, opened 2024-10-09
Removed check for same type in MarkerResolutions

sebHollersbacher:
Removed type check to enable the quick-fix of different types

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractErrorMarkerResolution.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractErrorMarkerResolution.java
@@ -13,7 +13,6 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.marker.resolution;
 
-import java.util.Arrays;
 import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IMarker;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/AbstractErrorMarkerResolution.java
```diff
@@ -62,8 +61,7 @@ public Image getImage() {
 	public IMarker[] findOtherMarkers(final IMarker[] markers) {
 		return Stream.of(markers)
 				.filter(other -> LibraryElementValidator.DIAGNOSTIC_SOURCE.equals(FordiacErrorMarker.getSource(other))
-						&& FordiacErrorMarker.getCode(marker) == FordiacErrorMarker.getCode(other)
-						&& Arrays.equals(FordiacErrorMarker.getData(other), FordiacErrorMarker.getData(marker)))
+						&& FordiacErrorMarker.getCode(marker) == FordiacErrorMarker.getCode(other))
```
mx990, line 64: Why are you attempting to remove this check? I don't believe it makes much sense use a single quickfix for completely unrelated types, at least with the current quickfixes. The current implementation would either create a new type based on the first problem or choose a single type and then just apply it to all selected problems.

mx990 (changes requested):

PR #623, opened 2024-10-15
removed target branch from dependabot

azoitl:
Dependabot should update the workflows wich are currently run from release. So the target branch as develop was not the right place.


azoitl:
release branch build is currently always failing. Still merging for testing the updated dependabot script.

PR #630, opened 2024-10-15
Force SystemManager initialisation

pazi146:
ensure it isn't initialised through LibraryManager

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/Import4diacProject.java
+++ b/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/Import4diacProject.java
@@ -55,6 +55,10 @@ public class Import4diacProject extends Task {
 			workspace.setDescription(desc);
 			Job.getJobManager().cancel(null);
 
+			// ensure SystemManager is loaded
+			@SuppressWarnings("unused")
+			final SystemManager mgr = SystemManager.INSTANCE;
+
 			final IPath projectPath = getProjectPath();
 			final IProjectDescription description = loadProjectDescription(workspace, projectPath);
 			final IProject project = workspace.getRoot().getProject(description.getName());


plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/Import4diacProject.java
```diff
@@ -55,6 +55,10 @@ public void execute() throws BuildException {
 			workspace.setDescription(desc);
 			Job.getJobManager().cancel(null);
 
+			// ensure SystemManager is loaded
+			SystemManager.INSTANCE.removeFordiacChangeListener();
+			SystemManager.INSTANCE.addFordiacChangeListener();
```
ernstblechaPT:
```suggestion
			SystemManager.INSTANCE.hashCode(); // ensure SystemManager is instantiated
```

I think this should also ensure that the instance exists without the side-effect of switching the change listener off&on again

Also: based on the comments in SystemManager this change might make the call to `TypeLibraryManager.INSTANCE.getTypeLibrary(project).reload();` on line 71 redundant as instatiation of the TypeLib is also ensured by SystemManager; I'd leave that in for now and test later

azoitl: I would prefere:

`@SuppressWarnings("unused")
finalSystemManager mgr = SystemManager.INSTANCE; `

As we are doing it in in the activator of the org.eclipse.fordiac.ide plugin.  There it is sufficient.

pazi146: replaced the line with this

ernstblechaPT (approved):

azoitl (approved):

PR #631, opened 2024-10-15
Fix race condition during quickfixes and repair quickfixes for configurable FBs

mx990:
Fix race condition during quickfixes by refactoring marker resolutions to use proper locking of the workspace. Also combine commands for the same file and save only when necessary after applying changes. Further repair quickfixes for configurable FBs by adding validations to produce the necessary error markers.


sebHollersbacher (changes requested):
This change breaks the option for Creating a new FB Type.
I guess by locking the workspace, the new FB type can not be created correctly

mx990:
> This change breaks the option for Creating a new FB Type. I guess by locking the workspace, the new FB type can not be created correctly

You are right, thanks for noticing that. Should work now.

sebHollersbacher (approved):

azoitl (approved):

PR #644, opened 2024-10-18
Add new FBType RangeBasedPulse for Distance based Impulse Generator.

franz-hoepfinger-4diac:
Add new FBType RangeBasedPulse for Distance based Impulse Generator.

Typical usage: 

this would e.g. paint the middle dash-line on a Street: 
5000 Line, with 5000 Gap. 

image


franz-hoepfinger-4diac:
Make Sense to add a Distance Offset! 

image

azoitl (changes requested):
Please don't use the GEN_ prefix for such blocks. The GEN_ prefix is used in Eclipse 4diac for configurable types (e.g., publish). Also distance is maybe not the best name as I could feed it any value to generate pulses. 

Would `RangeBasedPulse` be better?

diplfranzhoepfinger:
> Please don't use the GEN\_ prefix for such blocks. The GEN\_ prefix is used in Eclipse 4diac for configurable types (e.g., publish).

sorry, yes, normally i know this. 

>  Also distance is maybe not the best name as I could feed it any value to generate pulses.

you are right. it might be Time, it might be Distance, or it might be any Value from any Source. 

e.g. generate Pulses dependent of a Sensor Value. 

> Would `RangeBasedPulse` be better?

yes, will change

franz-hoepfinger-4diac:
@azoitl changed

azoitl (approved):

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-ide/pull/760

PR #648, opened 2024-10-18
Add new integral function block for determining approximate integral over time.

franz-hoepfinger-4diac:


data/typelibrary/utils-1.0.0/typelib/math/INTEGRAL.fbt
```diff
@@ -0,0 +1,56 @@
...
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="CNF" Type="Event" Comment="Execution Confirmation">
+				<With Var="OUT"/>
+				<With Var="OVERFLOW"/>
+			</Event>
+		</EventOutputs>
+		<InputVars>
+			<VarDeclaration Name="IN" Type="REAL" Comment="Input value" InitialValue="0.0"/>
+			<VarDeclaration Name="TM" Type="DINT" Comment="Time since last call in msec" InitialValue="0"/>
```
ernstblechaPT, line 26: I would prefer TM to have type `TIME` to better indicate the meaning of this value. But I fully understand the implications this has on the integration as soon as people start to use values in nanoseconds...

data/typelibrary/utils-1.0.0/typelib/math/INTEGRAL.fbt
```diff
@@ -0,0 +1,56 @@
...
+		</OutputVars>
+	</InterfaceList>
+	<SimpleFB>
+		<Algorithm Name="REQ" Comment="">
+			<ST><![CDATA[ALGORITHM REQ
+IF NOT OVERFLOW THEN
+	IF (1.0E38 - IN * DINT_TO_REAL(TM) < OUT OR -1.0E38 - IN * DINT_TO_REAL(TM) > OUT) THEN
+		OVERFLOW := TRUE;
+	ELSIF (TM > 0) THEN    // Time Difference since last call must be positive.
+		OUT := OUT + IN * DINT_TO_REAL(TM) / 1000.0;
```
ernstblechaPT, line 40: When testing this in the Interpreter I got the following case:
 - send RESET
 - set IN to 9.0E9, TM to 1 send REQ; OUT goes to 8999999.0
 - set the input to 1.0, send REQ; OUT does not change
 - set the input to -1.0, send REQ; OUT also does not change

this is caused by the variable resolution of the `REAL` datatype - a warning for the case where the integration does not work (i.e. OUT does not change with IN <> 0) could be nice-to-have

franz-hoepfinger-4diac: you are totally right: 
our colleges at CODESYS have same issue with same Function. 
this comes from the internal representation of REAL.

maybe we should also make a LREAL Version, as well a Fixed-comma Version in DINT. 
i will think about it, and in the MEantime make the warning as written above.

MartinMelikMerkumians:
Seeing the comments and the implications for the correct operation of this Function Block, I don't think this should be written as a SFB in ST at all, but as a SiFB, which can do all the relevant checking and error signalling on the system level.
TM can then be a TIME, which a check if TM's value is lower than current clock resolution, and IN and OUT could be ANY_NUMs, eliminating the need to define multiple, functional identical versions of this FB

diplfranzhoepfinger:
@MartinMelikMerkumians i totally agree to you. 
in my opinion we should make a First Version with severe Warnings about REAL.
 
and then step a big Moment back, go deep, take time, and make the improved Version with ANY_NUM, 
wich will imply even more Documentation and more testing. 

agree ?

azoitl:

PR #654, opened 2024-10-21
SWTBot tests on maven build

alliuqemal:


azoitl:
@alliuqemal could you move the second  commit into an own PR. I think this would be nicer from a review process. @Andrearium FYI.

PR #656, opened 2024-10-22
Prevent CreateSubAppCrossingConnectionsCommandfrom creating ANY subapp pins

oberlehner:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/CreateSubAppCrossingConnectionsCommand.java
```diff
@@ -121,11 +123,29 @@ public boolean canExecute() {
...
+				&& !IecTypes.GenericTypes.isAnyType(destination.getType())) {
+			return destination.getType();
+		}
+
+		if (!IecTypes.GenericTypes.isAnyType(source.getType())
+				&& IecTypes.GenericTypes.isAnyType(destination.getType())) {
+			return source.getType();
+		}
+
+		return null;
```
azoitl, line 146: I would return here the source type this avoids the null checks above.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/CreateSubAppCrossingConnectionsCommand.java
oberlehner: are you sure,
when we call buildPath with isRightPath=true, then we want the destination element.
i don't know if this has any side effects if i pass the source element to the "right" buildPath.

azoitl: This is a good question. The only difference would be if source and destination pin do not have the same type. If that is the case the connection can only be created if it is an implicit upcast. In these situations I also think using the source pin as the defining type is better as it is the more narrow one and more correct to the given requirements. But I may miss somthing here.

mx990: I believe we have several cases to consider:
1. if the source and destination types are equal, it does not matter,
2. if the destination type is assignable from the source type and neither is generic (simple upcast), we should create, as before:
    - on the left, output pins with the (narrower) source type, so they fit better into other destinations,
    - on the right, input pins with the (broader) destination type, so they fit better from other sources,
3. if either is generic, we should prefer the other, non-generic type, so we avoid generic subapp interfaces as much as possible,
4. if both are generic, we default to case 2.

What do you think?

oberlehner: > I believe we have several cases to consider:
> 
> 1. if the source and destination types are equal, it does not matter,
> 2. if the destination type is assignable from the source type and neither is generic (simple upcast), we should create, as before:
>    
>    * on the left, output pins with the (narrower) source type, so they fit better into other destinations,
>    * on the right, input pins with the (broader) destination type, so they fit better from other sources,
> 3. if either is generic, we should prefer the other, non-generic type, so we avoid generic subapp interfaces as much as possible,
> 4. if both are generic, we default to case 2.
> 
> What do you think?

thx for the explanation. This is exactly what this implementation should do.
In my opinion, this is also the most intuitive way. 
So i would leave it like it is, and we can change in the future if neccessary

azoitl: My main concern was that the code is now a bit more complex then I thought it should be. I tested it and it works as expected. Also it is nearer to the previous behavior. So I would merge it for now. Changing it would be rather straigt forward

azoitl (approved):

PR #667, opened 2024-10-23
Added new EventType (EInit)

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/TypeEditPart.java
```diff
@@ -250,8 +256,8 @@ public INamedElement getINamedElement() {
 	@Override
 	public boolean isDirectEditable() {
 		// allow direct edit only for VarDeclarations and AdapterDeclarations
```
azoitl: please update the comment

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editparts/TypeEditPart.java
```diff
@@ -250,8 +256,8 @@ public INamedElement getINamedElement() {
...
-		return super.isDirectEditable()
-				&& (getCastedModel() instanceof VarDeclaration || getCastedModel() instanceof AdapterDeclaration);
+		return super.isDirectEditable() && (getCastedModel() instanceof VarDeclaration
```
azoitl: I think now the `&& ( ... )` can be removed as a whole as it covers all interface elements.

azoitl (changes requested):

oberlehner:
image
when an event is connected it cannot be changed.
This was an earlier feature, to prevent broken applications. But now the validation is mature enough to deal with these kind of errors, therefore please remove this check so that the event type can be changed here

sebHollersbacher:
@oberlehner Has been changed to not consider Connections

azoitl:
> @oberlehner Has been changed to not consider Connections

for me this is at least for the pin property sheet still broken.

oberlehner (approved):
Tested in UI and works fine !

azoitl (approved):

PR #669, opened 2024-10-23
Only visible Varinout pins are now added to the model

PStmmr:
This means that the model children were updated for visible varinout pins


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractFBNElementEditPart.java
```diff
@@ -456,6 +460,14 @@ protected List<Object> getModelChildren() {
 		final List<Object> elements = new ArrayList<>();
 		elements.add(getInstanceName());
 		elements.addAll(getModel().getInterface().getAllInterfaceElements());
+
```
oberlehner, line 474: please extract this in a method , like removeHiddenElements(...)

oberlehner (approved):

azoitl (approved):

PR #687, opened 2024-10-31
Remove Comment Value Escaping

azoitl:
Our XML writer is correctly escaping attribute values. If we are also escaping them we get double escaped attribute values which are then wrong.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
@@ -1,6 +1,7 @@
 /********************************************************************************
- * Copyright (c) 2008 - 2017 Profactor Gmbh, TU Wien ACIN, fortiss GmbH
- * 				 2018 - 2020 Johannes Kepler University, Linz
+ * Copyright (c) 2008 - 2024 Profactor Gmbh, TU Wien ACIN, fortiss GmbH,
+ *                           Johannes Kepler University, Linz,
+ *                           Primetals Technologies Austria GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


mx990:
Should we also write the `Description` attribute in `CommonElementExporter.addIdentification(LibraryElement)` using the normal `XMLStreamWriter.writeAttribute(...)` and then get rid of both `CommonElementExporter.fullyEscapeValue(...)` and `writeAttributeRaw(...)` entirely or is there a reason to do this differently there?

https://github.com/eclipse-4diac/4diac-ide/blob/b49ba72a02d15ad67c4f97d0053d34888702d38c/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java#L329-L331

azoitl:
> Should we also write the `Description` attribute in `CommonElementExporter.addIdentification(LibraryElement)` using the normal `XMLStreamWriter.writeAttribute(...)` and then get rid of both `CommonElementExporter.fullyEscapeValue(...)` and `writeAttributeRaw(...)` entirely or is there a reason to do this differently there?
> 
> https://github.com/eclipse-4diac/4diac-ide/blob/b49ba72a02d15ad67c4f97d0053d34888702d38c/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java#L329-L331

Thx for pointing this out. This brought me to the root cause of the underlying problem. `fullyEscapeValue(..)'  was originally developed for a different use case and is misused here. The main reason the raw method was used is to better handle newlines in attribute values. But this can be handled much better. Changed that now.

mx990:
I have tried this most recent change and it will now double escape `\n` to `&amp;#10;`, which will be read in again as `&#10;` instead of `\n`.

azoitl:
> I have tried this most recent change and it will now double escape `\n` to `&amp;#10;`, which will be read in again as `&#10;` instead of `\n`.

ah damn, ok now I know why we had the raw thing.

mx990 (approved):

PR #688, opened 2024-11-01
Remove unnecessary Attribute from F_MOVE.fbt

franz-hoepfinger-4diac:
The commit removes an unnecessary Attribute from the F_MOVE.fbt file.


azoitl:
This attribute is needed for F_MOVE. It would be nice if you would save our time by not doing things that are not needed.

franz-hoepfinger-4diac:
AH OK, i did search in Forte, and thouht it was done by Mistake.

PR #689, opened 2024-11-03
Update Type Preserves any Instance Attributes

azoitl:
Attributes on FB instances where only preserved for muxers when changing the struct. With this change the basic configuration of the updated instances is done centrally in the base class
AbstractUpdateFBNElementCommand. This reduces the code and simplifies maintenance.

Furthermore the not used and wrong command ChangeSubaAppPinCommand was removed.


bwiesmayr (approved):
I thought that we had fixed this a million times already :D but maybe it's now the final time

azoitl:
> I thought that we had fixed this a million times already :D but maybe it's now the final time

I really hope so.

PR #690, opened 2024-11-03
Fixed Exception in Pin Refactoring for non connection change

azoitl:
Selecting No Change for connection updates in pin rename refactorings resulted in an exception.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ReconnectPinChange.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ReconnectPinChange.java
@@ -50,8 +50,12 @@ public class ReconnectPinChange extends AbstractCommandChange<FBNetworkElement> 
 
 	@Override
 	protected Command createCommand(final FBNetworkElement element) {
+		if (state.size() == 1 && state.contains(ChangeState.NO_CHANGE)) {
+			return new Command() {
+				// empty command that is not doing any change
+			};
+		}
 		return new ReconnectPinByName(oldName, newName, element, state);
-
 	}
 
 	@Override


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ReconnectPinChange.java
```diff
@@ -50,8 +50,12 @@ public ReconnectPinChange(final URI elementURI, final Class<FBNetworkElement> el
 
 	@Override
 	protected Command createCommand(final FBNetworkElement element) {
+		if (state.size() == 1 && state.contains(ChangeState.NO_CHANGE)) {
```
oberlehner, line 53: I think this only fixes the symptom, not the cause, in my opinion this should be fixed in the AbstractCommandChange.

The problem is that if `canExecute `returns false, then 
`org.eclipse.fordiac.ide.typemanagement.refactoring.AbstractCommandChange.performCommand(T)` throws an exception.

so we can either not throw an exception in the abstractComandChange, or we try to propagate the "no_change" option to the first refacotoring wizard page (which is hard to inject) so we don't get into this problem.

@mx990 what do you think about this ?

mx990: I believe it is important to fail noticeably in case the command cannot be executed. We have had several instances in the past where a command was silently skipped due to a bug. This left users wondering why nothing happened even though they performed the refactoring successfully.

If we want to reconsider at the last moment and silently skip a change, I believe we should do so explicitly in the code, ideally before the type has been copied and without saving afterwards. A simple solution could be to just override `isEnabled()` and return false when `NO_CHANGE` was chosen.

As an alternative, how about we simply remove the *No Change* option entirely and have the user deselect the change in the preview instead?

oberlehner: I think this is a good idea.
i made PR for that:
https://github.com/eclipse-4diac/4diac-ide/pull/711
Would be great if you can have a look at it.

azoitl:
Closed in favor fo #711

PR #698, opened 2024-11-04
Show Attribute Property Sheet for Pin Type Sel. in Interface Editor

azoitl:
When the type field of a pin is selected in the interface editor the same property sheets as for the pin selection is shown.


oberlehner:
image
shouldtn it be here ? or did i understand the PR wrong

azoitl:
Yes it should. I uploaded the wrong file. Now its fixed.

PR #703, opened 2024-11-05
Fix copying of adapter FB in adapter declarations

mx990:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
```diff
@@ -178,6 +179,7 @@ private static void copyAdapterList(final EList<AdapterDeclaration> destAdapters
 	public static AdapterDeclaration copyAdapter(final AdapterDeclaration adapter, final boolean copyComments) {
 		final AdapterDeclaration copy = LibraryElementFactory.eINSTANCE.createAdapterDeclaration();
 		copyInterfaceElement(adapter, copy, copyComments);
+		copy.setAdapterFB(EcoreUtil.copy(adapter.getAdapterFB()));
```
azoitl, line 182: This only works if the source type is basic, simple or SIFB. For CFB Types you need to use the `getAdapterNetworkFB()` method. I think the most future prove solution would be to ask if getAdapterFB() is null and in that case use `getAdapterNetworkFB()`. This would then also work for the cases when we will only parse type interfaces for instances in the future.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
mx990: I wanted to avoid the case where we have a duplicate `AdapterFB` in the adapter declaration and network for CFB instances, so I am intentionally setting the adapter FB to `null` on the copy when the original had none, that is, for CFBs.

With https://github.com/eclipse-4diac/4diac-ide/pull/697, the CFB instance will still receive its own `adapterNetworkFB` when the inner FB network is loaded with `loadCFBNetwork()`.

> I think the most future prove solution would be to ask if getAdapterFB() is null and in that case use getAdapterNetworkFB().

This is what I currently do when I need the adapter FB. Maybe we could also think about making this more convenient by having a non-containment `adapterFB` reference in `AdapterDeclaration` that is always set and an additional `adapterFBContent` or so that is only for holding the `AdapterFB` for non-CFBs?

azoitl: This is what I currently do when I need the adapter FB. Maybe we could also think about making this more convenient by having a non-containment adapterFB reference in AdapterDeclaration that is always set and an additional adapterFBContent or so that is only for holding the AdapterFB for non-CFBs?

That could releave us from some of the ifs I have in some of the editors. However after thinking a bit more on it I think for now for debugging the current setup should be fine. 

Am I right that the duplicated Adapter FB instance that you get if you open the CFB viewer is not a problem for monitoring or?

mx990: How would I get a duplicate Adapter FB instance from opening the CFB instance viewer? I believe it should show the Adapter FB inside the `cfbNetwork` of the `CFBInstance`, which should be the same as the `adapterNetworkFB` in the `AdapterDeclaration`?

azoitl: Not with the current code. In oder to show the adapterFB inside of the fbnetwork it is a member of the network. So I think I need to rework the what I did in #697. Also with combination of your suggestion above.

azoitl: Based on our discussion here I did PR #709 I think this PR improves adapterFB handling and is a replacement for this PR.

PR #709, opened 2024-11-05
Reworked AdapterFB Handling

azoitl:
With this change the adapterFB reference of any adapter declaration is always providing the correct adapterFB independent if it is a CFB or other FBs.


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
```diff
@@ -178,6 +179,8 @@ private static void copyAdapterList(final EList<AdapterDeclaration> destAdapters
 	public static AdapterDeclaration copyAdapter(final AdapterDeclaration adapter, final boolean copyComments) {
 		final AdapterDeclaration copy = LibraryElementFactory.eINSTANCE.createAdapterDeclaration();
 		copyInterfaceElement(adapter, copy, copyComments);
+		copy.setAdapterFB(EcoreUtil.copy(adapter.getAdapterFB()));
+		copy.setInterfaceOnlyAdapterFB(copy.getAdapterFB());
```
mx990, line 184: I believe you need to either set the `adapterDecl` reference in the copied `AdapterFB` or better make it an opposite of `AdapterDeclaration.adapterFB`, which would save a lot of extra `setAdapterDecl(...)` calls?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
azoitl: Thx for noticing this. I like the idea with the opposite. However with the current implementation of the monitoring adapters this would break the model. If the work on monitoring that you are currently doing is changing that we can reconsider. For now I just added the setAdapterDecl.

mx990:
I have created a PR https://github.com/azoitl/4diac-ide/pull/5 with a small fix for the changed containment in this PR.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
mx990: You are right, I did not consider the current monitoring and the new monitoring is indeed changing that.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
azoitl: Then I would do that change after the new monitoring. I was already half way through and it was really a nice clean-up.

PR #711, opened 2024-11-06
Adapted "No Change" option in refactoring preview

oberlehner:
The manual creation of error markers has been removed from the changes.
The error markers need to be created from the builder.
This avoids duplicate creation of error markers.

When no change is the default option, the change will be disabled.
This means, when a repair operation should be applied to a affected instance, it is necessary to enable the change first by selecting it in the preview.


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AbstractCommandChange.java
```diff
@@ -66,6 +67,18 @@ public abstract class AbstractCommandChange<T extends EObject> extends Change {
...
+
+	/**
+	 * It needs to be checked whether a change has been set to "NO_CHANGE"
+	 *
+	 */
+	private void initEnablement() {
+		if ((this instanceof final IFordiacPreviewChange change)
+				&& change.getDefaultSelection().contains(ChangeState.NO_CHANGE)) {
+			setEnabled(false);
+		}
```
mx990: I believe we should avoid introducing behavior for specific subclasses in the base class. I suggest either doing this only in the relevant subclasses or maybe introduce a common abstract class for preview changes?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AbstractCommandChange.java
oberlehner: you are right, i will make the ChangePreviewer abstract and make it as super class of the AbstractCommandChange.
If a change does not need a preview that it should also be fine, since they first have to be defined in the plugin.xml

oberlehner: i meant the other way . the new abstract class needs to inherit AbstractCommandChange.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AbstractCommandChange.java
```diff
@@ -69,7 +69,9 @@ protected AbstractCommandChange(final URI elementURI, final Class<T> elementClas
 	}
 
 	/**
-	 * Create an abstract command change
+	 *
+	 *
+	 * /** Create an abstract command change
```
mx990: Please fix the comment back to how it was (or just restore from develop)

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AttributeValueChange.java
mx990, line 1: I suggest restoring this file from develop.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/delete/SafeFBTypeDeletionChange.java
```diff
@@ -85,7 +86,7 @@ public Change perform(final IProgressMonitor pm) throws CoreException {
 		}
 	}
 
-	private static class DeleteInternalFBChange extends CompositeChange implements IFordiacPreviewChange {
+	private static class DeleteInternalFBChange extends ViewableChange<EObject> {
 		private final EnumSet<ChangeState> state = EnumSet.noneOf(ChangeState.class);
```
mx990: Is this field not redundant now?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/delete/SafeFBTypeDeletionChange.java
oberlehner: This will be removed of course. I am not finished yet with this PR. Marked it as draft now ;-)
Refactoring the SafeFBTypeDeletionChange requires more work then expected, but its a good chance to restructure it, since i think the current implementation is not ideal

mx990 (approved):

azoitl (approved):

PR #714, opened 2024-11-06
Added Copy/Paste handler for Imports to NatTable

sebHollersbacher:
Added Copy and Paste handler to NatTable to correctly handle imports when copy/pasting from/into cells

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
@@ -92,7 +92,8 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE);
 		inputDataLayer.setConfigLabelAccumulator(new VarDeclarationConfigLabelAccumulator(outputProvider,
 				this::getAnnotationModel, VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE));
-		final var columnProvider = new NatTableColumnProvider<>(VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE);
+		final NatTableColumnProvider<VarDeclarationTableColumn> columnProvider = new NatTableColumnProvider<>(
+				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE);
 		outputTable = NatTableWidgetFactory
 				.createRowNatTable(outputsGroup, inputDataLayer, columnProvider,
 						new UntypedSubappInterfaceEditableRule(getSectionEditableRule(),
@@ -123,7 +124,7 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG);
 		inputDataLayer.setConfigLabelAccumulator(new VarDeclarationConfigLabelAccumulator(inputProvider,
 				this::getAnnotationModel, VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG));
-		final var columnProvider = new NatTableColumnProvider<>(
+		final NatTableColumnProvider<VarDeclarationTableColumn> columnProvider = new NatTableColumnProvider<>(
 				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG);
 		inputTable = NatTableWidgetFactory.createRowNatTable(inputsGroup, inputDataLayer, columnProvider,
 				new UntypedSubappInterfaceEditableRule(getSectionEditableRule(),
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
@@ -19,6 +19,7 @@
 package org.eclipse.fordiac.ide.application.properties;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
@@ -27,6 +28,7 @@ import org.eclipse.fordiac.ide.application.commands.CreateSubAppInterfaceElement
 import org.eclipse.fordiac.ide.application.commands.ResizeGroupOrSubappCommand;
 import org.eclipse.fordiac.ide.application.commands.ResizingSubappInterfaceCreationCommand;
 import org.eclipse.fordiac.ide.application.utilities.GetEditPartFromGraficalViewerHelper;
+import org.eclipse.fordiac.ide.gef.nat.CopyDataImportCommandHandler;
 import org.eclipse.fordiac.ide.gef.nat.InitialValueEditorConfiguration;
 import org.eclipse.fordiac.ide.gef.nat.TypeDeclarationEditorConfiguration;
 import org.eclipse.fordiac.ide.gef.nat.VarDeclarationColumnAccessor;
@@ -53,6 +55,7 @@ import org.eclipse.gef.commands.Command;
 import org.eclipse.nebula.widgets.nattable.config.IEditableRule;
 import org.eclipse.nebula.widgets.nattable.data.IRowDataProvider;
 import org.eclipse.nebula.widgets.nattable.layer.DataLayer;
+import org.eclipse.nebula.widgets.nattable.selection.SelectionLayer;
 import org.eclipse.swt.widgets.Group;
 
 public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterfaceDataSection {
@@ -120,8 +123,9 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG);
 		inputDataLayer.setConfigLabelAccumulator(new VarDeclarationConfigLabelAccumulator(inputProvider,
 				this::getAnnotationModel, VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG));
-		inputTable = NatTableWidgetFactory.createRowNatTable(inputsGroup, inputDataLayer,
-				new NatTableColumnProvider<>(VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG),
+		final var columnProvider = new NatTableColumnProvider<>(
+				VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG);
+		inputTable = NatTableWidgetFactory.createRowNatTable(inputsGroup, inputDataLayer, columnProvider,
 				new UntypedSubappInterfaceEditableRule(getSectionEditableRule(),
 						VarDeclarationTableColumn.DEFAULT_COLUMNS_WITH_VISIBLE_AND_VAR_CONFIG, inputProvider),
 				null, this, true);
@@ -129,6 +133,10 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 		inputTable.addConfiguration(new TypeDeclarationEditorConfiguration(inputProvider));
 		inputTable.addConfiguration(new CheckBoxConfigurationNebula());
 		inputTable.configure();
+
+		final SelectionLayer selectionLayer = NatTableWidgetFactory.getSelectionLayer(inputTable);
+		selectionLayer.registerCommandHandler(new CopyDataImportCommandHandler(selectionLayer, columnProvider,
+				Map.of(VarDeclarationTableColumn.TYPE, eObject -> ((VarDeclaration) eObject).getType())));
 	}
 
 	@Override
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
@@ -30,6 +30,7 @@ import org.eclipse.fordiac.ide.application.commands.ResizingSubappInterfaceCreat
 import org.eclipse.fordiac.ide.application.utilities.GetEditPartFromGraficalViewerHelper;
 import org.eclipse.fordiac.ide.gef.nat.CopyDataImportCommandHandler;
 import org.eclipse.fordiac.ide.gef.nat.InitialValueEditorConfiguration;
+import org.eclipse.fordiac.ide.gef.nat.PasteDataImportFromClipboardCommandHandler;
 import org.eclipse.fordiac.ide.gef.nat.TypeDeclarationEditorConfiguration;
 import org.eclipse.fordiac.ide.gef.nat.VarDeclarationColumnAccessor;
 import org.eclipse.fordiac.ide.gef.nat.VarDeclarationConfigLabelAccumulator;
@@ -137,6 +138,9 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 		final SelectionLayer selectionLayer = NatTableWidgetFactory.getSelectionLayer(inputTable);
 		selectionLayer.registerCommandHandler(new CopyDataImportCommandHandler(selectionLayer, columnProvider,
 				Map.of(VarDeclarationTableColumn.TYPE, eObject -> ((VarDeclaration) eObject).getType())));
+		selectionLayer.registerCommandHandler(
+				new PasteDataImportFromClipboardCommandHandler(selectionLayer, getCurrentCommandStack()));
+
 	}
 
 	@Override
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
@@ -102,8 +102,7 @@ public class EditUntypedSubappInterfaceDataSection extends AbstractEditInterface
 		outputTable.addConfiguration(new InitialValueEditorConfiguration(outputProvider));
 		outputTable.addConfiguration(new TypeDeclarationEditorConfiguration(outputProvider));
 		outputTable.addConfiguration(new CheckBoxConfigurationNebula());
-		outputTable.addConfiguration(
-				new DefaultImportCopyPasteLayerConfiguration(columnProvider, this::getCurrentCommandStack));
+		outputTable.addConfiguration(new DefaultImportCopyPasteLayerConfiguration(columnProvider, this));
 		outputTable.configure();
 	}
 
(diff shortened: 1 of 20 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappInterfaceDataSection.java
```diff
@@ -194,4 +203,24 @@ private static boolean isConnected(final VarDeclaration varDeclaration) {
...
+			final ListDataProvider<?> provb = (ListDataProvider<?>) NatTableWidgetFactory.getDataLayer(natTable)
+					.getDataProvider();
+			final SelectionLayer selectionLayer = NatTableWidgetFactory.getSelectionLayer(natTable);
+			if (selectionLayer != null && selectionLayer.getSelectedCells().size() == 1) {
+				final var cell = selectionLayer.getSelectedCells().iterator().next();
+				final var rowObj = provb.getRowObject(cell.getRowIndex());
+				if (cell.getColumnIndex() == 1 && rowObj instanceof final VarDeclaration decl) {
+					final Clipboard clipboard = new Clipboard(Display.getDefault());
+					clipboard.setContents(new String[] { PackageNameHelper.getFullTypeName(decl.getType()) },
+							new Transfer[] { TextTransfer.getInstance() });
```
mx990: I believe this will simply copy the full qualified name instead of adding an import when pasted into a type cell? I would suggest creating a separate transfer that also includes the required imports and then add them in the destination type if possible and necessary.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java
```diff
@@ -173,7 +182,7 @@ public static NatTable createRowNatTable(final Composite parent, final DataLayer
 			public void configureUiBindings(final UiBindingRegistry uiBindingRegistry) {
 				uiBindingRegistry.unregisterKeyBinding(new KeyEventMatcher(SWT.MOD1, 'c'));
 				uiBindingRegistry.unregisterKeyBinding(new KeyEventMatcher(SWT.MOD1, 'x'));
-				uiBindingRegistry.registerKeyBinding(new KeyEventMatcher(SWT.MOD1, 'c'), new CopyDataFromTableAction());
+				uiBindingRegistry.registerKeyBinding(new KeyEventMatcher(SWT.MOD1, 'c'), copyAction);
 				uiBindingRegistry.registerKeyBinding(new KeyEventMatcher(SWT.MOD1, 'v'),
 						new PasteDataIntoTableAction(section));
```
mx990: I would suggest to definine custom command handlers for the `CopyDataToClipboardCommand` and `PasteDataCommand` instead of overriding the default `CopyDataAction` and `PasteDataAction`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/CopyDataImportCommandHandler.java
```diff
@@ -0,0 +1,82 @@
...
+			final ListDataProvider<?> provider = (ListDataProvider<?>) dataLayer.getDataProvider();
+			return Arrays.stream(assembledCopiedDataStructure).flatMap(Arrays::stream)
+					.filter(cell -> colMapper.containsKey(columnProvider.getColumns().get(cell.getColumnIndex())))
+					.flatMap(cell -> {
+						if (provider.getRowObject(cell.getRowIndex()) instanceof final EObject decl) {
+							final List<Import> libImports = ImportHelper.getContainerImports(decl);
+							return libImports.stream().filter(imp -> ImportHelper.matchesImport(
+									PackageNameHelper.getFullTypeName(colMapper
+											.get(columnProvider.getColumns().get(cell.getColumnIndex())).apply(decl)),
+									imp));
```
mx990: I believe this does not cover the case where a type is implicitly imported (i.e., in the same package) in the source, which means there will not be a corresponding `Import` in the enclosing object. I would suggest to simply transfer a set of `Strings` with the full-qualified type name instead and figuring out what to import when pasting.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/I4diacNatTableUtil.java
```diff
@@ -21,6 +21,8 @@ public interface I4diacNatTableUtil {
 
 	void removeEntry(Object entry, CompoundCommand cmd);
 
+	void addImport(Object imports, CompoundCommand cmd);
```
mx990: I would suggest to move this into a paste handler instead of duplicating the respective code 6 times in each section.

plugins/org.eclipse.fordiac.ide.datatypeeditor/src/org/eclipse/fordiac/ide/datatypeeditor/widgets/StructEditingComposite.java
```diff
@@ -223,6 +228,18 @@ public void addEntry(final Object entry, final boolean isInput, final int index,
...
 	}
 
+	@Override
+	public void addImport(final Object imports, final CompoundCommand cmd) {
+		if (imports instanceof final Import[] importsArray
+				&& EcoreUtil.getRootContainer(getType()) instanceof final LibraryElement libElement) {
+			for (final Import imp : importsArray) {
+				if (!ImportHelper.matchesImports(imp.getImportedNamespace(), ImportHelper.getImports(libElement))) {
+					cmd.add(new AddNewImportCommand(libElement, imp.getImportedNamespace()));
+				}
```
mx990: I believe this also needs to skip unnecessary or conflicting imports. In the latter case, we also need to use the FQN when pasting the type. If the import is possible in the destination but was not possible in the source, we should also use the unqualified name. I suggest performing a similar check as in https://github.com/eclipse-4diac/4diac-ide/blob/819cb6252b9688ab77d2530f7aa3333b54d770a7/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/widgets/ImportTypeSelectionProposalProvider.java#L50-L60 and then pasting the appropriate name.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,107 @@
...
+import org.eclipse.nebula.widgets.nattable.selection.SelectionLayer;
+import org.eclipse.swt.dnd.Clipboard;
+import org.eclipse.swt.widgets.Display;
+
+public class PasteDataImportFromClipboardCommandHandler extends PasteDataFromClipboardCommandHandler {
+
+	private final SelectionLayer selectionLayer;
+	private final CommandStack cmdStack;
+	private BiFunction<TypeLibrary, String, TypeEntry> typeResolver = (typeLib, name) -> typeLib.getDataTypeLibrary()
+			.getDerivedTypeEntry(name);
```
mx990: It seems the initial value is always overwritten in the constructor?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,107 @@
...
+		this.typeResolver = typeResolver;
+	}
+
+	@Override
+	protected boolean doCommand(final PasteDataCommand command) {
+		super.doCommand(command);
+		final String[] imports = getClipboardContent();
+		final LibraryElement rootElement = getRootElement();
+
+		if (rootElement != null) {
```
mx990, line 64: I suggest to perform this check at the beginning, before already pasting the values.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/PasteDataIntoTableAction.java
```diff
@@ -52,45 +42,12 @@ public PasteDataIntoTableAction(final I4diacNatTableUtil section) {
...
 	public void run(final NatTable natTable, final KeyEvent event) {
-		final Object clipboardCellsContents = new Clipboard(Display.getCurrent())
-				.getContents(TextTransfer.getInstance());
-		final Object clipboardElementsContents = org.eclipse.gef.ui.actions.Clipboard.getDefault().getContents();
+		final Object clipboardElementsContents = Clipboard.getDefault().getContents();
 
-		if (clipboardCellsContents != null) {
-			pasteClipboardCellsContents(natTable, clipboardCellsContents);
-		} else if (clipboardElementsContents != null) {
+		if (clipboardElementsContents != null) {
```
mx990, line 47: Is there a reason why you cannot also move this into the `PasteDataFromClipboardCommandHandler` and use the vanilla `PasteDataAction`?

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/PasteDataFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,99 @@
...
+	private final SelectionLayer selectionLayer;
+
+	public PasteDataFromClipboardCommandHandler(final SelectionLayer selectionLayer) {
+		this.selectionLayer = selectionLayer;
+	}
+
+	@Override
+	protected boolean doCommand(final PasteDataCommand command) {
+		final Clipboard clipboard = new Clipboard(Display.getDefault());
+		final Object clipboardCellsContents = clipboard.getContents(TextTransfer.getInstance());
```
mx990, line 42: Please make sure to dispose the clipboard afterwards.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ImportHelper.java
```diff
@@ -182,7 +182,7 @@ private static boolean matchesImports(final String name, final List<Import> impo
 		return imports.stream().anyMatch(imp -> matchesImport(name, imp));
 	}
 
-	private static boolean matchesImport(final String name, final Import imp) {
+	public static boolean matchesImport(final String name, final Import imp) {
```
mx990: Is this change still needed?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/PasteDataIntoTableAction.java
sebHollersbacher: I did it similar to the CopyDataFromTableAction
Also, moving it to the CommandHandler would require the RowHeaderDataProvider and I4diacNatTableUtil to be passed to the constuctor and make a things more confusing.

sebHollersbacher:
I'm currently struggling to add the new CommandHandler to every existing table.
The only way I could come up with is to manually add the CommandHandlers, like it is done now in EditUntypedSubappInterfaceDataSection.
@mx990 maybe you have some ideas to resolve this nicely.

mx990:
> I'm currently struggling to add the new CommandHandler to every existing table. The only way I could come up with is to manually add the CommandHandlers, like it is done now in EditUntypedSubappInterfaceDataSection. @mx990 maybe you have some ideas to resolve this nicely.

I would suggest to wrap the relevant code into a configuration, similar to https://github.com/eclipse-4diac/4diac-ide/blob/1826295d7012a6197a841be3e9b2170cf5927151/plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java#L505-L520
and then add the configuration to the relevant tables.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/CopyDataImportCommandHandler.java
```diff
@@ -0,0 +1,85 @@
...
+					.filter(cell -> colMapper.containsKey(columnProvider.getColumns().get(cell.getColumnIndex())))
+					.map(cell -> {
+						if (provider.getRowObject(cell.getRowIndex()) instanceof final EObject decl) {
+							final LibraryElement element = colMapper
+									.get(columnProvider.getColumns().get(cell.getColumnIndex())).apply(decl);
+							if (element instanceof final DataType dt
+									&& (IecTypes.ElementaryTypes.getAllElementaryType().contains(dt)
+											|| IecTypes.GenericTypes.isAnyType(dt))) {
+								return null;
+							}
```
mx990: I would suggest to filter whether there is a package name instead to cover types without a package as well, for example using `PackageNameHelper.getPackageName(element).isEmpty()`.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/CopyDataImportCommandHandler.java
```diff
@@ -0,0 +1,85 @@
...
+							return PackageNameHelper.getFullTypeName(element);
+						}
+						return null;
+					}).filter(Objects::nonNull).filter(s -> !s.isEmpty()).distinct().toArray(size -> new String[size]);
```
mx990:
```suggestion
					}).filter(Objects::nonNull).filter(Predicate.not(String::isEmpty)).distinct().toArray(String[]::new);
```

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/DefaultImportCopyPasteLayerConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+	// TODO: adjust DirectlyDerivedTypeTableColumn, TypedElementTableColumn,
+	// VariableTableColumn
+	private static BiFunction<TypeLibrary, String, TypeEntry> getTypeResolver(final NatTableColumn column) {
+		return switch (column) {
+		case final AttributeTableColumn col -> (typeLib, name) -> {
+			final AttributeTypeEntry attributeEntry = typeLib.getAttributeTypeEntry(name);
+			if (attributeEntry != null) {
+				return attributeEntry;
+			}
+			return typeLib.getDataTypeLibrary().getDerivedTypeEntry(name);
```
mx990, line 64: Would it not make more sense to distinguish between the different columns here, similar to `getColMap` below?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/DefaultImportCopyPasteLayerConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+		};
+		case final DirectlyDerivedTypeTableColumn col ->
+			(typeLib, name) -> typeLib.getDataTypeLibrary().getDerivedTypeEntry(name);
+		case final TypedElementTableColumn col ->
+			(typeLib, name) -> typeLib.getDataTypeLibrary().getDerivedTypeEntry(name);
+		case final VarDeclarationTableColumn col ->
+			(typeLib, name) -> typeLib.getDataTypeLibrary().getDerivedTypeEntry(name);
+		case final VariableTableColumn col -> (typeLib, name) -> typeLib.getDataTypeLibrary().getDerivedTypeEntry(name);
```
mx990: I believe it would improve readability by defining a common (static) method that returns `typeLib.getDataTypeLibrary().getDerivedTypeEntry(name)` and then using a method reference in each of these clauses.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,142 @@
...
+		if (conflicts.isEmpty() || selectionLayer.getSelectionModel().getSelections().isEmpty()) {
+			return content;
+		}
+		final var location = selectionLayer.getSelectionAnchor();
+		final int idx = columnProvider.getColumns().indexOf(columns.get(0));
+
+		for (final String[] row : content) {
+			if (conflicts.containsKey(row[idx - location.getColumnPosition()])) {
+				row[idx - location.getColumnPosition()] = conflicts.get(row[idx - location.getColumnPosition()]);
+			}
```
mx990, line 92: Please extract `idx - location.getColumnPosition()` into a common variable.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+	@Override
+	protected boolean doCommand(final PasteDataCommand command) {
+		final LibraryElement rootElement = getRootElement();
+		if (rootElement != null) {
+			Arrays.stream(getClipboardContent()).map(imp -> getImportNamespace(rootElement, imp))
+					.filter(Objects::nonNull)
+					.forEach(namespace -> cmdstk.get().execute(new AddNewImportCommand(rootElement, namespace)));
+		}
+		super.doCommand(command);
```
mx990, line 71: I believe you also need to clear the conflicts after you are done?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+	private String getImportNamespace(final LibraryElement rootElement, final String imp) {
+		final TypeEntry resolvedType = ImportHelper.resolveImport(PackageNameHelper.extractPlainTypeName(imp),
+				rootElement, name -> typeResolver.apply(rootElement.getTypeLibrary(), name), name -> null);
+
+		if (resolvedType == null) {
+			return imp;
+		}
+
+		if (resolvedType.getFullTypeName().equals(imp)) {
```
mx990:
```suggestion
		if (resolvedType.getFullTypeName().equalsIgnoreCase(imp)) {
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/DefaultImportCopyPasteLayerConfiguration.java
sebHollersbacher: When this is used, we don't have any information about the columns, so it would make things more complicated.
For now, it is also the only exception, so I would leaf it like that and change it only if it is required.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	@Override
+	protected boolean doCommand(final PasteDataCommand command) {
+		final LibraryElement rootElement = getRootElement();
+		if (rootElement != null) {
+			final var namespaces = Arrays.stream(getClipboardContent()).map(imp -> getImportNamespace(rootElement, imp))
+					.filter(Objects::nonNull).toList();
+			super.doCommand(command);
+			namespaces.forEach(namespace -> cmdstk.get().execute(new AddNewImportCommand(rootElement, namespace)));
```
mx990: I believe you need to add the imports before pasting the type in the super method for it to be resolved correctly?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/PasteDataImportFromClipboardCommandHandler.java
sebHollersbacher: Yes, you are correct. Thank you for pointing that out
Unfortunately, this introduces the bug, where the selection gets cleared after importing (for the StructEditingComposite).

mx990: My guess would be that the problem is caused by a refresh when adding the import? I would suggest to change the handler to use a `CommandExecutor` instead of the `CommandStack` and add a flag to block the refresh in `StructEditingComposite.executeCommand(Command)`, similar to `AbstractSection.executeCommand(Command)`.

mx990 (approved):

azoitl (approved):

PR #715, opened 2024-11-06
ignore non-type files while manually creating type entries

ernstblechaPT:
loading libraries containing files that were not part of the type library (e.g. addtional CSV, PDF,...) failed

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
@@ -499,12 +499,14 @@ public enum LibraryManager {
 		case final IFolder f -> true;
 		case final IFile file -> {
 			final TypeEntry entry = TypeEntryFactory.INSTANCE.createTypeEntry(file);
-			final TypeEntry oldEntry = cachedTypes.get(entry.getFullTypeName());
-			if (oldEntry != null) {
-				FordiacResourceChangeListener.updateTypeEntry(file, oldEntry);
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -498,13 +498,14 @@ private static void createTypeEntriesManually(final IFolder folder, final Map<St
...
 		case final IFolder f -> true;
 		case final IFile file -> {
-			final TypeEntry entry = TypeEntryFactory.INSTANCE.createTypeEntry(file);
-			final TypeEntry oldEntry = cachedTypes.get(entry.getFullTypeName());
-			if (oldEntry != null) {
-				FordiacResourceChangeListener.updateTypeEntry(file, oldEntry);
-				cachedTypes.remove(entry.getFullTypeName());
-			} else {
-				typeLibrary.createTypeEntry(file);
+			if (TypeEntryFactory.INSTANCE.createTypeEntry(file) instanceof final TypeEntry entry) {
```
azoitl: I would find a null check clearer here. The return type is always a TypeEntry.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
ernstblechaPT: yes, a null check is clearer

azoitl (approved):

PR #716, opened 2024-11-06
Simplify contract dialogs

Felix-Schmid:
Use a general-purpose dialog instead of having a unique dialog for each different contract rule. Making this change simplifies the code and makes it easier to add new contract rules in the future.


bwiesmayr:
just to give some quick feedback, looked it through and it seems good. I'd though like to check the change in detail once more

PR #717, opened 2024-11-06
Cleaned Outline view for FB Types

azoitl:


mx990 (approved):
Looks much cleaner now. Should we also show the values of attributes in the *Outline* and *System Explorer* (in the same way as for variables)?

azoitl:
> Looks much cleaner now. Should we also show the values of attributes in the _Outline_ and _System Explorer_ (in the same way as for variables)?

Ah yes this would be an interesting thing. Will consider it for a new PR.

azoitl:
Already changed attribute presentations as separate commit as part of that PR.

pazi146 (approved):

PR #723, opened 2024-11-11
Merge release

azoitl:


azoitl:
@mx990 @MartinMelikMerkumians  I tried to merge back changes from release as preparation for the upcoming release, could you please have a quick look if I haven't forgotten anything.

.github/workflows/maven.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Upload
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
       with:
         name: event_file
-        path: ${{ github.event_path }}
+        path: ${{ github.event_path }}
```
MartinMelikMerkumians: I would suggest to remove the extra spaces after }}

.github/workflows/maven.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Upload
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
```
MartinMelikMerkumians, line 15: Following up our short discussion, should the need for the concrete commit sha on the github action be documented somewhere? Otherwise, I see a high potential for people commit code to "fix" this to the version tag, as it is usually done

CONTRIBUTING.md
```diff
@@ -1,70 +1,168 @@
...
+Short description/summary as to why this change is being made
+  
+Fixes https://github.com/<Full URL to issues/issue #>
+```
+
+
+  Example: 
+
+```
+Fixed NPE when opening the instance viewer of CFB instances #321
```
MartinMelikMerkumians, line 153: While I appreciate the example commit message and consider the example best practice, it does not reflect the lived practice currently.
I only found 5 commits having an issue ID in the summary and none having a link to the issue in the commit message when looking through commit messages ranging back to begin of October.

While this question is a bit outside of this PR, should the requirements be enforced by some git hook, if possible?

MartinMelikMerkumians (changes requested):

CONTRIBUTING.md
azoitl: not sure, we currently have to little issues from outside, and we use PRs to document partly our work. So I think I would not enforce this.

.github/workflows/maven.yml
azoitl: Yes at github as it is required by dependabot: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-pre-written-building-blocks-in-your-workflow#using-release-management-for-your-custom-actions

.github/workflows/maven.yml
azoitl: done

.github/workflows/maven.yml
MartinMelikMerkumians: ok, after some more talk, I understood that the sha tags are generated by the dependabot, after a first manual switch from tags to shas

MartinMelikMerkumians (approved):

PR #730, opened 2024-11-12
Performance Improvement of Model Search

azoitl:
With this change the search pattern is precompiled. Furthermore a precheck of algorithms and methods is done to reduce the need for creating evaluators.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
+++ b/plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 Primetals Technologies Austria GmbH
+ * Copyright (c) 2022, 2024 Primetals Technologies Austria GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -13,21 +13,24 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.model.search;
 
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class ModelSearchPattern { // extends SearchPattern
 
-	private final String toTest;
-	private Pattern searchPattern;
-	private final String searchString;
+	private final Pattern searchPattern;
+	private final Pattern preScannSTPattern;
+	private final String preScannSTStringPattern;
 
-	public ModelSearchPattern(final String toTest, final ModelQuerySpec modelQuerySpec) {
-		this.toTest = toTest;
-		this.searchString = modelQuerySpec.getSearchString();
+	public ModelSearchPattern(final ModelQuerySpec modelQuerySpec) {
+		searchPattern = Pattern.compile(convertSearchStringToPattern(modelQuerySpec.getSearchString()));
+
+		preScannSTStringPattern = generatePreScanSTStringPattern(modelQuerySpec.getSearchString());
+		preScannSTPattern = (containsRegex(preScannSTStringPattern))
+				? Pattern.compile(preScannSTStringPattern, Pattern.CASE_INSENSITIVE)
+				: null;
 	}
 
-	private String convertSearchStringToPattern() {
+	private static String convertSearchStringToPattern(final String searchString) {
 		String temp = searchString;
 		if (searchString.contains("?")) { //$NON-NLS-1$
 			temp = searchString.replace("?", "[a-zA-Z0-9_]"); //$NON-NLS-1$ //$NON-NLS-2$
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchQuery.java
```diff
@@ -297,23 +301,57 @@ private boolean matchEObject(final INamedElement modelElement, final IProgressMo
...
+		final String text = getImplText(modelElement);
+		boolean preCheck = false;
+
+		if (text != null) {
+			String searchStringLC = modelQuerySpec.getSearchString().toLowerCase();
+			final int lastPackageSepeartor = searchStringLC.lastIndexOf("::"); //$NON-NLS-1$
+			if (lastPackageSepeartor != -1) {
+				searchStringLC = searchStringLC.substring(lastPackageSepeartor + 2);
+			}
+			if (text.toLowerCase().contains(searchStringLC)) {
```
mx990: What if the search string contains wildcards? If we search using a case-insensitive Pattern, we could also avoid converting to lower case.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
```diff
@@ -13,21 +13,17 @@
...
-	private final String toTest;
-	private Pattern searchPattern;
-	private final String searchString;
+	private final Pattern searchPattern;
 
-	public ModelSearchPattern(final String toTest, final ModelQuerySpec modelQuerySpec) {
-		this.toTest = toTest;
-		this.searchString = modelQuerySpec.getSearchString();
+	public ModelSearchPattern(final ModelQuerySpec modelQuerySpec) {
+		searchPattern = Pattern.compile(convertSearchStringToPattern(modelQuerySpec.getSearchString()));
```
mx990, line 25: Should the Pattern be case insensitive?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
```diff
@@ -13,21 +13,17 @@
...
 	}
 
-	private String convertSearchStringToPattern() {
+	private static String convertSearchStringToPattern(final String searchString) {
```
ernstblechaPT, line 33: This seems to only allow either `?` or `*`  - if you have both `?` wins

Also: if searchString contains characters that are already part of a regular expression this might give confusing results

Also: not sure if the call to `contains` actually improves performance here as the `searchString` should be rather short anyways...

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
azoitl: Yes it should

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchQuery.java
azoitl: That is a good idea. I was also not happy about the lower case conversion.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
azoitl: I just reworked the existing code to be a bit more efficient. After fixing the prescanning of ST parts we can have a look into improving the matcher. AFAIK this code is based on similar code in Eclipse Platform. But I'm not sure. Will talk with the original developer.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
azoitl: In my latest PR I didn't touch now the existing search as I was not perfectly sure about the impact on search behavior. I guess this should be investigated next.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
azoitl: In my latest PR I didn't touch now the existing search as I was not perfectly sure about the impact on search behavior. I guess this should be investigated next.

After some searching I found an approach that should be more efficient then `contains` however it is not as easy to read. Curios if it brings some benefit.

ernstblechaPT: Yes, this is absolutely something that should be done in a seperate PR

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
```diff
@@ -13,21 +13,25 @@
...
-	private final String searchString;
+	private final Pattern searchPattern;
+	private final Pattern preScannSTPattern;
+	private final String preScannSTStringPattern;
 
-	public ModelSearchPattern(final String toTest, final ModelQuerySpec modelQuerySpec) {
-		this.toTest = toTest;
-		this.searchString = modelQuerySpec.getSearchString();
+	public ModelSearchPattern(final ModelQuerySpec modelQuerySpec) {
+		modelQuerySpec.getSearchString();
```
mx990: This statement seems to have no effect?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchPattern.java
azoitl: yes that should have been removed. Done.

mx990 (approved):

PR #732, opened 2024-11-13
Make "Type" column editable on Untyped Subapp Interface Data Section

alliuqemal:


mx990 (approved):

azoitl (approved):
This fixes the type column. However also the initial value and visible column should be editable. Could you please do this in a separate commit.

PR #738, opened 2024-11-14
Propagate Initial value from old pin that has been renamed

oberlehner:


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ReconnectPinChange.java
```diff
@@ -80,7 +83,7 @@ class ReconnectPinByName extends Command {
 
 	final String oldName;
 	final String newName;
-	final FBNetworkElement element;
+	FBNetworkElement element;
```
mx990: I believe this change is unnecessary?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringProcessor.java
```diff
@@ -90,19 +90,22 @@ private void createChildChanges(final CompositeChange change) {
...
-		String oldName = ""; //$NON-NLS-1$
-		if (eChild instanceof final IInterfaceElement varDecl) {
-			oldName = varDecl.getName();
-		}
 
 		for (final EObject eObject : result) {
-			if (eObject instanceof FBNetworkElement) {
-				change.add(new ReconnectPinChange(EcoreUtil.getURI(eObject), FBNetworkElement.class, newName, oldName));
+			if (eObject instanceof final FBNetworkElement element
+					&& eChild instanceof final IInterfaceElement varDecl) {
```
mx990: I suggest moving the `eChild instanceof final IInterfaceElement varDecl` out of the for loop.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringProcessor.java
```diff
@@ -90,19 +90,22 @@ private void createChildChanges(final CompositeChange change) {
...
+				createRenameVarDeclChanges(change, element, varDecl);
 			}
 		}
 
 	}
 
+	private void createRenameVarDeclChanges(final CompositeChange change, final FBNetworkElement element,
+			final IInterfaceElement varDecl) {
```
mx990: I believe this is also used for other interface elements and not just `VarDeclaration`s? In that case, I would suggest to change the naming to reflect that.

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #763, opened 2024-11-20
Adjusted test JobGroup timeouts

pazi146:


tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java
```diff
@@ -99,7 +102,7 @@ void cleanupAfterEach() throws Exception {
 		}
 		FordiacMarkerHelper.updateMarkers(project.getFile(LibraryManager.MANIFEST), FordiacErrorMarker.LIBRARY_MARKER,
 				Collections.emptyList(), true);
-		FordiacMarkerHelper.JOB_GROUP.join(1000, null);
+		FordiacMarkerHelper.JOB_GROUP.join(timeout, null);
```
mx990: I had a look at the different `JobGroups` you are using here and noticed that you have `seedJobsCount` set to `0`. This means the `join` call will return immediately, see also the JavaDoc for `JobGroup.JobGroup(String, int, int)`.

You should also be aware that a `JobGroup` will be canceled by default if one of its `Job`s fails. This means, for example, if one of the marker updates fails, all future marker updates may be canceled immediately.

In summary, I do *not* believe `JobGroup` is the right solution for your problem and may even be harmful during normal operation.

I would suggest to instead override `Job.belongsTo(Object)` in the relevant jobs to associate them with a common `family` and then use `IJobManager.join(Object, IProgressMonitor)` to join on all currently running jobs of the particluar family in your tests. See also `AutoBuildJob.belongsTo(Object)` and `BuildManager.waitFor(Job)` for an example on how to implement this.

mx990 (changes requested):

tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java
pazi146: I'll look into this - just some notes: 
- `join` doesn't return immediately just because `seedJobsCount` is `0` - the jobGroup won't transition into the `NONE` state until that number of jobs has completed; this can also simply be tested with a breakpoint inside a job -> `join` indeed will wait until it times out in this use case
- as for canceling: a `JobGroup` won't be canceled if one of its `Job`s fails - the `Job`'s status will change from `RUNNING` to `NONE` and be treated as completed by the `JobGroup` (and increase the `failedJobsCount` by 1)

pazi146: Additionally we use `WorkspaceJob`s here which have the the standard implementation of `belongsTo(Object)` returning `false` - thus not getting handled by `IJobManager.join(Object, IProgressMonitor)`

mx990: > - `join` doesn't return immediately just because `seedJobsCount` is `0` - the jobGroup won't transition into the `NONE` state until that number of jobs has completed; this can also simply be tested with a breakpoint inside a job -> `join` indeed will wait until it times out in this use case

The `JobGroup` is in the state `NONE` by default. `join` waits until *both* no jobs of the group are currently running or scheduled to run (i.e., state is `NONE`) *and* at least `seedJobsCount` have completed. So as long as you have already scheduled a job for the group, `join` will indeed wait even if `seedJobsCount` is zero. Are you certain this is always the case in your tests or could it be that some of the jobs for which you need to wait are scheduled after calling `join`?

> - as for canceling: a `JobGroup` won't be canceled if one of its `Job`s fails - the `Job`'s status will change from `RUNNING` to `NONE` and be treated as completed by the `JobGroup` (and increase the `failedJobsCount` by 1)

When a `Job` completes, `JobManager.updateJobGroup(IJobChangeEvent)` cancels the `JobGroup` if `InternalJobGroup.shouldCancel(IStatus, int, int)` returns true. The default implementation returns `numberOfFailedJobs > 0`, so it will cancel the group when there are any failed jobs.

> Additionally we use `WorkspaceJob`s here which have the the standard implementation of `belongsTo(Object)` returning `false` - thus not getting handled by `IJobManager.join(Object, IProgressMonitor)`

As suggested above, the whole point is to override `Job.belongsTo(Object)` in each of the jobs, so it returns true for the respective new family.

mx990: I would also suggest to update the markers synchronously instead of using a workspace job, since you are running either in a workspace job already or in isolation inside a test:
- https://github.com/eclipse-4diac/4diac-ide/blob/b409b8c3fcce0596d101289ee0472f6fe426d9ce/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java#L832-L833
- https://github.com/eclipse-4diac/4diac-ide/blob/b409b8c3fcce0596d101289ee0472f6fe426d9ce/tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java#L100-L102

This avoids unnecessary jobs and also has the added benefit that it saves you from having to wait for the markers to be updated in your tests.

pazi146: Removed the `JobGroup`s in their entirety and replaced them with job families - removing all the pitfalls you mentioned.

I have not added extra code for marker updates as they already get executed synchronously in the test cases (at least where it is relevant)

mx990 (approved):

tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java
```diff
@@ -56,6 +57,9 @@ class LibraryImportTest {
 	private static final String V2_0_0 = "2.0.0"; //$NON-NLS-1$
 	private static final String LIB_LOC = "WORKSPACE_LOC/.lib/"; //$NON-NLS-1$
 	private static final String PROJECT = "testproject"; //$NON-NLS-1$
+
+	private final long timeout = 10000;
```
azoitl: this variable is not used anymore.

azoitl (changes requested):

tests/org.eclipse.fordiac.ide.test.library/src/org/eclipse/fordiac/ide/test/library/LibraryImportTest.java
pazi146: removed

azoitl (approved):

PR #768, opened 2024-11-20
Ensure that for var_in_out pins the attributes are shown for inputs

azoitl:


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/filters/AttributeFilter.java
```diff
@@ -35,22 +36,26 @@ public static Object parseObject(final Object input) {
...
+			// for var_in_outs we can only have attributes on the input side
+			if (!interfaceElement.isIsInput() && interfaceElement instanceof final VarDeclaration varDecl
+					&& varDecl.isInOutVar()) {
+				return varDecl.getInOutVarOpposite();
+			}
 		}
-		return null;
+
+		return (inputHelper instanceof final FBNetwork fbNetwork) ? fbNetwork.eContainer() : inputHelper;
+
```
mx990: This is extremely hard to read. I would like to suggest an alternate implementation:
```java
	public static ConfigurableObject parseObject(final Object input) {
		return switch (input) {
		// for var_in_outs we can only have attributes on the input side
		case final VarDeclaration varDecl when varDecl.isInOutVar() && !varDecl.isIsInput() ->
			varDecl.getInOutVarOpposite();
		// handle exception: interface elements of functions
		case final IInterfaceElement interfaceElement when interfaceElement
				.eContainer() instanceof final InterfaceList interfaceList
				&& interfaceList.eContainer() instanceof FunctionFBType ->
			null;
		case final ConfigurableObject configurableObject -> configurableObject;
		case final FBNetwork fbNetwork -> parseObject(fbNetwork.eContainer());
		case final EditPart editpart -> parseObject(editpart.getModel());
		case final IAdaptable adaptable -> adaptable.getAdapter(ConfigurableObject.class);
		case null, default -> null;
		};
	}
```
This also allowed me to notice that the entire clause for the `IInterfaceElement` is currently not considered when the input is an `IInterfaceElement` already, because of the `if (input instanceof ConfigurableObject)` at the beginning.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/filters/AttributeFilter.java
azoitl: that is a good point. Missed that.

mx990 (approved):

PR #770, opened 2024-11-21
Divided Update Changes into separate rename/delete changes

oberlehner:
Delete Changes should not change anything, therefore they are disabled per default.
Rename Changes need to actually rename the regarding element as default option, so it is not possible to use the same changes for both refactorings.
Therefore, they are splitted up which also gives the benefits to provide more specific repair options.

Furthermore, if a change is enabled, there will be always a selection in the viewer, so it is not possible for the user to enable a change without a corresponding operation that needs to exectuted

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/Messages.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/Messages.java
@@ -114,6 +114,7 @@ public final class Messages extends NLS {
 	public static String DeleteFBTypeParticipant_Change_UpdateSubappPins;
 	public static String FBTypeComposedAdapterFactory_FBTypecomposedAdapterFactoryShouldNotBeInsantiated;
 	public static String IFordiacPreviewChange_Reconnect0;
+	public static String ConfigurableChange_noChangeErrorMessage;
 
 	public static String ImportChange_ImportedNamespaceChanged;
 
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameUpdateFBTypeInterfaceChange.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameUpdateFBTypeInterfaceChange.java
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Primetals Technologies Austria GmbH
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Daniel Lindhuber
+ *     - initial API and implementation and/or initial documentation
+ *   Michael Oberlehner - outsourced inner class to own file
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.typemanagement.refactoring.rename;
+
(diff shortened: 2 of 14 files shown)


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/Messages.java
```diff
@@ -113,6 +113,8 @@ public final class Messages extends NLS {
 	public static String DeleteFBTypeParticipant_Change_UpdateManipulator;
 	public static String DeleteFBTypeParticipant_Change_UpdateSubappPins;
 	public static String FBTypeComposedAdapterFactory_FBTypecomposedAdapterFactoryShouldNotBeInsantiated;
+	public static String IFordiacPreviewChange_0;
```
mx990: Please change the name of the constant to something meaningful.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameUpdateFBTypeInterfaceChange.java
```diff
@@ -0,0 +1,107 @@
...
+
+	public RenameUpdateFBTypeInterfaceChange(final FBType type, final StructuredType struct) {
+		super(MessageFormat.format(Messages.DeleteFBTypeParticipant_Change_DeleteFBTypeInterface, type.getName(),
+				struct.getName()), EcoreUtil.getURI(type), FBType.class);
+		this.struct = struct;
+	}
+
+	@Override
+	public EnumSet<ChangeState> getAllowedChoices() {
+		return EnumSet.of(ChangeState.RENAME, ChangeState.CHANGE_TO_ANY, ChangeState.DELETE);
```
mx990: Should we even offer the choices `CHANGE_TO_ANY` and `DELETE` when renaming?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameUpdateStructDataTypeMemberVariableChange.java
```diff
@@ -0,0 +1,87 @@
...
+	public RenameUpdateStructDataTypeMemberVariableChange(final VarDeclaration varDeclaration) {
+		super(MessageFormat.format(Messages.DeleteFBTypeParticipant_Change_UpdateMemberVariable,
+				varDeclaration.getName(), varDeclaration.getTypeName(),
+				((INamedElement) varDeclaration.eContainer()).getName()), EcoreUtil.getURI(varDeclaration),
+				VarDeclaration.class);
+	}
+
+	@Override
+	public EnumSet<ChangeState> getAllowedChoices() {
+		return EnumSet.of(ChangeState.RENAME, ChangeState.DELETE, ChangeState.CHANGE_TO_ANY, ChangeState.NO_CHANGE);
```
mx990: same here

mx990 (changes requested):
I would also suggest adding a fatal error in `isValid` for each of these changes when `NO_CHANGE` is selected, so the refactoring cannot proceed when any such changes are enabled.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameUpdateFBTypeInterfaceChange.java
oberlehner: you are right, i think i will remove it unless someone explicitly wants it

mx990 (approved):

azoitl (approved):

PR #780, opened 2024-11-25
Revert 777 freeze

oberlehner:
Merge develop to freeze was accidential


mx990:
I believe this reverts the wrong commit, that is, the merge from freeze to develop instead of vice versa?

oberlehner:
> I believe this reverts the wrong commit, that is, the merge from freeze to develop instead of vice versa?

i fucked something up. I think this should be ok now:
https://github.com/eclipse-4diac/4diac-ide/pull/782

PR #781, opened 2024-11-25
Do not preload mapped subapp networks during import

mx990:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java
```diff
@@ -262,9 +261,6 @@ private void parseMapping() throws XMLStreamException {
 		final FBNetworkElement toElement = (fromElement instanceof CommunicationChannel)
 				? findMappingTargetFromName(toValue, fromElement)
 				: findMappingToElement(fromValue, toValue);
-		if (fromElement instanceof SubApp) {
-			FBNetworkHelper.loadSubappNetwork(fromElement);
```
oberlehner, line 266: are you sure that this has no influence on elements which are mapped inside an untyped subapp ?

mx990: The network of untyped subapps is always loaded when the untyped subapp is parsed. The implementation in `UntypedSubapp` thus simply returns the already loaded network, so this is a no-op for untyped subapps: https://github.com/eclipse-4diac/4diac-ide/blob/2f045539eaae3fdb73287bf688c1a0ccdf9d74e0/plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/UntypedSubAppImpl.java#L181-L183

azoitl (approved):

PR #784, opened 2024-11-26
Override basicSetTypeEditable to update system entry after refactoring

oberlehner:
This fix solves the following problem:
On reopening the editor uses a different library element then the refactoring operation.
Therefore, the XML is not in sync with the library element. 

how to reproduce:
1) Rename Interfaceelement in an FBType Editor
2) Select reconnect (all connections of the block instances will be reconnected)
3) close the editor of the instance (e.g AutomationEditor)
4) reopen the editor after refactoring

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/SystemEntryImpl.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/SystemEntryImpl.java
@@ -15,6 +15,7 @@
  ******************************************************************************/
 package org.eclipse.fordiac.ide.model.typelibrary.impl;
 
+import org.eclipse.emf.common.notify.NotificationChain;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.fordiac.ide.model.dataexport.AbstractTypeExporter;
 import org.eclipse.fordiac.ide.model.dataexport.SystemExporter;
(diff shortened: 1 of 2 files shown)


oberlehner:
This fix only works in combination with https://github.com/eclipse-4diac/4diac-ide/pull/740

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/SystemEntryImpl.java
```diff
@@ -80,4 +81,11 @@ protected AbstractTypeExporter getTypeExporter(final AutomationSystem type) {
 	public EClass getTypeEClass() {
 		return LibraryElementPackage.Literals.AUTOMATION_SYSTEM;
 	}
+
+	@Override
+	protected synchronized NotificationChain basicSetTypeEditable(final LibraryElement newTypeEditable,
+			final NotificationChain notifications) {
+		return super.basicSetType(newTypeEditable, notifications);
+	}
+
```
mx990, line 90: I believe you also need to similarly override `basicGetTypeEditable()` for this to work correctly.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/SystemEntryImpl.java
azoitl: ah yes you are right. It gets time that we get rid of the editable.

mx990 (approved):

azoitl (approved):

Issue #799, opened 2024-11-29
Impossible to use watches on distributed application


melekhove:
Steps to reproduce:
-  Create simple distributed application as described in [Tutorial Step 2](https://github.com/eclipse-4diac/4diac-documentation/blob/main/src/tutorials/distribute4diac.adoc)
- Select all devices on the application tab and toggle Watches
- Start two FORTE instances
- Deploy to both instances
- Right Click on any device in system configuration, select Monitor As/Monitor
- Observe the error dialog

Here is the stack trace
```
!STACK 0
java.lang.IllegalStateException: Duplicate key TestTwoDevicesApp.E_SR (
attempted merging values org.eclipse.fordiac.ide.deployment.debug.watch.FBNetworkElementWatch@12169ca9 
and org.eclipse.fordiac.ide.deployment.debug.watch.FBNetworkElementWatch@12169ca9)
	at java.base/java.util.stream.Collectors.duplicateKeyException(Collectors.java:135)
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:182)
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base/java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet.lambda$entryConsumer$0(Collections.java:1778)
	at java.base/java.util.concurrent.ConcurrentSkipListMap$EntrySpliterator.forEachRemaining(ConcurrentSkipListMap.java:3360)
	at java.base/java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet$UnmodifiableEntrySetSpliterator.forEachRemaining(Collections.java:1803)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
[...]

PR #800, opened 2024-11-29
Fix watchpoints handling (#799)

melekhove:
The reason of the problem is that watches are added to all devices regardless of where the target FB is actually located.


mx990:
Thank you for looking into this. Unfortunately, your current solution will not work for watches on interface elements or inside nested subapp and composite FB networks. I also believe this approach would not scale well when including all possible sub-elements.

I have created a PR https://github.com/eclipse-4diac/4diac-ide/pull/801 with an alternative solution to address these concerns and would be glad if you could have a look or maybe even try it out.

melekhove:
Yes, your solution is obviously better. I overlooked the possibility to filter in `DeploymentWatchpoint`, shame on me :-)
Thanks!

PR #805, opened 2024-12-01
Reworked SystemEditor for Editing the List of Applications

azoitl:
The system editor now shows only the list of applications. There now applications are created, deleted, and reordered.

The NewApplicationWizard has been removed as this should not be done anymore via the system explorer.


oberlehner:
i am not able to enter a comment
image

azoitl:
Changing names or editing the comment is not yet possible. Will follow after more clean-ups

azoitl:
Finished all basic features and removed all problematic old code. Kept it in separate commits to make it easier reviewable. I also think they should be merged as individual commits.

PR #811, opened 2024-12-02
Update infrastructure for changes of full qualfied instance names

oberlehner:


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.libraryElement.INamedElement;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
+import org.eclipse.gef.commands.Command;
+
+public abstract class QualNameAffectedCommand extends Command implements ScopedCommand {
```
mx990: I would suggest changing this into an interface to make it easier to add to an existing class hierarchy of commands.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -0,0 +1,55 @@
...
+		this.oldQualName = oldQualName;
+	}
+
+	protected String getOldQualName() {
+		return oldQualName;
+	}
+
+	protected abstract String getNewQualName();
+
+	protected abstract INamedElement getNotifier();
```
mx990: I would suggest renaming this to `getRenamedElement` (or something similar) to make it more clear what is being returned.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -0,0 +1,55 @@
...
+	 * encapsulate the change to not provide the command to the receiver
+	 */
+	public QualNameChange getQualNameChange() {
+		return new QualNameChange(getOldQualName(), getNewQualName(), getNotifier(), getTypeEntry(getNotifier()));
+	}
+
+	protected static TypeEntry getTypeEntry(final INamedElement notifier) {
+		final EObject rootContainer = EcoreUtil.getRootContainer(notifier);
+		Assert.isTrue(rootContainer instanceof LibraryElement);
+		if (rootContainer instanceof final LibraryElement element) {
```
mx990: I believe having another instanceof check after an assertion makes little sense. I would suggest removing the assertion.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListener.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import org.eclipse.core.commands.operations.AbstractOperation;
+import org.eclipse.fordiac.ide.model.libraryElement.INamedElement;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
+
+public abstract class QualNameChangeListener {
```
mx990, line 23: Please extract the implementation into a separeate class and change this into an interface.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,133 @@
...
+		final IConfigurationElement[] config = Platform.getExtensionRegistry()
+				.getConfigurationElementsFor(EXTENSION_POINT_ID);
+		for (final IConfigurationElement element : config) {
+			try {
+				final Object obj = element.createExecutableExtension("class");
+				if (obj instanceof QualNameChangeListener) {
+					listeners.add((QualNameChangeListener) obj);
+				}
+			} catch (final Exception e) {
+				e.printStackTrace();
```
mx990: Please properly log the exception (including the contributing plugin).

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,133 @@
...
+	private static final String EXTENSION_POINT_ID = "org.eclipse.fordiac.ide.model.commands.QualNameChangeListener";
+	private final List<QualNameChangeListener> listeners = new ArrayList<>();
+
+	public void initialize() {
+		final IConfigurationElement[] config = Platform.getExtensionRegistry()
+				.getConfigurationElementsFor(EXTENSION_POINT_ID);
+		for (final IConfigurationElement element : config) {
+			try {
+				final Object obj = element.createExecutableExtension("class");
+				if (obj instanceof QualNameChangeListener) {
```
mx990: Please log an error if the class is not valid (e.g., force an exception by removing the instanceof).

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,133 @@
...
+			}
+		}
+	}
+
+	public static void addCommandStackEventListener(final CommandStack commandStack) {
+		commandStack.addCommandStackEventListener(INSTANCE);
+	}
+
+	public void removeCommandStackEventListener(final CommandStack commandStack, final Object notfier) {
+		listeners.forEach(l -> l.flush(notfier));
```
mx990, line 62:
```suggestion
	public void removeCommandStackEventListener(final CommandStack commandStack, final Object notifier) {
		listeners.forEach(l -> l.flush(notifier));
```

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+		}
+
+		if (event.getDetail() == CommandStack.POST_MARK_SAVE && event.getSource() instanceof final CommandStack stack) {
+
+			final TypeEntry typeEntry = getTypeEntryKeyFromCommandStack(stack);
+			if (typeEntry != null) {
+				notifiyCommit(typeEntry);
+			}
+		}
```
mx990, line 92: What about `POST_FLUSH` (e.g., because a reload of the editor)?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+	}
+
+	private static TypeEntry getTypeEntryKeyFromCommandStack(final CommandStack stack) {
+		final IEditorPart currentActiveEditor = EditorUtils.getCurrentActiveEditor();
```
mx990: I believe this does not handle the case where multiple editors are saved at once (e.g., *Save All*)? I would suggest adding a dedicated listener instance per editor to keep track of the type entry.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeNameCommand.java
```diff
@@ -48,9 +48,11 @@ private ChangeNameCommand(final INamedElement element, final String name) {
 	}
 
 	private ChangeNameCommand(final INamedElement element, final String name, final boolean validateName) {
+		super(element.getQualifiedName());
 		this.element = Objects.requireNonNull(element);
 		this.name = name;
 		this.validateName = validateName;
+		this.oldName = element.getQualifiedName();
```
mx990: The `oldName` is supposed to be the (unqualified) name of the element and is assigned in `execute()`.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeNameCommand.java
```diff
@@ -168,4 +170,25 @@ public Set<EObject> getAffectedObjects() {
...
+		final StringBuilder sb = new StringBuilder();
+		for (int i = 0; i < split.length; i++) {
+			if (i == split.length - 1) {
+				sb.append(name);
+			} else {
+				sb.append(split[i]);
+				sb.append("."); //$NON-NLS-1$
+			}
+		}
+		return sb.toString();
```
mx990:
```suggestion
		return getOldQualName().substring(0, getOldQualName().length() - oldName.length()) + name;
```

plugins/org.eclipse.fordiac.ide.hierarchymanager.ui/META-INF/MANIFEST.MF
```diff
@@ -19,7 +19,10 @@ Require-Bundle: org.eclipse.core.resources,
  org.eclipse.fordiac.ide.ui,
  org.eclipse.fordiac.ide.typemanagement,
  org.eclipse.ui.ide,
- org.eclipse.ui.views.properties.tabbed
+ org.eclipse.emf.mwe.core,
```
mx990: Do we actually need `org.eclipse.emf.mwe.core` as a dependency?

mx990 (changes requested):
I may have a simpler solution for handling qualified name changes in the listener:
1. collect all execute/undo/redo changes in a single list,
2. when performing the changes, first merge all changes for the same element in order.

This way, we should be able to avoid keeping a separate history for changes in addition to the command stack. What do you think?

plugins/org.eclipse.fordiac.ide.hierarchymanager.ui/META-INF/MANIFEST.MF
oberlehner: ### no

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
oberlehner: yes that makes sense

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeNameCommand.java
oberlehner: you are right. i misused it

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
oberlehner: i changed it to `getChangedElement` , i think its better then renamed, we also have the case of deletion.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
oberlehner: The reason for that Assert, is that i want to know early if we get a wrong element here.
For example calling with a cached object where the eContainer can be null.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListener.java
oberlehner: I see no reason why i should invent another layer of abstraction.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
oberlehner: I think a postflush should also flush the pending changes ? what do you think ?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
oberlehner: Changed that

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
mx990: In that case, I would suggest replacing the second instanceof with a simple cast, since the if can only be true after the assertion.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListener.java
mx990: I guess we can wait until the need arises.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
mx990: A `POST_FLUSH` should discard all pending charges, since the editor was either reloaded or closed without saving.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameChangeListenerManager.java
```diff
@@ -0,0 +1,136 @@
...
+	INSTANCE;
+
+	QualNameChangeListenerManager() {
+		initialize();
+	}
+
+	private static final String EXTENSION_POINT_ID = "org.eclipse.fordiac.ide.model.commands.QualNameChangeListener";
+	private final List<QualNameChangeListener> listeners = new ArrayList<>();
+
+	public void initialize() {
```
azoitl: this method should be private, maybe also initializeListeners would be a better name.

azoitl (changes requested):

pazi146 (approved):

PR #829, opened 2024-12-04
MappingTargetCreator didn't Correctly Moved Parameters to Mapped Model

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/MappingTargetCreator.java
```diff
@@ -46,6 +46,7 @@ public static FBNetworkElement createMappingTarget(final Resource res, final FBN
 			if (srcElement instanceof final ConfigurableFB srcConfFB) {
 				setupConfigureableFB(srcConfFB, (ConfigurableFB) created);
 			}
+			transferFBParams(srcElement, created);
```
mx990: How about using `InterfaceListCopier.copy(srcElement.getInterface(), true, true)` when copying the interface list above instead?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/MappingTargetCreator.java
azoitl: Good idea

mx990 (approved):

PR #830, opened 2024-12-04
Create Connections for Mapped FBs Created During Mapping

azoitl:
For FBs that are created during mapping the required connections are now also created in the resource model.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java
@@ -20,10 +20,12 @@ package org.eclipse.fordiac.ide.model.dataimport;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java
```diff
@@ -459,4 +471,32 @@ private void parseApplication(final AutomationSystem automationSystem)
...
+	 * check if connections need to be generated in resource model.
+	 *
+	 */
+	private void generateMappedFBConnections() {
+		for (final FBNetworkElement fbnEl : mappedFBs) {
+			final FBNetworkElement srcResFb = fbnEl.getOpposite();
+			final Resource res = fbnEl.getResource();
+			fbnEl.getInterface().getOutputs().flatMap(ie -> ie.getOutputConnections().stream()) //
+					.filter(con -> con.getDestinationElement().getResource() == res) //
+					.forEach(con -> res.getFBNetwork().addConnection(createResourceCon(srcResFb, con)));
```
mx990, line 489: I assume you are creating only outgoing connections to avoid duplicates? What if the source of a connection did exist in the resource network but the destination (and mutual connections) did not? What about the output side of InOut variables?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java
azoitl: resource only connections will be handled specially. For that I'm currently reviewing and reworking the connection parsing code.

Output side of InOut variables is included in the stream provided by getOutputs.

mx990 (approved):

PR #841, opened 2024-12-06
Add the Properties View to the Debug Perspective

azoitl:
During debugging it is also necessary to inspect properties of elements in the graphical editors.


bwiesmayr (approved):
works flawlessly

PR #842, opened 2024-12-07
Key bindings for graphical editors

azoitl:
To avoid clashes with text editing and other eclipse based editors an own editor context for GEF based editors and especially for FBNetwork Editors was introduced.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide.application/plugin.properties
@@ -1,19 +1,19 @@
- ###############################################################################
- # Copyright (c) 2008 - 2017 Profactor GmbH, TU Wien ACIN, fortiss GmbH
- #				 2019 - 2020 Andrea Zoitl
- # 
- # This program and the accompanying materials are made available under the
- # terms of the Eclipse Public License 2.0 which is available at
- # http://www.eclipse.org/legal/epl-2.0.
- #
- # SPDX-License-Identifier: EPL-2.0
- #
- # Contributors:
- #   Gerhard Ebenhofer, Alois Zoitl, Monika Wenger
- #     - initial API and implementation and/or initial documentation
- #   Andrea Zoitl
- #     - Externalized all translatable strings 
- ###############################################################################
+###############################################################################
+# Copyright (c) 2008, 2024 Profactor GmbH, TU Wien ACIN, fortiss GmbH, 
+#                          Andrea Zoitl, Primetals Technologies Austria GmbH
+# 
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0.
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Gerhard Ebenhofer, Alois Zoitl, Monika Wenger
+#     - initial API and implementation and/or initial documentation
+#   Andrea Zoitl
+#     - Externalized all translatable strings 
+###############################################################################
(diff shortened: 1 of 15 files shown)


plugins/org.eclipse.fordiac.ide.application/plugin.properties
mx990, line 1: Should this also be added to the `build.properties`?

plugins/org.eclipse.fordiac.ide.gef/plugin.xml
```diff
@@ -139,30 +161,45 @@
...
+            contextId="org.eclipse.fordiac.ide.gef"
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
-            sequence="M1+M2+0">
-      	</key>
+            sequence="M1+M2+0"/>
+      	<key
+	        commandId="org.eclipse.fordiac.ide.gef.zoomIn"
+	        contextId="org.eclipse.fordiac.ide.gef"
+			schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
+	        sequence="M1++"/>
```
mx990, line 173: We may also want to register a keybinding for the `M1+=` combination, same as the window-wide version.

mx990 (changes requested):
The keybindings currently do not work inside the `SubAppNetworkBreadCrumbEditor`, because we are not using the nested editor site there. I have created a PR https://github.com/azoitl/4diac-ide/pull/6 to fix the problem, hopefully without breaking anything else.

plugins/org.eclipse.fordiac.ide.gef/plugin.xml
azoitl: I was not sure about it, I guess it is more used in english keyboards, added it. Also I switched back to the orginal zoom command id from the text editor. Otherwise the toolbar entries will only be active on graphical editors which felt strange.

plugins/org.eclipse.fordiac.ide.application/plugin.properties
azoitl: Ah yes, thx for noting. I experimented with combining the properties files and the approach I tried didn't work. And then I forgot to update the build.properties. I'm rather happy with my new approach as now translators only need to look into one file. Maybe later when we have more translation it will make sense to move them into a dedicated i18n directory.

azoitl:
> The keybindings currently do not work inside the `SubAppNetworkBreadCrumbEditor`, because we are not using the nested editor site there. I have created a PR [azoitl#6](https://github.com/azoitl/4diac-ide/pull/6) to fix the problem, hopefully without breaking anything else.

Will look into it.

plugins/org.eclipse.fordiac.ide.gef/plugin.properties
```diff
@@ -1,19 +1,20 @@
...
- #
- # Contributors:
- #   Gerhard Ebenhofer, Alois Zoitl
- #     - initial API and implementation and/or initial documentation
- #   Andrea Zoitl
- #	   - externalized translatable strings
- ###############################################################################
+###############################################################################
+# Copyright (c) 2008, 2024 Profactor GbmH, fortiss GmbH, Andrea Zoitl,
+						   Primetals Technologies Austria GmbH
```
mx990: Please add the missing `#` at the start of the line.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.gef/plugin.properties
azoitl: fixed

mx990 (approved):

PR #846, opened 2024-12-09
Added open type in editor on instance viewer screen

alliuqemal:


oberlehner (approved):

plugins/org.eclipse.fordiac.ide.application/plugin.xml
```diff
@@ -1544,19 +1563,25 @@
       </command>
       <command commandId="org.eclipse.fordiac.ide.application.commands.openType">
```
azoitl, line 1500: this is now twice in the pop-up menu

azoitl (changes requested):
With your changes it feels now that we have the open type in three different places in the pop-up menu. Can you please harmonize to one.

oberlehner (changes requested):
image

In this situation, i want to 
1) enable "Open Type in Editor"
2) It needs to jump to the FBNetwork Editor to Test->Subapp

PR #848, opened 2024-12-09
Draft Commit for QuickFix Widget

PStmmr:
Retrieving markers from files for types not visible in the view.
Still testing with the wrong icon.


features/.gitignore
oberlehner, line 1: This file must not be comitted

features/pom.tycho
oberlehner, line 1: remove

plugins/.gitignore
oberlehner, line 1: remove

plugins/pom.tycho
oberlehner, line 1: remove this

tests/pom.tycho
oberlehner, line 1: remove

PR #855, opened 2024-12-11
[Clean-up] Reworked EMF Resource saving to not use typeeditable

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
```diff
@@ -116,21 +132,37 @@ protected void doSave(final OutputStream outputStream, final Map<?, ?> options)
...
-			typeEntryForFile.setTypeEditable(EcoreUtil.copy((LibraryElement) getContents().get(0)));
-			typeEntryForFile.setLastModificationTimestamp(typeEntryForFile.getFile().getModificationStamp());
-			AbstractTypeExporter.saveType(typeEntryForFile, outputStream);
+			saveType(getContents().get(0), outputStream);
 		} catch (final Exception e) {
 			throw new IOWrappedException(e);
 		}
 	}
 
+	public static void saveType(final EObject le, final OutputStream outputStream) {
```
oberlehner: what is le ?
i think libraryElement would be better.
I am not a big fan of short variable names, but that is a matter of taste

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
```diff
@@ -116,21 +132,37 @@ protected void doSave(final OutputStream outputStream, final Map<?, ?> options)
...
+		final AbstractTypeExporter exporter = getTypeExporter(le);
+		if (exporter != null) {
+			try (InputStream inputStream = exporter.getFileContent()) {
+				inputStream.transferTo(outputStream);
+			} catch (final IOException e) {
+				FordiacLogHelper.logError(e.getMessage(), e);
```
mx990: Please throw the exception from `doSave()` instead of silently logging it to allow proper error handling in the caller.

mx990 (changes requested):
As the next step, I would suggest to create individual resources for each type. We could then make `FordiacTypeResource` an abstract base class and determine the exporter with an abstract `getTypeExporter()` (similar to the type entries) instead of the switch.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
azoitl: yes you are right. will fix it right after.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
azoitl: ok will do that. Another advantage of my clean-up this was hidden in the AbstractTypeExporter bevore.

azoitl:
> As the next step, I would suggest to create individual resources for each type. We could then make `FordiacTypeResource` an abstract base class and determine the exporter with an abstract `getTypeExporter()` (similar to the type entries) instead of the switch.

I mainly did this to reduce that usages of getTypeEditable and to make our current resource implementation a bit better. We do not have to much users of this resource so I would only spend the effort if we will more widely use resources in 4diac IDE (e.g., merge TypeEntry and Resource).

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
```diff
@@ -115,22 +130,28 @@ protected void doSave(final OutputStream outputStream, final Map<?, ?> options)
...
+		return switch (content) {
+		case final FunctionFBType funcType -> new FCTExporter(funcType);
+		case final AdapterType adpType -> new AdapterExporter(adpType);
+		case final SubAppType subAppType -> new SubApplicationTypeExporter(subAppType);
+		case final FBType fbType -> new FbtExporter(fbType);
+		case final AnyDerivedType dataType -> new DataTypeExporter(dataType);
+		case final AutomationSystem system -> new SystemExporter(system);
+		case final GlobalConstants globalConstants -> new GlobalConstantsExporter(globalConstants);
+		case final AttributeDeclaration attributeType -> new AttributeTypeExporter(attributeType);
+		default -> null;
```
mx990: Maybe we should also throw an `IOException` if the content is not supported?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
mx990: Yes, I very much appreciate the clean-up and am glad we noticed this now.

mx990:
> > As the next step, I would suggest to create individual resources for each type. We could then make `FordiacTypeResource` an abstract base class and determine the exporter with an abstract `getTypeExporter()` (similar to the type entries) instead of the switch.
> 
> I mainly did this to reduce that usages of getTypeEditable and to make our current resource implementation a bit better. We do not have to much users of this resource so I would only spend the effort if we will more widely use resources in 4diac IDE (e.g., merge TypeEntry and Resource).

I only mentioned it because of the disconnect between having an extensible implementation for type entries but not for resources. However, I would see this as very low priority in any case.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResource.java
azoitl: I think this is a good idea. Better save then sorry.

mx990 (approved):

PR #858, opened 2024-12-11
Fix unavailable attributes of removed marker annotations

mx990:
The attributes of removed marker annotations are no longer available after the corresponding marker has been removed. This caches the attributes inside the marker annotation to avoid the problem.


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/ResourceMarkerGraphicalAnnotationModel.java
```diff
@@ -112,6 +112,7 @@ protected void markerChanged(final IMarkerDelta delta, final Set<GraphicalAnnota
 					added.add(annotation);
 				}
 			} else {
+				annotation.refresh();
```
oberlehner, line 115: what if the marker has been deleted from the problems view in the meantime, when you call refresh. are you sure that this will never throw a NPE

mx990: That is not possible, because:
1. the workspace is locked while executing resource change listeners, 
2. the resulting `CoreException` would be caught in `refresh()` in any case.

PR #867, opened 2024-12-11
Moved Connection Parsing and Checking to Connection Helper

azoitl:
In order to make the ConnectionBuilder better maintainable and allow it to be used for creating mapped to unmapped FB connections all connection parsing, creation, and checking code is moved to the ConnectionBuilder.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
@@ -119,6 +119,22 @@ public final class ConnectionHelper {
 			}
 		}
 
+		public void repair() {
+			if (isMissingConnectionSource()) {
+				connection.setSource(getConnectionEndPoint(sourceString, false));
+			}
+			if (isMissingConnectionDestination()) {
+				connection.setDestination(getConnectionEndPoint(destinationString, true));
+			}
+			connectionState.clear();
+			connectionState.add(ConnectionState.VALID);
+			validate();
+			handleErrorCases();
+			if (getConnection() != null) {
+				importer.getFbNetwork().addConnection(connection);
+			}
+		}
+
 		public void handleErrorCases() {
 			if (isMissingConnectionDestination()) {
 				handleMissingConnectionDestination();
@@ -377,6 +393,10 @@ public final class ConnectionHelper {
 			FBNetworkElement element = importer.findFBNetworkElement(elementName);
 			while (element == null && separatorPos != -1) {
 				separatorPos = path.indexOf('.', separatorPos + 1);
+				if (separatorPos == -1) {
+					// we have no element with the name
+					return null;
+				}
 				elementName = path.substring(0, separatorPos);
 				element = importer.findFBNetworkElement(elementName);
 			}
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021-2022 Primetals Technologies Austria GmbH
+ * Copyright (c) 2021, 2024 Primetals Technologies Austria GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -10,6 +10,7 @@
  * Contributors:
  *   Michael Oberlehner - initial API and implementation and/or initial documentation
  *                      - added data type check for connection
+ *   Alois Zoitl        - moved connection checking and building here
  *******************************************************************************/
 package org.eclipse.fordiac.ide.model.dataimport;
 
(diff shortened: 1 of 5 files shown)


oberlehner (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -10,39 +10,71 @@
...
 			this.connectionState = EnumSet.of(ConnectionState.VALID);
-			this.source = source;
-			this.destination = destination;
+			this.sourceString = sourceString;
+			this.destinationString = destinationString;
+			this.connection = connection;
+			this.importer = importer;
+
+			connection.setSource(getConnectionEndPoint(sourceString, false));
+			connection.setDestination(getConnectionEndPoint(destinationString, true));
```
mx990, line 72: How about we always resolve and validate connections at the end, also for regular networks? That should save us the extra repair step and also make regular networks immune to the order of XML elements?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -79,21 +111,67 @@ public void validate() {
 				connectionState.remove(ConnectionState.VALID);
 			}
 
-			if (destination != null && destInterfaceList != null) {
+			if (destinationString != null && destInterfaceList != null) {
```
mx990: Why is this condition different to the block for the source above?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -179,82 +253,224 @@ public boolean isEmptyConnection() {
...
 
-		public InterfaceList getSrcInterfaceList() {
+		private InterfaceList getSrcInterfaceList() {
 			return srcInterfaceList;
 		}
 
-		public void setSrcInterfaceList(final InterfaceList srcInterfaceList) {
+		private void setSrcInterfaceList(final InterfaceList srcInterfaceList) {
 			this.srcInterfaceList = srcInterfaceList;
 		}
```
mx990: Should we inline these simple field accesses?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -179,82 +253,224 @@ public boolean isEmptyConnection() {
...
+
+			// we need a special treatment for FB's that lost their type
+			if (oppositeEndpoint == null) {
+				getConnectionState().add(ConnectionState.MISSING_TYPE);
+				return null;
+			}
+
+			DataType type = oppositeEndpoint.getType();
+			if (type == null) {
+				type = IecTypes.GenericTypes.ANY;
```
mx990: Should this not also use `determineConnectionType(...)` above to properly handle event or adapter connections?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
azoitl: with the work I did here we could reconsider later. But as the order in the xml is defined by the dtd I see less need for all networks.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
azoitl: I guess that is a copy paste error from the original code. The code here is only migrated from the original FBNetwork importer will fix that.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -79,21 +111,67 @@ public void validate() {
 				connectionState.remove(ConnectionState.VALID);
 			}
 
-			if (destination != null && destInterfaceList != null) {
+			if (destinationEndpoint != null && destInterfaceList != null) {
```
mx990: Should this not rather just be:
```suggestion
			if (destInterfaceList != null) {
```
as above: https://github.com/eclipse-4diac/4diac-ide/blob/cd46a5449024fe2847367cd23856d8ba7f783ee7/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java#L100-L101
or am I missing something?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
oberlehner: I think martin is right

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -79,21 +111,67 @@ public void validate() {
 				connectionState.remove(ConnectionState.VALID);
 			}
 
-			if (destination != null && destInterfaceList != null) {
+			if (destinationEndpoint != null) {
```
mx990:
```suggestion
			if (destInterfaceList != null) {
```

mx990 (changes requested):

mx990 (approved):

PR #870, opened 2024-12-12
Optimize Sys file generation

azoitl:
Mapped entities are not written into the sys file anymore. As they are recreated on parsing this is not a problem and reduces the size of sys files as well as the changes of git conflicts in multi user settings.


mx990 (approved):
Works nicely!

PR #883, opened 2024-12-18
Displaying ErrorPins for only non system Attributes.

PStmmr:
Also fixed a case where a method was renamed incorrectly.


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateFBNElementCommand.java
```diff
@@ -513,4 +521,9 @@ public FBNetworkElement getNewElement() {
 	public Set<EObject> getAffectedObjects() {
 		return Set.of(network);
 	}
+
+	private static boolean isInternal(final Attribute attribute) {
```
oberlehner: i think this method would also be usefull for others.
Please rename it to **isInternalAttribute** and move it to `org.eclipse.fordiac.ide.model.datatype.helper.InternalAttributeDeclarations`

oberlehner:
Currently it is not working when i delete an event pin.

Before deletion of the output pin CNF:
image

After deletion there is no error marker generated for CNF.

PStmmr:
@oberlehner I see it's currently only checked for input/outputvars. Which pins besides events should also be checked for?

oberlehner:
> @oberlehner I see it's currently only checked for input/outputvars. Which pins besides events should also be checked for?

you can also check how varInOut behave

oberlehner (approved):

PR #892, opened 2024-12-19
Fixed configurable fb updates

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
```diff
@@ -183,11 +186,14 @@ static String getMemberVarName(final MemberVarDeclaration varDecl) {
...
 		if (!(muxer.getDataType() instanceof StructuredType)) {
 			// e.g., error data type
-			muxer.getMemberVars().clear();
-			getEventWithPins(muxer).getWith().clear();
-		} else if (muxer instanceof final Demultiplexer demux && demux.isIsConfigured()) {
+			return;
+		}
+		if (muxer instanceof final Demultiplexer demux && demux.isIsConfigured()) {
 			// updating requires finding all elements again - struct may have changed!
 			updateConfiguredDemuxConfiguration(demux, buildVisibleChildrenString(muxer.getMemberVars()));
```
mx990: I believe `buildVisibleChildrenString(muxer.getMemberVars())` will now always return an empty string, since `memberVars` has been cleared above? Maybe we could build the string from the struct itself instead?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
```diff
@@ -243,10 +247,6 @@ private static void updateConfiguredDemuxConfiguration(final Demultiplexer demux
 		if (visibleChildren == null || !(demux.getDataType() instanceof StructuredType) || !demux.isIsConfigured()) {
 			updateStructManipulatorConfiguration(demux);
```
mx990: It seems strange that we are also called from `updateStructManipulatorConfiguration`? If I understand the flow of code correctly, we would get in the else clause of `updateStructManipulatorConfiguration` from here, since `isIsConfigured()` would be false? On the other hand, we would get in the else clause here when called from `updateStructManipulatorConfiguration`, since `isIsConfigured()` would be true then? Could we please untangle this?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
bwiesmayr: It seems that the cleaning should not happen as it is now. Unfortunately, building the string from the struct itself does not work because we might have also grandchildren of the struct enabled, which would not be found.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConfigurableFBManagement.java
bwiesmayr: I think that this was an extra safety check in case we get from a place where we shouldn't ... just in case someone (me?) misuses the methods later. It is particularly there because error markers sometimes leave me clueless what is supposed to happen -> what might have been a configured demux could be just a "normal" demux later if the pin was removed. A way to detangle could be to handle error markers in a special method and call it from both.

azoitl:
Given all your comments I noticed that my mapping code was using the muxer configuration generation wrongly, and I was "fixing" at the wrong place. Therefore I will revert that part of my change. If we ever need a clean-up of the configurable FB code we can do that in a separate commit.

mx990 (approved):

PR #893, opened 2024-12-19
Added all 4diac IDE type files as XML content

azoitl:


mx990 (approved):

azoitl:
Unfortunately my first version was not good enough as our editors where not registered as default editors and to be totally save and have more control I added sub content entries for all main editors and IEc 61499 file types.

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
```diff
@@ -115,19 +126,19 @@
    <extension point="org.eclipse.ui.commands">
     	<command
         	id="org.eclipse.fordiac.ide.typemanagement.openType"
-        	name="Open Type">
+        	name="%command.Open Type">
```
mx990: I believe you need to remove the spaces?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
azoitl: yes you are totally right. done.

mx990 (approved):

PR #557, opened 2024-09-25
Reduce compiler warnings in the eclipse project

ernstblechaPT:
This tries to reduce the amount of warnings that is generated for the 4diac IDE plugins in Eclipse.

PR #895, opened 2024-12-20
Update compiler settings

mx990:
Update compiler settings and ignore unhandled suppress warnings.


azoitl:
The drawback I'm seeing that if code is changing and we don't get any info we keep these unhandled warnings for ever.

mx990:
> The drawback I'm seeing that if code is changing and we don't get any info we keep these unhandled warnings for ever.

This is just for `@SupressWarnings` that are not understood by Eclipse, such as those for SonarLint. If there is a `@SupressWarnings` that is not necessary (anymore), we should still get a warning.

azoitl:
> > The drawback I'm seeing that if code is changing and we don't get any info we keep these unhandled warnings for ever.
> 
> This is just for `@SupressWarnings` that are not understood by Eclipse, such as those for SonarLint. If there is a `@SupressWarnings` that is not necessary (anymore), we should still get a warning.

Yes you are right. Still I'm unsure about this change. What are the others thinking?

oberlehner:
I don't see a benefit in this commit. You can filter the warnings in your problems view

ernstblechaPT:
I already did something like this back in September: https://github.com/eclipse-4diac/4diac-ide/pull/557#pullrequestreview-2331638866

I decided agaist fully disabling the message and just set it to "INFO" as this is a warning that can almost always be safely ignored - but sometimes becomes relevant after code was deleted during refactoring.

Filtering the warnings in the problems view does not achieve the same result for me (unless there is a way of adding multiple filters at the same time that I do not know of) as I (most of the time) do not care about these warnings at all but they still increase the number of listed warnings (and thus require more attention on my side as a developer).

mx990:
> I decided agaist fully disabling the message and just set it to "INFO" as this is a warning that can almost always be safely ignored - but sometimes becomes relevant after code was deleted during refactoring.

There are two distinct settings:
- `unhandledWarningToken` for `@SuppressWarnings` that are unknown to Eclipse (e.g., SonarQube checks),
- `unusedWarningToken` for `@SuppressWarnings` that are unnecessary (e.g. `static-method` but the method could not be static anyhow).

So we can ignore unknown `@SuppressWarnings` to get rid of markers for disabled SonarQube checks, while still keeping warnings (or infos) for `@SuppressWarnings` that are unnecessary.

> Filtering the warnings in the problems view does not achieve the same result for me (unless there is a way of adding multiple filters at the same time that I do not know of) as I (most of the time) do not care about these warnings at all but they still increase the number of listed warnings (and thus require more attention on my side as a developer).

Same for me. It would also have to be done manually by every developer and the markers still show up as annotations in the code, in the overview ruler on the side, and as decorations in the *Project Explorer*. All of this takes valuable time and concentration away from actual problems that can and should be fixed.

PR #905, opened 2024-12-28
Harmonize release and develop maven github workflow

azoitl:
As dependabot only checks the default branch the same settings between the branches are important. This commit moves the develop settings to release so that in the future it easier to merge from release to develop.


azoitl:
release branch is currently not building. Merge needed for other tests.

Issue #925, opened 2025-01-07
FB Types are not accessible


Dazwennn:
### Info:
- Pre build nightly
- Version: 3.0.0
- Build id: 2025-01-04_0644
- OS: Ubuntu 24.04.1 LTS

### Description:
After creating a new project in a fresh IDE instance, there seems to be no way to access the standard FB Types. In the second step of the "Create Project" wizard, all Types are shown and selected. However, after clicking "Finish," the "Type Library", "External Libraries" and "Standard Libraries" in the Project Explorer remain empty.
Additionally, using right click -> "insert FB" there are no suggestions and there is nothing inside the palette. 
The .fbt files itself do exist and are accessible in the IDE's folder structure.

### Workarounds:
- Just import the Typelibrary from the repo.
- Build the ide yourself. Tested branch 3.0.x at commit [d276a12](https://github.com/eclipse-4diac/4diac-ide/commit/d276a12f423b361e630e4c4b64ec42338d13a31b) from Dec 14, 2024. There the content of the Type Library is not visable for the first created project but for the second one.

![image](https://github.com/user-attachments/assets/299dc6d1-f0b1-423a-957a-f0219c811043)
![image](https://github.com/user-attachments/assets/23bb935e-d2e8-47eb-83a3-da71bccd04b4)

azoitl:
Can you please also post your log file.

Dazwennn:
[log.log](https://github.com/user-attachments/files/18330904/log.log)

pazi146:
Could you try again with the current develop branch?

We switched from linking libraries inside the `Type Library` folder to linking them in `Standard Libraries` and `External Libraries`

These changes haven't yet been merged into `freeze` and `3.0.x` - a conversion from old to new is in place, though opening "newer" projects with these branches will lead to strange behaviour

Dazwennn:
> Could you try again with the current develop branch?

Sure, same behavior as the nightly pre-build.


> We switched from linking libraries inside the `Type Library` folder to linking them in `Standard Libraries` and `External Libraries`
> 
> These changes haven't yet been merged into `freeze` and `3.0.x` - a conversion from old to new is in place, though opening "newer" projects with these branches will lead to strange behaviour

Thanks for the work there and the clarification.

pazi146:
I've had issues with projects behaving strangely when using the same workspace with different versions - especially when switching back to an older one (switching from `develop` to `freeze` as an example)

[...]

PR #927, opened 2025-01-07
Fixed Initialization of LibraryManager #925

azoitl:
The LibraryManager was lazy initialized on resolving entries. This was to late when a new project is created. Especially for the first project in a new workspace.

In order to get this working I needed to turn several fields from static final to only final, as in Java the constructor of an enum is not allowed to access static final fields.

https://github.com/eclipse-4diac/4diac-ide/issues/925


Dazwennn:
Fixed the problem for me. Thank you!

pazi146 (approved):

PR #936, opened 2025-01-09
add event trigger function block

MandKastner:
This FB will trigger all unconnected events of the specified type within a resource on REQ

related to [https://github.com/eclipse-4diac/4diac-forte/pull/298](url)

### DIFF
--- a/data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
+++ b/data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<FBType Name="E_TRIG" Comment="Trigger FB for a specified Event Type" >
+<FBType Name="E_TRIG" Comment="Trigger unconnected input events of the specified type inside a resource" >
 	<Identification Standard="61499-2">
 	</Identification>
 	<VersionInfo Version="1.0" Author="Mario Kastner" Date="2025-01-08">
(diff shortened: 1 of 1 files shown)


data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FBType Name="E_TRIG" Comment="Trigger FB for a specified Event Type" >
```
oberlehner: i dont like the name and the comment. For me its not really clear what this block should do. 
If the name is defined by the IEC 61499 standard, then i would leave it like, otherwise we should give it a more descriptive way that makes it clear that all blocks with the event type will be triggered.
@azoitl what do you think about that ?

data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
```diff
@@ -0,0 +1,26 @@
...
+			<Event Name="REQ" Type="Event" Comment="Request trigger of unconnected event pins of type EVENTTYPE" >
+				<With Var="EVENTTYPE"/>
+			</Event>
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="CNF" Type="Event" Comment="Confirmation of Requested Service" >
+			</Event>
+		</EventOutputs>
+		<InputVars>
+			<VarDeclaration Name="EVENTTYPE" Type="STRING" Comment="Event type to be triggered" />
```
ernstblechaPT: `'EInit'` could be a sensible InitialValue for `EVENTTYPE`?

data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
azoitl: I would also go for an improved comment. There is no name in the standard for this block. But I thought keeping it short and simple and in line with existing blocks (E_RTRIG, E_FTRIG) would be a good idea.

data/typelibrary/events-1.0.0/typelib/E_TRIG.fbt
azoitl: definitly

azoitl (approved):

PR #972, opened 2025-01-19
Correctly handle the right interfacebar in collision detection

azoitl:
To correctly handle collisions of children of expanded subapps the width of the subapp has to be reduced.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
@@ -20,6 +20,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.draw2d.geometry.Dimension;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
```diff
@@ -111,32 +115,68 @@ static boolean validateCollisions(final Optional<FBNetworkElement> parent,
...
+				diagnostics.add(createCollisionDiagnostic(element, parent));
+			}
+		}
+		return parentCollision;
+	}
+
+	private static Dimension getParentSize(final FBNetworkElement parent) {
+		int commentLines = 1;
+		if (parent.getComment() != null && !parent.getComment().isBlank()) {
+			commentLines += parent.getComment().chars().filter(c -> c == '\n').count();
```
oberlehner, line 136: are you sure that this will work on every OS ?

azoitl: yes, we store in our comments at least one \n for every newline in a comment.

oberlehner (approved):

PR #975, opened 2025-01-20
Allow treating partial access as variable

ernstblechaPT:


plugins/org.eclipse.fordiac.ide.model.eval.st/src/org/eclipse/fordiac/ide/model/eval/st/StructuredTextEvaluator.java
```diff
@@ -668,6 +668,8 @@ protected Variable<?> evaluateVariable(final STExpression expr) throws Evaluator
 			evaluateVariable(stBuiltinFeatureExpression);
 		case final STFeatureExpression stFeatureExpression -> evaluateVariable(stFeatureExpression);
 		case final STMemberAccessExpression stMemberAccessExpression -> evaluateVariable(stMemberAccessExpression);
+		case final STMultibitPartialExpression stMultiBitPartialExpression ->
+			evaluateVariable(stMultiBitPartialExpression);
```
mx990, line 672: This will produce an infinite recursion, since it just calls itself. It should also not be necessary, since partial access is already handled in https://github.com/eclipse-4diac/4diac-ide/blob/5637d81c8e39c9ab96fb838b26c3d2a268d5fd68/plugins/org.eclipse.fordiac.ide.model.eval.st/src/org/eclipse/fordiac/ide/model/eval/st/StructuredTextEvaluator.java#L729-L730

A partial access without corresponding receiver makes no sense and should never occur.

mx990 (changes requested):

PR #978, opened 2025-01-21
Provide text action bar contributor for text editors in type editor

mx990:
The text editors inside a multi-page type editor need to have a proper text action contributor to properly support switching of actions between pages and also for the new search overlay. This provides a text action bar contributor for text editors inside a type editor.


plugins/org.eclipse.fordiac.ide.systemmanagement.ui/plugin.xml
```diff
@@ -68,7 +68,7 @@
          point="org.eclipse.ui.editors">
       <editor
             class="org.eclipse.fordiac.ide.systemmanagement.ui.editors.AutomationSystemEditor"
-            contributorClass="org.eclipse.fordiac.ide.gef.editors.GraphicalMultipageEditorContributor"
+            contributorClass="org.eclipse.fordiac.ide.typeeditor.TypeEditorContributor"
```
azoitl, line 71: with that change the typeeditor plugin needs to be added to the manifest otherwise I get a class not found exception when opening a system editor.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/plugin.xml
mx990: Done

azoitl (approved):

PR #990, opened 2025-01-24
Fix premature apply when selecting proposal in initial value editor

mx990:
Fix premature apply when selecting proposal in initial value editor due to loss of focus while proposal popup is open.


azoitl (approved):
This is really an improvement. Can we have that also for the other initial value editors: when pin is selected and inline editing in graphical editor?

mx990:
> This is really an improvement. Can we have that also for the other initial value editors: when pin is selected and inline editing in graphical editor?

I hope that together with https://github.com/eclipse-4diac/4diac-ide/pull/992, I have now covered all of the initial value editors (i.e., nat tables, inline editing in graphical editor, and pin property sheet). If it is still not working somewhere, please let me know.

azoitl:
oh I was not patient enough. Will test later.

PR #992, opened 2025-01-24
Fix 2nd premature apply when selecting proposal in initial value editor

mx990:
Fix second premature apply when selecting proposal in initial value editor due to loss of focus while proposal popup is open.


azoitl:
For me this does not work in the pin property sheet and in direct edit field in the graphical editor for initial values. If I select a content proposal with enter I loose the focus and the editor is closed.

mx990:
It should now work in all locations when selecting a proposal both via the mouse *and* via the enter key, as well as pressing escape to close the popup. Thanks for checking!

oberlehner:
The initial value editor anyway behaves strange.
I am not sure if this is part of this PR, but this should be documented.
I think there shouldn't be so many entries here. Tested on your PR

image

mx990:
I agree that the code completion still has considerable room for improvement, not just for initial values but also ST code in general. However, I would see that as a separate issue.

azoitl (approved):

azoitl:
Yes now it works as the cell editors for the NAT tables.

PR #993, opened 2025-01-25
New Wizard for Creating Projects based on Examples

azoitl:
This wizard will take zip files located in an examples directory of 4diac IDE and list them. Users can select an example and import them. The example will be copied so that users can modify them. Per default the target location is the workspace but a directory can be chosen.

This new wizard is also added to the title page for easy access for new users.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/wizard/New4diacExampleWizard.java
+++ b/plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/wizard/New4diacExampleWizard.java
@@ -1,0 +1,68 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Johannes Kepler University Linz
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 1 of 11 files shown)


oberlehner (approved):

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/wizard/New4diacExampleWizard.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *     Alois Zoitl     - the zip handling
```
bwiesmayr: which one is correct?

azoitl: Thx for noticing. The second one is correct. Somehow I have regularly troubles to update full copyright headers. The reason is that originally I had the zip handling code from GEF Classic in the Wizard. Only later noticed it is better to have it in its own class.

bwiesmayr (approved):

PR #996, opened 2025-01-26
Enumeration support

azoitl:


mx990 (changes requested):
Would it be possible to inherit `EnumeratedValue` from `INamedElement`?

azoitl:
> Would it be possible to inherit `EnumeratedValue` from `INamedElement`?

When I did the model update first I thought it is not possible because of circular dependencies between models. But as you asked I noticed that we have something similar in the struct data type. So I did change. I think this is much nicer now.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/DataTypeImporter.java
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2020 Johannes Kepler University, Linz
+ * Copyright (c) 2020, 2205 Johannes Kepler University, Linz
```
bwiesmayr: 2025

PR #998, opened 2025-01-26
Removed Draw2d & GEF Examples from Target Platform

azoitl:
The examples are no dependency we should use. Furthermore with latest changes in the examples they are more demanding and not fitting to our current target platform.


azoitl:
Merging as it fixes the other build errors.

PR #1000, opened 2025-01-27
Add property page for automatic type export

MandKastner:


plugins/org.eclipse.fordiac.ide.export.forte_lua/plugin.xml
```diff
@@ -7,7 +7,8 @@
             class="org.eclipse.fordiac.ide.export.forte_lua.ForteLuaExportFilter"
             description="FORTE Lua Export"
             name="FORTE Lua"
-            sortIndex="20">
+            sortIndex="20"
+            id="org.eclipse.fordiac.ide.export.exportFilter.lua">
```
mx990, line 11: Please make sure to also update the schema with the added `id` attribute in [plugins/org.eclipse.fordiac.ide.export/schema/org.eclipse.fordiac.ide.export.exportFilter.exsd](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.export/schema/org.eclipse.fordiac.ide.export.exportFilter.exsd).

plugins/org.eclipse.fordiac.ide.export/plugin.properties
```diff
@@ -13,7 +13,9 @@
  ###############################################################################
  
 AutomaticTypeExport_Enable=Enable automatic type export at save action
+AutomaticTypeExport_ExporterDefault=Default Exporter
```
mx990: I believe this should be just *Exporter*, since *Default* implies that this can be changed or overridden, which does not seem to be the case.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/AutomaticTypeExportPropertyPage.java
```diff
@@ -74,41 +97,72 @@ public void widgetSelected(final SelectionEvent e) {
...
-		directoryEditor.load();
-		directoryEditor.setEnabled(getPreferenceStore().getBoolean(PreferenceConstants.ENABLE_AUTOMATIC_TYPE_EXPORT),
-				directoryEditorContainer);
+		directoryEditor.setPreferenceStore(getPreferenceStore());
+		directoryEditor.setPage(this);
+	}
+
+	private String getDefaultDirectory() {
+		if (getProject().getFolder(EXTERNAL_MODULE_DIR_NAME).exists()) {
+			return getProject().getFolder(EXTERNAL_MODULE_DIR_NAME).getRawLocation().toOSString();
```
mx990: I believe the (default) output folder should be a project-relative path. Since this is a project-specific setting, an absolute path would not necessarily be valid when the project is shared between users.

plugins/org.eclipse.fordiac.ide.export/plugin.xml
```diff
@@ -14,6 +14,23 @@
                class="org.eclipse.fordiac.ide.export.builder.ExportBuilder">
          </run>
       </builder>
+   </extension>
+   <extension
+         point="org.eclipse.ui.propertyPages">
+      <page
+            class="org.eclipse.fordiac.ide.export.properties.AutomaticTypeExportPropertyPage"
+            id="org.eclipse.fordiac.ide.export.settingspage"
+            name="Automatic Type Export Settings">
```
mx990: I would suggest to rename this to just *Type Export*, since both *automatic* and *settings* feels a bit redundant in this context.

plugins/org.eclipse.fordiac.ide.export/plugin.properties
```diff
@@ -12,8 +12,11 @@
  #     - externalized all translatable strings
  ###############################################################################
  
+AutomaticTypeExport_Enable=Enable automatic type export at save action
```
mx990: I believe this is misleading, since the export happens during build and not as a save action. I would instead suggest *Export types during build*.

plugins/org.eclipse.fordiac.ide.export/plugin.properties
```diff
@@ -12,8 +12,11 @@
...
+AutomaticTypeExport_ExporterDefault=Default Exporter
+AutomaticTypeExport_ExternalModulesDirectory=Path to external modules directory
```
mx990: I would suggest to rename this to *Output Folder*, as the term *external modules directory* is something specific to current FORTE builds.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/utils/ExportFilterUtil.java
```diff
@@ -0,0 +1,55 @@
...
+		return Stream.of(elems).sorted((o1, o2) -> {
+			try {
+				final int sortIndex1 = Integer.parseInt(o1.getAttribute(SORT_INDEX));
+				final int sortIndex2 = Integer.parseInt(o2.getAttribute(SORT_INDEX));
+				return sortIndex1 - sortIndex2;
+			} catch (final NumberFormatException e2) {
+				FordiacLogHelper.logError(e2.getMessage(), e2);
+			}
+			return 0;
+		}).toArray(IConfigurationElement[]::new);
```
mx990, line 52: It may be sensible to cache this information, see [plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/internal/TypeEditorPageFactory.java](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/internal/TypeEditorPageFactory.java) for an example.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,229 @@
...
+			setChangeButtonText(JFaceResources.getString("openBrowse"));//$NON-NLS-1$
+			setValidateStrategy(VALIDATE_ON_FOCUS_LOST);
+			createControl(parent);
+		}
+
+		@Override
+		protected String changePressed() {
+			final IPath projectRelativePath = new Path(getTextControl().getText())
+					.makeRelativeTo(getProject().getFullPath());
+			final IFolder startDirectory = getProject().getFolder(projectRelativePath);
```
mx990: I get an exception here when I press *Browse...* and the text field is empty or the path does not start with the project name.

Also, why do you need the `makeRelativeTo` in the first place? Can we not just have the project-relative path in the text field already?

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,229 @@
...
+	private Composite directoryEditorContainer;
+	private Composite exporterEditorContainer;
+
+	private static final String OUTPUT_FOLDER_NAME = "out"; //$NON-NLS-1$
+
+	@Override
+	protected Control createContents(final Composite parent) {
+		final Composite composite = new Composite(parent, SWT.NONE);
+		GridLayoutFactory.fillDefaults().applyTo(composite);
+		GridDataFactory.fillDefaults().applyTo(composite);
```
mx990: It seems your inner composite does not properly *grab* the available space.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,229 @@
...
+			String fileName = getTextControl().getText();
+			fileName = fileName.trim();
+			if (fileName.isEmpty() && isEmptyStringAllowed()) {
+				return true;
+			}
+			return getProject().getFolder(new Path(fileName).lastSegment()).exists();
+		}
+
+		private Optional<IFolder> getDirectory(final IFolder startingDirectory) {
+			final DirectoryDialog fileDialog = new DirectoryDialog(getShell(), SWT.OPEN | SWT.SHEET);
```
mx990: Please do not use the standard file chooser for selecting a workspace folder. I would suggest to use an `ElementTreeSelectionDialog` and then filter for folders:
```java
final ElementTreeSelectionDialog dialog = new ElementTreeSelectionDialog(getShell(), new WorkbenchLabelProvider(), new WorkbenchContentProvider());
// set title, message, comparator, etc.
dialog.setInput(getProject());
dialog.addFilter(new ViewerFilter() {

  @Override
  public boolean select(final Viewer viewer, final Object parentElement, final Object element) {
    return element instanceof IFolder; // and other filters, such as not a source folder, virtual or linked
  }
});
```

There is also a `ContainerSelectionDialog` but that one unfortunately always show the entire workspace.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
MandKastner: Thanks thats way better. Do you know how i can filter the .settings folder out of the file dialog? 
I thought this is a hidden folder because it is also not shown in the system explorer, but that did not work.

azoitl: In the system explorer it is a dedicated filter. I fear you need to filter yourself.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,245 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Mario Kastner
+ *     - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+package org.eclipse.fordiac.ide.export.properties;
```
mx990: Please also move this to `org.eclipse.fordiac.ide.export.ui.properties` in the `org.eclipse.fordiac.ide.export.ui` plugin.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,245 @@
...
+			fileDialog.setInput(getProject());
+			fileDialog.setAllowMultiple(false);
+			fileDialog.setTitle(Messages.TypeExport_OutputFolder);
+			fileDialog.setMessage(Messages.TypeExport_FileDialogMessage);
+
+			fileDialog.addFilter(new ViewerFilter() {
+				@Override
+				public boolean select(final Viewer viewer, final Object parentElement, final Object element) {
+					return (element instanceof final IFolder folder) && !folder.isVirtual() && !folder.isLinked()
+							&& !folder.getName().startsWith(".") && !(element instanceof SourceFolder); //$NON-NLS-1$
```
mx990: The element is never a `SourceFolder` here. In order to filter out source folders, you would have to load the buildpath and then check if the folder is not inside a source folder using:
```java
final Buildpath buildpath = BuildpathUtil.loadBuildpath(getProject());
BuildpathUtil.findSourceFolder(buildpath, folder).isEmpty();
```

However, since we add the entire project as a source folder by default, filtering out source folders curently means we end up with no choices. I would therefore suggest to remove the check for now.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,245 @@
...
+		protected boolean doCheckState() {
+			String fileName = getTextControl().getText();
+			fileName = fileName.trim();
+			if (fileName.isEmpty() && isEmptyStringAllowed()) {
+				return true;
+			}
+			return getProject().getFolder(new Path(fileName).lastSegment()).exists();
+		}
+
+		private Optional<IFolder> getDirectory() {
```
mx990: I would suggest to give this a more descriptive name, such as `chooseOutputFolder`.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,245 @@
...
+			setValidateStrategy(VALIDATE_ON_FOCUS_LOST);
+			createControl(parent);
+		}
+
+		@Override
+		protected String changePressed() {
+			final Optional<IFolder> selectedDirectory = getDirectory();
+
+			if (selectedDirectory.isPresent() && selectedDirectory.get().exists()) {
+				return selectedDirectory.get().getFullPath().toPortableString();
```
mx990: I would still suggest to only use the project-relative name as the value of the property and text field.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		private Optional<IFolder> getDirectory() {
+			final ElementTreeSelectionDialog fileDialog = new ElementTreeSelectionDialog(getShell(),
+					new WorkbenchLabelProvider(), new WorkbenchContentProvider());
+
+			fileDialog.setInput(getProject());
+			fileDialog.setAllowMultiple(false);
+			fileDialog.setTitle(Messages.TypeExport_OutputFolder);
+			fileDialog.setMessage(Messages.TypeExport_FileDialogMessage);
```
mx990: I would suggest to also set the initial selection to the current folder based on the text field.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,250 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Mario Kastner
+ *     - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+package org.eclipse.fordiac.ide.export.properties;
```
mx990: Please still move this to `org.eclipse.fordiac.ide.export.ui.properties`:
```suggestion
package org.eclipse.fordiac.ide.export.ui.properties;
```

plugins/org.eclipse.fordiac.ide.export/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,9 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.fordiac.ide.model,
  org.eclipse.fordiac.ide.ui,
  org.eclipse.core.resources,
- org.eclipse.fordiac.ide.util
+ org.eclipse.fordiac.ide.util,
+ org.eclipse.ui.workbench,
+ org.eclipse.ui.ide
```
mx990: Please also remove those unnecessary dependencies, since you moved the properties page to the `org.eclipse.fordiac.ide.export.ui` plugin.

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,250 @@
...
+		prefStore.setDefault(PreferenceConstants.OUTPUT_FOLDER, getOutputFolder());
+		return prefStore;
+	}
+
+	@Override
+	public boolean performOk() {
+		checkboxEditor.store();
+		exporterEditor.store();
+
+		if (!directoryEditor.isValid()) {
```
mx990: I believe this check should only be performed when the export is enabled?

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		@Override
+		protected boolean doCheckState() {
+			String fileName = getTextControl().getText();
+			fileName = fileName.trim();
+			if (fileName.isEmpty() && isEmptyStringAllowed()) {
+				return true;
+			}
+			return getProject().getFolder(new Path(fileName).lastSegment()).exists();
```
mx990: Why do you only use the last segment here? It works for me with:
```suggestion
			return getProject().getFolder(fileName).exists();
```

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/properties/TypeExportPropertyPage.java
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	@Override
+	public boolean performOk() {
+		checkboxEditor.store();
+		exporterEditor.store();
+
+		if (!directoryEditor.isValid()) {
+			directoryEditor.showErrorMessage();
+			FordiacLogHelper.logError(directoryEditor.getErrorMessage());
```
mx990: I do not believe we should log every attempt to save an invalid directory?

mx990 (changes requested):

mx990 (approved):

Issue #752, opened 2024-11-18
Spurious test failures in library tests


mx990:
There are spurious test failures in `org.eclipse.fordiac.ide.test.library`, see https://github.com/eclipse-4diac/4diac-ide/actions/runs/11895553157/job/33145397171 for an example.

pazi146:
fixed in #763

mx990:
It seems there is again a flaky test in `org.eclipse.fordiac.ide.test.library`, see https://github.com/eclipse-4diac/4diac-ide/actions/runs/12883969711.

mx990:
It seems the test also sometimes just hangs, see https://github.com/eclipse-4diac/4diac-ide/actions/runs/12987905232/job/36217739077.

mx990:
The tests seem to be stable for now.

PR #1001, opened 2025-01-27
Move edit cell on enter in variables dialog

mx990:
Move to edit the next cell below when closing the current editor by pressing enter in the variables dialog.


mx990:
Re-triggered workflows due to flaky test, see https://github.com/eclipse-4diac/4diac-ide/issues/752.

azoitl:
Did I understand right that enter brings me down only when the cell editor is closed. So when I'm editing I need to press enter twice.

mx990:
No, the same enter that is closing the editor moves you down.

azoitl (approved):
Sorry I tested in the wrong location. I read NAT I think of property sheets. Works as advertised.

PR #1002, opened 2025-01-27
Encapsulate global constants in namespace of enclosing type

mx990:
Encapsulate global constants in namespace of enclosing type:
- change ST parser to encapsulate global constants in namespace of enclosing type,
- add validation and quickfix for global constants declaration name,
- update proposals for global constants in ST editor to use qualified name and import enclosing type,
- update global constants in type library,
- add refactoring for global constants types,
- fix file editor input for ST editor.


data/typelibrary/utils-1.0.0/typelib/splitting/SPLIT_BYTE_INTO_QUARTERS.fct
```diff
@@ -10,26 +10,26 @@
...
 	<InterfaceList>
 		<EventInputs>
-			<Event Name="REQ" Type="Event" Comment="Normal Execution Request">
+			<Event Name="REQ" Type="Event">
 				<With Var="IN"/>
 			</Event>
 		</EventInputs>
 		<EventOutputs>
-			<Event Name="CNF" Type="Event" Comment="Execution Confirmation">
+			<Event Name="CNF" Type="Event">
```
oberlehner, line 18: why do you remove the comment here ?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
```diff
@@ -123,7 +123,7 @@ public Change createChange(final IProgressMonitor monitor) throws CoreException,
 			if (type instanceof FBType) {
 				return createFBDataChange();
 			}
```
oberlehner, line 125: i think you should check here explicit for global const.
If we rename for example .adp , .dev, .seg ,... this code will be executed. 
I don't know if that will work out

data/typelibrary/utils-1.0.0/typelib/splitting/SPLIT_BYTE_INTO_QUARTERS.fct
mx990: The interface of function FBs is (re-)generated by the ST reconciler on each save, so any comments are lost. But there should be no way to add any comments in the first place, besides editing the XML file manually.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
mx990: Do we not need to update those other type entries anyhow?

oberlehner: yes can make sense

PR #1003, opened 2025-01-28
LibraryManager didn't Correctly Handle Closing Projects

azoitl:
When closing project the library manager tried to resolve the dependencies of the closed project which resulted in a broken typelibrary instance. On project reopening this led to broken types.

Furthermore as part of this work we could remove a workspace job from the resouce change listener making it cleaner and hopefully more stable.


mx990 (approved):
With this PR, I am no longer able to reproduce the problem on my end.

PR #1010, opened 2025-01-30
Added line numbers in Model Search for ST Code

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchQuery.java
```diff
@@ -328,6 +332,31 @@ private boolean matchEObject(final INamedElement modelElement, final IProgressMo
...
+
+	public int[] getLineNumber(String input) {
+		final List<Integer> list = new ArrayList<>();
+
+		String searchString = modelQuerySpec.searchString();
+		if (!modelQuerySpec.checkCaseSensitive()) {
+			input = input.toLowerCase();
+			searchString = searchString.toLowerCase();
+		}
+		final String[] lines = input.split("\\R"); //$NON-NLS-1$
```
oberlehner: I think input.lines() is faster

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchQuery.java
```diff
@@ -328,6 +332,31 @@ private boolean matchEObject(final INamedElement modelElement, final IProgressMo
 		return false;
 	}
 
+	private int[] getLinesInSt(final Algorithm algo) {
+		final String text = getImplText(algo);
+		if (text == null || pattern.preScanST(text)) {
+			return getLineNumber(text);
```
oberlehner: should we enter this condition if the text is null ? That doesn't make sense to me

sebHollersbacher:
@mx990 Do you know a better way to get the line numbers of ST-Code for the model-search?
Simply comparing strings does not feel right here.

mx990:
Please have a look at the `org.eclipse.fordiac.ide.model.search.ISearchSupport`, `ISearchFactory`, and `TextMatch`.

sebHollersbacher:
@mx990 
I'm not sure if this is the right way to do it
Unfortunately, I could not use the SearchFactory/SearchSupports from `org.eclipse.fordiac.ide.model.search.st` without creating a circular dependency, maybe I'm doing something wrong?

mx990:
It seems the `org.eclipse.fordiac.ide.model.search.st` plug-in is missing from the corresponding `org.eclipse.fordiac.ide.structuredtext.feature`, so it was not included in 4diac IDE.

If you add it to the feature in [features/org.eclipse.fordiac.ide.structuredtext.feature/feature.xml](https://github.com/eclipse-4diac/4diac-ide/blob/develop/features/org.eclipse.fordiac.ide.structuredtext.feature/feature.xml), it should work without the additional extension, copied files, or a dependency to `org.eclipse.fordiac.ide.model.search.st` in the `org.eclipse.fordiac.ide.model.search` plug-in.

oberlehner:
image
I am searching for EINIT, which is in that case a PIN and an Algorithm.
I get bot results, which is correct.
The only thing missing here, is the line number of the algorithm.

All other cases are looking good :)

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchResultPage.java
```diff
@@ -269,11 +282,12 @@ private static boolean isInternalFb(final EObject eobj) {
 
 	protected static TableLayout createTableLayout() {
 		final TableLayout layout = new TableLayout();
-		layout.addColumnData(new ColumnPixelData(TYPE_COLUMN_WIDTH));
-		layout.addColumnData(new ColumnPixelData(NAME_COMMENT_COLUMN_WIDTH));
-		layout.addColumnData(new ColumnPixelData(NAME_COMMENT_COLUMN_WIDTH));
-		layout.addColumnData(new ColumnPixelData(TYPE_COLUMN_WIDTH));
-		layout.addColumnData(new ColumnPixelData(FULL_HIERARCHICAL_NAME_COLUMN_WIDTH));
+		layout.addColumnData(new ColumnWeightData(10));
```
oberlehner: Please replace that by constants with the name of the column like ELEMENT_KIND_COLUMN_WIDHT, NAME_COLUMN_WIDHT,etc..

oberlehner (approved):

PR #1017, opened 2025-02-03
Added update for plant hierachy after move subapp

oberlehner:
When moving untyped subapps through hierachy, the path of the plant hierarchy is adjusted accordingly now.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
@@ -25,13 +25,11 @@ import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
 
 public interface QualNameAffectedCommand extends ScopedCommand {
 
-	String getOldQualName(INamedElement elemt);
+	String getOldQualName(INamedElement element);
 
 	String getNewQualName(INamedElement element);
 
 	List<INamedElement> getChangedElements();
-
-	// private
 
 	/**
 	 * encapsulate the change to not provide the command to the receiver
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
@@ -36,7 +36,7 @@ public interface QualNameAffectedCommand extends ScopedCommand {
 	/**
 	 * encapsulate the change to not provide the command to the receiver
 	 */
-	default List<QualNameChange> getQualNameChange(final QualNameChangeState state) {
+	default List<QualNameChange> getQualNameChanges(final QualNameChangeState state) {
 
 		final List<QualNameChange> qualNameChanges = new ArrayList<>();
 
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
@@ -12,6 +12,9 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.model.commands;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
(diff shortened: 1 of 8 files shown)


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -22,18 +25,27 @@
...
-	INamedElement getChangedElement();
+	// private
 
 	/**
 	 * encapsulate the change to not provide the command to the receiver
 	 */
-	default QualNameChange getQualNameChange(final QualNameChangeState state) {
-		return new QualNameChange(getOldQualName(), getNewQualName(), getChangedElement(),
-				getTypeEntry(getChangedElement()), state);
+	default List<QualNameChange> getQualNameChange(final QualNameChangeState state) {
```
mx990:
```suggestion
	default List<QualNameChange> getQualNameChanges(final QualNameChangeState state) {
```

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -22,18 +25,27 @@
 
 public interface QualNameAffectedCommand extends ScopedCommand {
 
-	String getOldQualName();
+	String getOldQualName(INamedElement elemt);
```
mx990:
```suggestion
	String getOldQualName(INamedElement element);
```

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
```diff
@@ -22,18 +25,27 @@
...
+
+	String getNewQualName(INamedElement element);
 
-	String getNewQualName();
+	List<INamedElement> getChangedElements();
```
mx990, line 32: Should we better return a Set here?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeNameCommand.java
```diff
@@ -169,18 +171,18 @@ public Set<EObject> getAffectedObjects() {
 	}
 
 	@Override
-	public INamedElement getChangedElement() {
-		return getElement();
+	public List<INamedElement> getChangedElements() {
+		return Arrays.asList(getElement());
```
mx990: I would suggest using `List.of()` (or `Set.of()`) here.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeNameCommand.java
```diff
@@ -169,18 +171,18 @@ public Set<EObject> getAffectedObjects() {
...
 	}
 
 	@Override
-	public String getNewQualName() {
-		return getOldQualName().substring(0, getOldQualName().length() - oldName.length()) + name;
+	public String getOldQualName(final INamedElement elemt) {
+		return oldQualName;
```
mx990, line 183: I believe we should at least check the element here and below.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/QualNameAffectedCommand.java
oberlehner: no because most of the commands uses lists and therefore i don't want to change them every time

mx990: I am a bit confused, since you are creating a new list every time in https://github.com/eclipse-4diac/4diac-ide/blob/d9ee27a5b94016be363457921eeeb3dc1d995424/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/MoveAndReconnectCommand.java#L307-L309

oberlehner: Otherwise i would have to create a new list and cast every element. So this should not make a difference, since FBNetworkElement is instance of INamedElement

mx990:
Please rebase to exclude the unrelated commit https://github.com/eclipse-4diac/4diac-ide/pull/1017/commits/f94e29cc3c15c69a65c73e5e52bf8138a88b3622 from this PR.

oberlehner:
> Please rebase to exclude the unrelated commit [f94e29c](https://github.com/eclipse-4diac/4diac-ide/commit/f94e29cc3c15c69a65c73e5e52bf8138a88b3622) from this PR.

Done.
Thx for recognizing

mx990 (approved):

PR #1019, opened 2025-02-03
Added the attributes type to the search

oberlehner:
In order to update the **data type** of an attribute, it is necessary to consider attributes in the instance search

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
+++ b/plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
@@ -16,11 +16,16 @@ import java.util.stream.Stream;
 
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.fordiac.ide.model.data.StructuredType;
+import org.eclipse.fordiac.ide.model.libraryElement.Application;
+import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
 import org.eclipse.fordiac.ide.model.libraryElement.AttributeDeclaration;
 import org.eclipse.fordiac.ide.model.libraryElement.AutomationSystem;
 import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableFB;
+import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableObject;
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
```diff
@@ -93,6 +104,20 @@ public Stream<? extends EObject> getChildren(final EObject obj) {
...
+			}
+
+			if (obj instanceof final IInterfaceElement interfaceElement) {
+				return interfaceElement.getAttributes().stream();
+			}
+
+			if (obj instanceof final Attribute atrr) {
+				return Stream.of(atrr.getType());
+
+			}
```
mx990, line 128: I believe you also need to handle the remaining subclasses of `ConfigurableObject`, such as `Connection`? Maybe we should add a catch-all if for `ConfigurableObject` at the end?

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
```diff
@@ -81,9 +88,13 @@ public Stream<? extends EObject> getChildren(final EObject obj) {
 				return SearchChildrenProviderHelper.getFBTypeChildren(fbType);
```
mx990, line 89: I believe you also need to consider attributes in the appropriate helper methods of `SearchChildrenProviderHelper`?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
oberlehner: Makes sense

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/DataTypeInstanceSearch.java
oberlehner: Done

mx990 (approved):

PR #1034, opened 2025-02-10
Unnecessary space after comment

ernstblechaPT:
For some reason the comment-XML attribute had a stray whitespace after the attribute. Based on the examples I had available removing this whitespace should pose no problems.


ernstblechaPT:
To clear up why I made this change: For VarDeclarations if a comment is added and a InitialValue the space in between those attributes gets doubled up:

image

azoitl (approved):

PR #1050, opened 2025-02-12
Preserve Tree Top Item of Struct Manipulator Section

azoitl:
In order to preserve the struct manipulator's tree viewer correctly on refresh the tree path has to be taken and restored.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/StructManipulatorSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/StructManipulatorSection.java
@@ -20,6 +20,10 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.properties;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.application.Messages;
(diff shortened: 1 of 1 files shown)


bwiesmayr (approved):

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/StructManipulatorSection.java
```diff
@@ -124,15 +129,61 @@ private void createStructSelector(final Composite composite) {
...
+		if (topItemPath != null) {
+			Display.getDefault().asyncExec(() -> {
+				// Force tree refresh
+				memberViewerTree.update();
+				memberViewerTree.redraw();
+				final TreeItem topItem = getTreeItemFromPath(topItemPath);
+				if (topItem != null && !topItem.isDisposed()) {
+					final TreeItem[] originalSelection = memberViewerTree.getSelection();
+					memberViewerTree.setTopItem(topItem);
+					memberViewerTree.setSelection(originalSelection);
```
sebHollersbacher: Although it feels much better, when an item is selected it jumps to it when checking an item
Maybe switching these lines would be better

sebHollersbacher (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/StructManipulatorSection.java
azoitl: could you try that

azoitl: I tried it out in the mean-time and it feals much better. Thx for finding that.

sebHollersbacher (approved):

PR #1060, opened 2025-02-14
Add some all-purpose sequence Blocks

franz-hoepfinger-4diac:
6 Different Types of sequence FBs: 

- E --> Event triggered one-Shot
- E_looped --> Event triggered looped
- T --> Time triggered one-Shot
- T_looped --> Time triggered looped
- ET --> Event or Time triggered one-Shot
- ET_looped --> Event or Time triggered looped

### DIFF
--- a/data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
+++ b/data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<GlobalConstants Name="sequence" Comment="Global constants">
+	<Identification Standard="61499-1" Description="Copyright (c) 2024 HR Agrartechnik GmbH  &#10; &#10;This program and the accompanying materials are made  &#10;available under the terms of the Eclipse Public License 2.0  &#10;which is available at https://www.eclipse.org/legal/epl-2.0/  &#10; &#10;SPDX-License-Identifier: EPL-2.0">
+	</Identification>
+	<VersionInfo Organization="HR Agrartechnik GmbH" Version="1.0" Author="Franz Höpfinger" Date="2024-03-02">
+	</VersionInfo>
+	<CompilerInfo>
+	</CompilerInfo>
+	<GlobalConstants>
+		<VarDeclaration Name="NO_TIME" Type="TIME" Comment="" InitialValue="TIME#49d17h2m47s295ms"/>
+		<VarDeclaration Name="State_00" Type="SINT" Comment="" InitialValue="SINT#0"/>
+		<VarDeclaration Name="State_01" Type="SINT" Comment="" InitialValue="SINT#1"/>
+		<VarDeclaration Name="State_02" Type="SINT" Comment="" InitialValue="SINT#2"/>
+		<VarDeclaration Name="State_03" Type="SINT" Comment="" InitialValue="SINT#3"/>
+		<VarDeclaration Name="State_04" Type="SINT" Comment="" InitialValue="SINT#4"/>
+		<VarDeclaration Name="State_05" Type="SINT" Comment="" InitialValue="SINT#5"/>
+		<VarDeclaration Name="State_06" Type="SINT" Comment="" InitialValue="SINT#6"/>
+		<VarDeclaration Name="State_07" Type="SINT" Comment="" InitialValue="SINT#7"/>
+		<VarDeclaration Name="State_08" Type="SINT" Comment="" InitialValue="SINT#8"/>
+	</GlobalConstants>
+	<OriginalSource><![CDATA[VAR_GLOBAL CONSTANT
+	NO_TIME : TIME := TIME#49d17h2m47s295ms;
+	State_00 : SINT := SINT#0;
+	State_01 : SINT := SINT#1;
+	State_02 : SINT := SINT#2;
+	State_03 : SINT := SINT#3;
+	State_04 : SINT := SINT#4;
+	State_05 : SINT := SINT#5;
+	State_06 : SINT := SINT#6;
+	State_07 : SINT := SINT#7;
+	State_08 : SINT := SINT#8;
+END_VAR
+
+]]></OriginalSource>
+</GlobalConstants>
--- a/data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
+++ b/data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
@@ -19,7 +19,7 @@
 		<VarDeclaration Name="State_08" Type="SINT" Comment="" InitialValue="SINT#8"/>
 	</GlobalConstants>
 	<OriginalSource><![CDATA[VAR_GLOBAL CONSTANT
-	NO_TIME : TIME := TIME#49d17h2m47s295ms;
+	INF_TIME : TIME := TIME#-1ns;
 	State_00 : SINT := SINT#0;
 	State_01 : SINT := SINT#1;
 	State_02 : SINT := SINT#2;
--- a/data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
+++ b/data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
@@ -80,17 +80,17 @@
 				<ECAction Output="timeOut.START"/>
 			</ECState>
 			<ECState Name="sState_00" x="2666.67" y="4000">
+				<ECAction Output="timeOut.STOP"/>
 				<ECAction Algorithm="State_04_X" Output="EO_S4"/>
 				<ECAction Algorithm="State_00_C" Output="CNF"/>
 				<ECAction/>
-				<ECAction Output="timeOut.STOP"/>
 			</ECState>
 			<ECState Name="sRESET" x="-666.67" y="2666.67">
+				<ECAction Output="timeOut.STOP"/>
 				<ECAction Algorithm="State_01_X" Output="EO_S1"/>
 				<ECAction Algorithm="State_02_X" Output="EO_S2"/>
 				<ECAction Algorithm="State_03_X" Output="EO_S3"/>
 				<ECAction Algorithm="State_04_X" Output="EO_S4"/>
-				<ECAction Output="timeOut.STOP"/>
 			</ECState>
 			<ECTransition Source="sState_01" Destination="sState_02" Condition="timeOut.TimeOut" Comment="" x="2873.33" y="1686.67"/>
 			<ECTransition Source="sState_02" Destination="sState_03" Condition="timeOut.TimeOut" Comment="" x="2860" y="2373.33"/>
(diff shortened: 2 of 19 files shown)


franz-hoepfinger-4diac:
@bwiesmayr

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
```diff
@@ -0,0 +1,35 @@
...
+		<VarDeclaration Name="State_02" Type="SINT" Comment="" InitialValue="SINT#2"/>
+		<VarDeclaration Name="State_03" Type="SINT" Comment="" InitialValue="SINT#3"/>
+		<VarDeclaration Name="State_04" Type="SINT" Comment="" InitialValue="SINT#4"/>
+		<VarDeclaration Name="State_05" Type="SINT" Comment="" InitialValue="SINT#5"/>
+		<VarDeclaration Name="State_06" Type="SINT" Comment="" InitialValue="SINT#6"/>
+		<VarDeclaration Name="State_07" Type="SINT" Comment="" InitialValue="SINT#7"/>
+		<VarDeclaration Name="State_08" Type="SINT" Comment="" InitialValue="SINT#8"/>
+	</GlobalConstants>
+	<OriginalSource><![CDATA[VAR_GLOBAL CONSTANT
+	NO_TIME : TIME := TIME#49d17h2m47s295ms;
```
azoitl: why is this no time?

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
```diff
@@ -0,0 +1,35 @@
...
+	State_00 : SINT := SINT#0;
```
azoitl, line 23: can you give this states more meaningful names. Just numbers are not really helping to understand what they are for. Would an enum be better?

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
```diff
@@ -0,0 +1,204 @@
...
+				<With Var="DT_S1_S2"/>
+				<With Var="DT_S2_S3"/>
+				<With Var="DT_S3_S4"/>
+				<With Var="DT_S4_START"/>
+			</Event>
+			<Event Name="RESET" Type="Event" Comment="Reset from any State to START" >
+			</Event>
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="CNF" Type="Event" Comment="Execution Confirmation" >
```
azoitl, line 21: why do you have cnf and individual events for states?

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
```diff
@@ -0,0 +1,204 @@
...
+				<ECAction Output="timeOut.START"/>
+			</ECState>
+			<ECState Name="sState_00" x="2666.67" y="4000">
+				<ECAction Algorithm="State_04_X" Output="EO_S4"/>
+				<ECAction Algorithm="State_00_C" Output="CNF"/>
+				<ECAction/>
+				<ECAction Output="timeOut.STOP"/>
+			</ECState>
+			<ECState Name="sRESET" x="-666.67" y="2666.67">
+				<ECAction Algorithm="State_01_X" Output="EO_S1"/>
```
azoitl, line 90: do you really want to issue all EO events in case of reset?

azoitl (changes requested):

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
franz-hoepfinger-4diac: This is the largest thinkable Time. 
with this Constant you can "switch off" a Transition on Time. 

so for the ET Block you make that Transition then Event only.

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
franz-hoepfinger-4diac: do we have ENUMS now ?

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
franz-hoepfinger-4diac: CNF goes with the State Number, to display current Step in a Display, 
the individual Events go for the Individual Outputs
image

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
franz-hoepfinger-4diac: yes, otherwise the Outputs stay on if we go with the Concept you see in the Comment above. 

Question is if we want to reset the STATE_NR Output also to 0.

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
franz-hoepfinger-4diac: it could be similar seen as the 0xFFFF for 16-Bit INT for "no Value" which is commonly used in Bus Systems.

azoitl: This is definitely not the largest possible value for TIME. The largest value how Eclipse 4diac implements it is larger then 290000 years. So it is just some arbitrarily made up value.  Which is not good. Why not 0ns or -1ns? 

Also the name does not really convey the meaning that you use it for.

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
azoitl: With the 4days of Eclipse 4diac edition we have basic support for enums in 4diac IDE. However I just noticed that we do not yet have 4diac FORTE support and with that also no exporters. Also as I read below you need the value that is not possible in IEC 611313- with enums. There is a dedicated data type for that, which we are not supporting.

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
azoitl: interesting approach

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
azoitl: This is different to other FBs with Reset inputs.

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
franz-hoepfinger-4diac: -1ns i like most yes !

franz-hoepfinger-4diac: what do you think about:

`INF_TIME : TIME := TIME#-1ns;`

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
franz-hoepfinger-4diac: as my resarch no: 
E_TOF.fbt if we press 'R' all Outputs go to FALSE. 
do you know any where the Outputs stay on ?

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
azoitl: T#-1ns is a total valid value for time calling it INF_TIME seems wrong to me. Please note that the Time data type is a delta time so it can be positive and negative. As you need some magic number for your design you should choose one wisely and give it a name that fits to the magic that you are using it for.  A generic name will not work here.

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
azoitl: I'm not talking about the values I'm talking about the events that are sent on reset.

franz-hoepfinger-4diac: if we go so: 
image
then we need all this Events, 

if we don't, i would make it like E_CTU, where we have R and RO, 
to have a dedicated Reset Output Event ? 

right ?

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
franz-hoepfinger-4diac: we could also make CNF and RO like written below.

data/typelibrary/utils-1.0.0/typelib/sequence-control/time/sequence_T_04.fbt
azoitl: Yes this was what I wanted to point out. But as I do not really understand what the block is doing and what you need it for it is hard for me to say what a good design is.

franz-hoepfinger-4diac: Traffic Light Example: 
"Austria"
image

"International"
image

franz-hoepfinger-4diac: @azoitl as said, 
a all-purpose Seqence control, 
avoiding everybody must go the C++ way for simple Sequences. 

Pneumatic, Hydraulic, Traffic Light, etc..

franz-hoepfinger-4diac:
@azoitl 
sequence_ET_04.fbt
i have adjusted this one now. 
if OK, i will change all to this Schema. 

is more 61499-ish. 

Thanks, 
Franz

data/typelibrary/utils-1.0.0/typelib/sequence-control/sequence.gcf
franz-hoepfinger-4diac: > TIME

is it like this: 
https://infosys.beckhoff.com/english.php?content=../content/1033/tf6100_tc3_opcua_server/15548005131.html&id=1652078184739309755

213503d23h34m33s709ms551us615ns
upper Limit Time ?

franz-hoepfinger-4diac: no. as 4diac does not take this constant. at least in IDE.

franz-hoepfinger-4diac: The T#-1ns is replaced by the IDE to T#-1s999ms999us999ns
i will file a Issue.

franz-hoepfinger-4diac: https://github.com/eclipse-4diac/4diac-ide/issues/1349 issue here, but i need to fill out 
do not read it until it is announced in the Issue.

PR #1066, opened 2025-02-16
Add Event conversion FBTypes E_TO_EI and EI_TO_E.

franz-hoepfinger-4diac:
Added new FBTypes for converting between Event and EInit types.


azoitl:
Thx for this FBs. But for me these are not the solution to the original problem. At this point I first want to have a clear definition how event types should be used. And then we can come up with potential cast operators.

franz-hoepfinger-4diac:
for anybody stumbling over here: https://github.com/eclipse-4diac/4diac-ide/pull/1132

PR #1074, opened 2025-02-17
Update E_TABLE and E_TABLE_CTRL with event name changes and version info.

franz-hoepfinger-4diac:
- Renamed INIT to START in FBType E_TABLE and E_TABLE_CTRL
- Updated version info for both FBTypes


franz-hoepfinger-4diac:
rename to START, because this INIT is no Real INIT, and should not have type EInit.

azoitl:
The standard defines this block to have INIT as the name for that event. Can you therefore not change things that are defined in the standard.

diplfranzhoepfinger:
> The standard defines this block to have INIT as the name for that event. Can you therefore not change things that are defined in the standard.

ah, but then we LET it name INIT, but make the Type to Event ? 
right ?

diplfranzhoepfinger:
> The standard defines this block to have INIT as the name for that event. Can you therefore not change things that are defined in the standard.

yes, i Agree... Standard ist Standard ...

azoitl:
> > The standard defines this block to have INIT as the name for that event. Can you therefore not change things that are defined in the standard.
> 
> ah, but then we LET it name INIT, but make the Type to Event ? right ?

maybe

diplfranzhoepfinger:
Deal !

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-ide/pull/1084

PR #1077, opened 2025-02-18
logiBUS IO Module, including Support for ESP32 Button

franz-hoepfinger-4diac:
https://components.espressif.com/components/espressif/button


diplfranzhoepfinger:
@jwalt you asked for.

azoitl:
given the legal sitation with logibus I don't think this should be here.

franz-hoepfinger-4diac:
good i did it early! 

so Trademark-wise this is a Problem right ? Because logiBUS is a registered Trademark owned by HR Agrartechnik GmbH
(see https://register.dpma.de/DPMAregister/marke/register/3020232159100/DE)

BUT: I have to reply, that you use **other** Trademarks in the IO Folder too. Like WAGO, ADS, eliteboard, embrick, revpi. 

so, you think it is better to name these IOs "ESP32 IO" ? 

i was thinking logiBUS as a better name, as if we change CPU Platform one day, Customers can still use the same mapping FBs ? 

if you wish, i can rename them, but as well as Espressif and ESP32 are also Trademarks. So what difference does it make ?

Thanks,
Franz

azoitl:
You explained my once that the software license behind the libraries to access logibus is a problem. Or did I missunderstand something here?

diplfranzhoepfinger:
> You explained my once that the software license behind the libraries to access logibus is a problem. Or did I missunderstand something here?

this is the REASON for this commit. 

i had a "Universal" IO Library before, and now i split it up into several Libs, 
because Forte new IO allow to have several IO Libs. 

**this** one should only be for the "local" IOs on the ESP32 itself. 

another one will handle the ISOBUS related IOs, there we need to find the Solution how to interface with Commercial Libraries (TODO) and another one for BUS-Interface IOs from Datapanel.

franz-hoepfinger-4diac:
C++ Code here: 
https://github.com/eclipse-4diac/4diac-forte/pull/322

azoitl:

PR #1087, opened 2025-02-19
Bulk Editor

sebHollersbacher:


sebHollersbacher:
@azoitl Can you confirm if this is the right way to add the Open Editor option?

plugins/org.eclipse.fordiac.ide.bulkeditor/plugin.xml
```diff
@@ -0,0 +1,38 @@
...
+            id="org.eclipse.fordiac.ide.bulkeditor.BulkEditor"
+            name="Bulk Editor">
+      </editor>
+	</extension>
+	
+	<extension point="org.eclipse.ui.menus">
+    <menuContribution locationURI="popup:org.eclipse.ui.popup.any?after=group.openWith">
+        <command
+            commandId="org.eclipse.fordiac.ide.bulkeditor.editors.openEditor"
+            label="Open Bulk Editor">
```
azoitl: I think "Open in Bulk Editor" would be a better text.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/OpenBulkEditorHandler.java
```diff
@@ -0,0 +1,43 @@
...
+	@Override
+	public Object execute(final ExecutionEvent event) throws ExecutionException {
+		final IStructuredSelection selection = (IStructuredSelection) HandlerUtil.getCurrentSelection(event);
+		final Object firstElement = selection.getFirstElement();
+
+		if (firstElement instanceof final IProject project) {
+			final IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+
+			try {
+				window.getActivePage().openEditor(new NullEditorInput(), // new BulkEditorInput(project),
```
azoitl: do you need your own editor input or could you use the fileEditorInput here.

oberlehner: should we think of a file, similar like the .plant file for that ?
Maybe we want to restore something after closing the editor.

But i am not sure about that, i also don't want to make our life harder

azoitl: I'm not sure. The idea of the bulk editor is to edit several files across the whole project. having now a dedicated file feels a bit strange.

mx990: There is also the issue of what saving in the editor would mean then, that is, saving the affected types vs. saving the search to the file.

Maybe it would be enough to remember past searches in the workspace using a memento? We could also add import/export of searches in the future, similar to how it works for preferences or breakpoints?

azoitl: Yes I think memento or import export of searches would be a good idea.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorInput.java
```diff
@@ -0,0 +1,50 @@
...
+ *   Sebastian Hollersbacher - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.bulkeditor.editors;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.ui.IMemento;
+import org.eclipse.ui.part.FileEditorInput;
+
+public class BulkEditorInput extends FileEditorInput {
```
mx990: I do not believe `FileEditorInput` is an appropriate base class, since you do not have an actual file to edit. I would instead suggest to implement `IEditorInput` directly.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorSettings.java
```diff
@@ -0,0 +1,79 @@
...
+public class BulkEditorSettings {
+	private static final String TAG_BULKEDITOR_SETTINGS = "BULKEDITOR_SETTINGS"; //$NON-NLS-1$
+	private static final String MODE_TAG = "_mode"; //$NON-NLS-1$
+	private static final String NAME_SUB_SETTING_ADDITION = "_name_"; //$NON-NLS-1$
+
+	public int modeSelection = 0;
+	public BulkEditorSubSettings nameSubSettings = new BulkEditorSubSettings(NAME_SUB_SETTING_ADDITION);
+
+	public void saveState(final IMemento memento) {
+		memento.putInteger(TAG_BULKEDITOR_SETTINGS + MODE_TAG, modeSelection);
```
mx990: Please use child-mementos instead of assembling path names, see also `IMemento.createChild(String)`.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/FilterComposite.java
```diff
@@ -0,0 +1,83 @@
...
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Text;
+
+public class FilterComposite extends Composite {
+	public final Filter nameFilter;
+	public final Filter typeFilter;
+	public final Filter commentFilter;
+	public final Filter valueFilter;
```
mx990: I would suggest to make this filter composite extensible using an arbitrary list of filter fields, since we will most likely be  needing more/other fields in the future.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/OpenBulkEditorHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.core.resources.IProject;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.handlers.HandlerUtil;
+
+public class OpenBulkEditorHandler extends AbstractHandler {
+	@Override
+	public Object execute(final ExecutionEvent event) throws ExecutionException {
+		final IStructuredSelection selection = (IStructuredSelection) HandlerUtil.getCurrentSelection(event);
```
mx990:
```suggestion
		final IStructuredSelection selection = HandlerUtil.getCurrentStructuredSelection(event);
```

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/AttributeColumnAccessor.java
```diff
@@ -56,6 +57,8 @@ public Object getDataValue(final Attribute rowObject, final AttributeTableColumn
 		}
 		case VALUE -> InitialValueHelper.getInitialOrDefaultValue(rowObject);
 		case COMMENT -> CommentHelper.getInstanceComment(rowObject);
+		case LOCATION -> EcoreUtil.getURI(rowObject).toPlatformString(true);
+
```
mx990: Please avoid adding arbitrary empty lines.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/CopyElementRecord.java
```diff
@@ -0,0 +1,24 @@
...
+ *   Sebastian Hollersbacher - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.bulkeditor.editors;
+
+import java.util.List;
+
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+
+public record CopyElementRecord(LibraryElement copy, List<EObject> mappedElements) {
```
mx990: I would suggest to avoid field names in records that are also verbs, since the resulting getters are easily misinterpreted as actually performing an action (e.g., `copy()`).

mx990 (changes requested):

oberlehner:
image
Initial Value is cut off.
Can you increase the size a little bit in this column ?

sebHollersbacher:
> image Initial Value is cut off. Can you increase the size a little bit in this column ?

@oberlehner, should be fixed now

oberlehner (approved):

azoitl (approved):

PR #1089, opened 2025-02-20
Gloabl constants reference search

lindhuber:
Replaced the xtext find reference search with the 4diac model search in the global constants editor. It currently does not match on the fully qualified name (e.g. "value2" instead of "GlobalTest::value2") as otherwise it would not match the occurrences in the ST code anymore.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/META-INF/MANIFEST.MF
+++ b/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/META-INF/MANIFEST.MF
@@ -27,7 +27,10 @@ Require-Bundle: org.eclipse.fordiac.ide.globalconstantseditor,
  org.eclipse.fordiac.ide.typemanagement,
  org.eclipse.fordiac.ide.systemmanagement,
  org.eclipse.ui,
- org.eclipse.ui.editors
+ org.eclipse.ui.editors,
+ org.eclipse.ui.navigator,
+ org.eclipse.fordiac.ide.model.search,
+ org.eclipse.search
 Import-Package: org.apache.log4j,
  org.eclipse.ui,
  org.eclipse.xtext.ui.codemining;resolution:=optional
--- a/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
+++ b/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
@@ -206,7 +206,7 @@
 	</extension>
 	<extension point="org.eclipse.ui.handlers">
 		<handler
-			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.xtext.ui.editor.findrefs.FindReferencesHandler"
+			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.fordiac.ide.globalconstantseditor.ui.handlers.GlobalConstantsFindReferencesHandler"
 			commandId="org.eclipse.xtext.ui.editor.FindReferences">
 			<activeWhen>
 				<reference
--- a/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/handlers/GlobalConstantsFindReferencesHandler.java
+++ b/plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/handlers/GlobalConstantsFindReferencesHandler.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Primetals Technologies Austria GmbH
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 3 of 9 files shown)


plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/handlers/GlobalConstantsFindReferencesHandler.java
```diff
@@ -0,0 +1,62 @@
...
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.model.search.ModelQuerySpec;
+import org.eclipse.fordiac.ide.model.search.ModelQuerySpec.SearchScope;
+import org.eclipse.fordiac.ide.model.search.ModelSearchQuery;
+import org.eclipse.fordiac.ide.structuredtextcore.stcore.STVarDeclaration;
+import org.eclipse.search.ui.NewSearchUI;
+import org.eclipse.xtext.ui.editor.findrefs.FindReferencesHandler;
+
+@SuppressWarnings("restriction")
+public class GlobalConstantsFindReferencesHandler extends FindReferencesHandler {
```
oberlehner, line 29: Please give a more generic name. we might want to reuse that handler in other places too in the future.
Maybe FordiacFindReferencesHandler

mx990: I believe we need to use it also in the other ST languages to handle references to global constants from other ST sources. Please move it to the corresponding `structuredtextcore.ui` plugin. By the existing naming convention, it should then also be renamed to `STCoreFindReferencesHandler`.

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
```diff
@@ -206,7 +206,7 @@
 	</extension>
 	<extension point="org.eclipse.ui.handlers">
 		<handler
-			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.xtext.ui.editor.findrefs.FindReferencesHandler"
+			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.fordiac.ide.globalconstantseditor.ui.handlers.GlobalConstantsFindReferencesHandler"
```
mx990, line 209: Please change the handlers in the other ST languages, as well.

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/handlers/GlobalConstantsFindReferencesHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+	@Override
+	protected void findReferences(final EObject target) {
+		if (target instanceof final STVarDeclaration varDec) {
```
mx990, line 33: Please check that it is a global constant and not a local variable with the additional condition:
```suggestion
		if (target instanceof final STVarDeclaration varDec && varDec.eContainer() instanceof STVarGlobalDeclarationBlock) {
```

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/META-INF/MANIFEST.MF
```diff
@@ -27,7 +27,10 @@ Require-Bundle: org.eclipse.fordiac.ide.globalconstantseditor,
  org.eclipse.fordiac.ide.typemanagement,
  org.eclipse.fordiac.ide.systemmanagement,
  org.eclipse.ui,
- org.eclipse.ui.editors
+ org.eclipse.ui.editors,
+ org.eclipse.ui.navigator,
```
mx990, line 31: Do you actually need the `navigator` dependency?

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchQuery.java
```diff
@@ -282,11 +283,44 @@ private void matchStruct(final StructuredType struct, final IProgressMonitor mon
...
+	    )
+	    .forEach(searchResult::addResult);
+		// @formatter:on
+	}
+
+	private boolean searchInterfaceValue(final INamedElement modelElement) {
+		if (modelElement instanceof final VarDeclaration varDecl) {
+			final var value = varDecl.getValue();
+			return value != null && !value.getValue().isEmpty()
+					&& value.getValue().contains(modelQuerySpec.searchString());
```
mx990, line 307: In order to accurately search for global constants in packages, I would suggest to first strip any leading packages for simple string pre-matching and then use the ST search to determine if it is actually a match.

mx990 (changes requested):

diplfranzhoepfinger:
WOW ! i would like this. 

also it would be great to have a Function when i see a Constant, to jup to the Definition.

oberlehner:
I don't find refernces in interfaces of untyped subapps:
image

oberlehner:
The cleanup needs to be done by someone else, because the daniel ist available for the next months.
So i will merge it, because the feature does not break anything.
Nevertheless we should not forget about the clean up.

oberlehner (approved):

PR #1096, opened 2025-02-24
Reveal type sub editor

azoitl:


plugins/org.eclipse.fordiac.ide.fbtypeeditor.network/src/org/eclipse/fordiac/ide/fbtypeeditor/network/actions/OpenTypedFBNetworkEditorAction.java
```diff
@@ -0,0 +1,90 @@
...
+		switch (root) {
+		case final SubAppType subApp -> openInSubappTypeEditor(subApp, fbNetworkToOpen.eContainer());
+		case final FBType fbType -> openCFBNetworkEditor(fbType);
+		default -> {
+			/* nothing to do in default case */ }
+		}
+	}
+
+	private void openCFBNetworkEditor(final FBType fbType) {
+		openInFBTypeEditor(fbType, fbNetworkToOpen.eContainer());
```
PStmmr, line 51: Couldn't fbNetworkToOpen theoretically be zero and thus cause a problem here?

azoitl: we are only getting here if the root object searhec in line 40 is a fbType. This can only happen if fbNetworkToOpen is not null. So we should be save here.

oberlehner (approved):

PStmmr (approved):

PR #1097, opened 2025-02-24
Added changes regarding connection negation to the model.

PStmmr:
Since it shall be possible to invert the connection between two block pins of type boolean in the future, changes to the model were made.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/model/fordiac.genmodel
+++ b/plugins/org.eclipse.fordiac.ide.model/model/fordiac.genmodel
@@ -295,6 +295,10 @@
         <genParameters ecoreParameter="lib.ecore#//Connection/validateVarInOutConnectionsFormsNoLoop/diagnostics"/>
         <genParameters ecoreParameter="lib.ecore#//Connection/validateVarInOutConnectionsFormsNoLoop/context"/>
       </genOperations>
+      <genOperations ecoreOperation="lib.ecore#//Connection/setNegated" body="org.eclipse.fordiac.ide.model.libraryElement.impl.ConnectionAnnotations.setNegated(this, negated);">
+        <genParameters ecoreParameter="lib.ecore#//Connection/setNegated/negated"/>
+      </genOperations>
+      <genOperations ecoreOperation="lib.ecore#//Connection/isNegated" body="return org.eclipse.fordiac.ide.model.libraryElement.impl.ConnectionAnnotations.isNegated(this);"/>
     </genClasses>
     <genClasses ecoreClass="lib.ecore#//ConnectionRoutingData">
       <genFeatures createChild="false" ecoreFeature="ecore:EAttribute lib.ecore#//ConnectionRoutingData/dx1"/>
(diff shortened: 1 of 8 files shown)


plugins/org.eclipse.fordiac.ide.model/model/fordiac.genmodel
```diff
@@ -295,6 +295,10 @@
         <genParameters ecoreParameter="lib.ecore#//Connection/validateVarInOutConnectionsFormsNoLoop/diagnostics"/>
         <genParameters ecoreParameter="lib.ecore#//Connection/validateVarInOutConnectionsFormsNoLoop/context"/>
       </genOperations>
+      <genOperations ecoreOperation="lib.ecore#//Connection/setInverted" body="org.eclipse.fordiac.ide.model.libraryElement.impl.ConnectionAnnotations.setInverted(this, inverted);">
+        <genParameters ecoreParameter="lib.ecore#//Connection/setInverted/inverted"/>
```
oberlehner: please change to negated.
It should only be applicable to BOOL connections. so negated would be a better term in my opinion.

@azoitl what do you think ?

The rest looks fine 👍

PStmmr: Done

azoitl: Yes I also think it should be negated.

azoitl:
@PStmmr @oberlehner can connections have attributes? if yes I would prefer to store the negated value as attribute, similar how we do it with the visible flag of pins. Otherwise we pay for every connection instance for this boolean value even though we are not using it or it is not even allowed (e.g., struct connection)

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -348,4 +351,21 @@ static String getConnectionQualifiedName(final Connection connection) {
...
+		}
+	}
+
+	private static void setNegated(final HiddenElement connection, final String negated) {
+		connection.setAttribute(InternalAttributeDeclarations.NEGATED, negated, "");
+	}
+
+	public static boolean isNegated(final HiddenElement connection) {
+		final String negatedAttribute = connection.getAttributeValue(LibraryElementTags.CONNECTION_NEGATED);
+		return !"false".equalsIgnoreCase(negatedAttribute); //$NON-NLS-1$
```
mx990: It seems this makes the attribute true by default (e.g., when negatedAttribute is null or empty)?

mx990 (changes requested):

azoitl:
> @PStmmr @oberlehner can connections have attributes? if yes I would prefer to store the negated value as attribute, similar how we do it with the visible flag of pins. Otherwise we pay for every connection instance for this boolean value even though we are not using it or it is not even allowed (e.g., struct connection)

I just had a second look. It seems I  have missread the model change. It is already as I meant it to be.

mx990 (approved):

PR #1108, opened 2025-02-25
Fixed a case where refactoring can't happen even when read-only is unchecked

PStmmr:


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
```diff
@@ -31,9 +32,18 @@ protected boolean isSupported(final Object element) {
 			return isSupported(editPart.getModel());
 		}
 
-		return ((element instanceof final IInterfaceElement ie && !(element instanceof ErrorMarkerInterface))
-				&& (ie.getFBNetworkElement() instanceof final FB fb
-						&& !fb.getType().getTypeEntry().getFile().isReadOnly()))
-				|| (element instanceof final FB fbb && fbb.eContainer() instanceof BaseFBType);
+		if (element instanceof final IInterfaceElement ie && !(ie instanceof ErrorMarkerInterface)) {
+			if (ie.getFBNetworkElement() instanceof final FB fb) {
+				return !fb.getType().getTypeEntry().getFile().isReadOnly();
```
mx990: You can directly get the type entry from the FB network element. I believe we should also remove the instanceof to properly handle (typed) subapps. I would further suggest adding null checks, since the type entry may be null for untyped subapps and file may be null for error type entries.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
```diff
@@ -31,9 +32,18 @@ protected boolean isSupported(final Object element) {
...
+			}
+
+			if (ie.eContainer() instanceof final InterfaceList il
+					&& il.eContainer() instanceof final BaseFBType fbType) {
+				return !fbType.getTypeEntry().getFile().isReadOnly();
+			}
+			return false;
```
mx990: This would still exclude internal variables and interface elements inside subapp, CFB, and SIFB types.

I would suggest adding the following after line 33 to check if `element` is inside a read-only type instead:
```
if (element instanceof EObject eObject && EcoreUtil.getRootContainer(eObject) instanceof final LibraryElement libraryElement && libraryElement.getTypeEntry().getFile().isReadOnly()) {
  return false;
}
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
```diff
@@ -31,9 +35,25 @@ protected boolean isSupported(final Object element) {
...
+				&& EcoreUtil.getRootContainer(eObject) instanceof final LibraryElement libraryElement
+				&& libraryElement.getTypeEntry().getFile().isReadOnly()) {
+			return false;
+		}
+
+		if (element instanceof final IInterfaceElement ie && !(ie instanceof ErrorMarkerInterface)) {
+			if (ie.getFBNetworkElement() != null && ie.getFBNetworkElement().getType() != null
+					&& ie.getFBNetworkElement().getType().getTypeEntry() != null
+					&& ie.getFBNetworkElement().getType().getTypeEntry().getFile() != null) {
+				return !ie.getFBNetworkElement().getType().getTypeEntry().getFile().isReadOnly();
```
mx990: As mentioned before, you can directly get the type entry from the FB network element:
```suggestion
			if (ie.getFBNetworkElement() != null && ie.getFBNetworkElement().getTypeEntry() != null
					&& ie.getFBNetworkElement().getTypeEntry().getFile() != null) {
				return !ie.getFBNetworkElement().getTypeEntry().getFile().isReadOnly();
```
This is not only shorter but also avoids potentially (re-)loading the type unnecessarily.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
```diff
@@ -31,9 +35,25 @@ protected boolean isSupported(final Object element) {
...
+			}
+			if (ie.eContainer() instanceof final InterfaceList il && il.eContainer() instanceof final BaseFBType fbType
+					&& fbType.getTypeEntry() != null && fbType.getTypeEntry().getFile() != null) {
+				return !fbType.getTypeEntry().getFile().isReadOnly();
+			}
+			return false;
```
mx990: This still returns false for interface elements not in the interface list of a BFB. Now that you have the check in line 38, you should be able to remove this check here and just return true, as was the original behavior:
```suggestion
			return true;
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
```diff
@@ -31,9 +35,24 @@ protected boolean isSupported(final Object element) {
...
+
+		if (element instanceof final IInterfaceElement ie && !(ie instanceof ErrorMarkerInterface)) {
+			if (ie.getFBNetworkElement() != null && ie.getFBNetworkElement().getTypeEntry() != null
+					&& ie.getFBNetworkElement().getTypeEntry().getFile() != null) {
+				return !ie.getFBNetworkElement().getTypeEntry().getFile().isReadOnly();
+			}
+			if (ie.eContainer() instanceof final InterfaceList il && il.eContainer() instanceof final BaseFBType fbType
+					&& fbType.getTypeEntry() != null && fbType.getTypeEntry().getFile() != null) {
+				return !fbType.getTypeEntry().getFile().isReadOnly();
+			}
```
mx990: Is there a reason this check is still needed, considering the more general check in line 38?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/RefactorElementPropertyTester.java
PStmmr: No, I was thinking wrong by mistake.

mx990 (approved):

PR #1118, opened 2025-02-27
Export 4diac FORTE Core Includes with Full Source Directory

azoitl:
In order to allow the number of inlcude directory options core includes are now exported with their full 4diac FORTE source directory.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
@@ -37,11 +37,11 @@ abstract class ForteNgExportTemplate extends ExportTemplate {
 		«FOR include : types.map[generateDefiningInclude].toSet.sort»
 			«include.generateDependencyInclude»
 		«ENDFOR»
-		«generateDependencyInclude("iec61131_functions.h")»
-		«generateDependencyInclude("forte_array_common.h")»
-		«generateDependencyInclude("forte_array.h")»
-		«generateDependencyInclude("forte_array_fixed.h")»
-		«generateDependencyInclude("forte_array_variable.h")»
+		«generateDependencyInclude("core/iec61131_functions.h")»
+		«generateDependencyInclude("core/datatypes/forte_array_common.h")»
+		«generateDependencyInclude("core/datatypes/forte_array.h")»
+		«generateDependencyInclude("core/datatypes/forte_array_fixed.h")»
+		«generateDependencyInclude("core/datatypes/forte_array_variable.h")»
 	'''
 
 	def protected generateDependencyInclude(String path) {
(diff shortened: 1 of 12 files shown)


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
```diff
@@ -37,11 +37,11 @@ abstract class ForteNgExportTemplate extends ExportTemplate {
 		«FOR include : types.map[generateDefiningInclude].toSet.sort»
 			«include.generateDependencyInclude»
```
azoitl, line 38: @mx990 @MartinMelikMerkumians  do you have a tip for me how I can adjust this loop here to also give standard data type includes the `core/datatypes/` prefix.

jwalt (approved):
I am not much of an expert on the IDE code, but the changes look exactly like I would have expected them to be (in a genrally speaking way).

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
azoitl: I debugged a bit and found a place where it should not have to much impact on other parts. I hope this is correct

mx990: That is the same place I would have suggested. I guess you beat me to it :wink:.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
```diff
@@ -189,6 +189,8 @@ final class ForteNgExportUtil {
 
 	def static String generateTypePath(INamedElement type) {
 		switch (type) {
+			DataType case !(type instanceof AnyDerivedType):
```
mx990: I believe we need a slightly different condition here, since `AdapterType` is also a `DataType` and not an `AnyDerivedType` but should not go in `core/datatypes/`. The `AnyDerivedType` itself, on the other hand, is actually in `core/datatypes/forte_any_derived.h`.

I would suggest to instead test for:
```suggestion
			AnyType case type.typeEntry === null: 
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
azoitl: The condition was one of the tings a I struggled most and I totally forgot about the Adapter type. Therefore, thanks a lot for your suggestion will take it.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteNgExportTemplate.xtend
azoitl: Today in the morning I got curios and wanted to better understand our code generator. Happy to hear that I found the right spot.

mx990 (approved):

PR #1119, opened 2025-02-28
Fix: ensure correct internal name reflection in XML

### LINKED ISSUE #284: Bug in Resource Naming When Copied in System Explorer
### Description:
In the resources folder, there is a resource that appears correct. However, when right-clicking and selecting "Open in Type Editor," the internal name is shown incorrectly. This issue is specific to resources.

### Steps to Reproduce:

1. Locate a resource in the resources folder.
2. Right-click on the resource and select "Open in Type Editor."
3. Observe the internal name displayed in the type editor.
4. Copy the resource (emb_res) and paste it, renaming it to res2.
5. Check the XML of the copied resource and notice that the new name is not reflected correctly.

### Expected Behavior:
The internal name of the copied resource should reflect the new name correctly in the type editor and XML.

### Actual Behavior:
The internal name of the copied resource does not reflect the new name correctly in the type editor and XML.

### Screenshots:

**Before:**

image

**After copying and changing the name, it is not reflecting:**

image

anurag-sisodiya:
### 🔘Description
**Fixes #284** 
**Issue**:  When renaming a resource, the internal name in the XML and Type Editor does not reflect the new name correctly. This is because the save method in ResourceTypeEntryImpl was not implemented.

### 🔘Fix

-Implemented the save method to update the Name attribute in the XML file.

-Used a regex-based approach to directly update the Name attribute without modifying the rest of the XML structure.

-Added error handling for file read/write operations.

### 🔘Changes
**File Modified**: org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/ResourceTypeEntryImpl.java

### 🔘Steps to Verify

- Rename a resource in the System Explorer (e.g., emb_res ➔ new_name).

- Open the XML file and confirm: The Name attribute is updated (e.g., <ResourceType ... Name="new_name">).

- Open the resource in the Type Editor and verify that the internal name matches the new name.

### 🔘Demo Video:
Shows renaming and verifying the internal name

https://github.com/user-attachments/assets/dd27001d-5c14-4bf1-b27a-e1f0e5b0bc8c

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
@@ -1,0 +1,81 @@
+package org.eclipse.fordiac.ide.model.dataexport;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import javax.xml.stream.XMLStreamException;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.fordiac.ide.model.LibraryElementTags;
+import org.eclipse.fordiac.ide.model.libraryElement.CompilerInfo;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetwork;
+import org.eclipse.fordiac.ide.model.libraryElement.ResourceType;
+import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
+import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
+
+public class ResourceTypeExporter extends AbstractTypeExporter {
+
+	private final ResourceType resourceType;
+
+	public ResourceTypeExporter(final ResourceType type) {
+		super(type);
+		this.resourceType = type;
+	}
+
+	@Override
+	protected String getRootTag() {
+		return LibraryElementTags.RESOURCETYPE_ELEMENT;
+	}
+
+	@Override
+	protected void createXMLEntries() throws XMLStreamException {
+		getWriter().writeDTD(LINE_END
+				+ "<!DOCTYPE ResourceType SYSTEM \"http://www.holobloc.com/xml/LibraryElement.dtd\">" + LINE_END);
+		super.createXMLEntries();
+	}
+
+	@Override
+	protected void createTypeSpecificXMLEntries() throws XMLStreamException {
+
+		for (final VarDeclaration varDecl : resourceType.getVarDeclaration()) {
+			addVarDeclaration(varDecl);
+		}
+
+		exportFBNetwork();
+
+		final CompilerInfo compilerInfo = resourceType.getCompilerInfo();
+		if (compilerInfo != null) {
+			addCompilerInfo(compilerInfo);
+		}
+	}
+
+	private void exportFBNetwork() throws XMLStreamException {
+		final FBNetwork network = resourceType.getFBNetwork();
+		if (network != null && !network.getNetworkElements().isEmpty()) {
+			new FBNetworkExporter(this).createFBNetworkElement(network);
+		}
+	}
+
+	public void export(final IFile file, final IProgressMonitor monitor) throws CoreException {
+		final IProgressMonitor progressMonitor = monitor != null ? monitor : new NullProgressMonitor();
+		try (InputStream content = getFileContent()) {
+			if (file.exists()) {
+				file.setContents(content, true, true, progressMonitor);
+			} else {
+				file.create(content, true, progressMonitor);
+			}
+		} catch (final IOException e) {
+			FordiacLogHelper.logError("Error during resource type export", e);
+			throw new CoreException(Status.error("Error exporting ResourceType", e));
+		}
+	}
+
+	@Override
+	public ResourceType getType() {
+		return resourceType;
+	}
+}
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
@@ -32,13 +32,6 @@ public class ResourceTypeExporter extends AbstractTypeExporter {
 	}
 
 	@Override
-	protected void createXMLEntries() throws XMLStreamException {
-		getWriter().writeDTD(LINE_END
-				+ "<!DOCTYPE ResourceType SYSTEM \"http://www.holobloc.com/xml/LibraryElement.dtd\">" + LINE_END);
-		super.createXMLEntries();
-	}
-
-	@Override
 	protected void createTypeSpecificXMLEntries() throws XMLStreamException {
 
 		addCompilerInfo(getType().getCompilerInfo());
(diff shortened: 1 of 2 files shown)


azoitl (changes requested):
Thanks a lot for investigating the problem and finding the root cause of it. 

However the solution is unfortunately not adequate. The problem is that it breaks the contract of the save method for type entries. This contracts assumes that the content of the given `toSave` `LibraryElement` will be saved to file. With your fix only the type name is updated. 

In order to fix this an own `TypeExporter` would need to be implemented. With all the improvements we did in the last two years this is not that much work. The `DataTypeExporter` can sever as a starting point for that. 

@ANURAGSISODIYAA could you do such a change and improve this PR in that direction?

anurag-sisodiya:
Got it, I’ll implement a ResourceTypeExporter and update the save method accordingly.

azoitl:
> Got it, I’ll implement a ResourceTypeExporter and update the save method accordingly.

Cool! Thx!

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.fordiac.ide.model.LibraryElementTags;
+import org.eclipse.fordiac.ide.model.libraryElement.CompilerInfo;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetwork;
+import org.eclipse.fordiac.ide.model.libraryElement.ResourceType;
+import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
+import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
+
+public class ResourceTypeExporter extends AbstractTypeExporter {
+
+	private final ResourceType resourceType;
```
azoitl: you don't need to store the resourceType as additiona field. Please override the getType method and cast its return value to ResourceType

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/ResourceTypeEntryImpl.java
```diff
@@ -40,24 +36,11 @@ public ResourceTypeEntryImpl() {
 
 	@Override
 	public void save(final LibraryElement toSave, final IProgressMonitor monitor) throws CoreException {
```
azoitl: remove this method. by Providing a TypeExporter in getTypeExporter it is not needed anymore. This is all handled in the base-class

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	private void exportFBNetwork() throws XMLStreamException {
+		final FBNetwork network = resourceType.getFBNetwork();
+		if (network != null && !network.getNetworkElements().isEmpty()) {
+			new FBNetworkExporter(this).createFBNetworkElement(network);
+		}
+	}
+
+	public void export(final IFile file, final IProgressMonitor monitor) throws CoreException {
```
azoitl: Please remove this method. This is correctlyhandled by base-classes.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
...
+				file.create(content, true, progressMonitor);
+			}
+		} catch (final IOException e) {
+			FordiacLogHelper.logError("Error during resource type export", e);
+			throw new CoreException(Status.error("Error exporting ResourceType", e));
+		}
+	}
+
+	@Override
+	public ResourceType getType() {
```
azoitl, line 55: As written above please remove

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
+package org.eclipse.fordiac.ide.model.dataexport;
```
azoitl: you need to provide a copyright header for this file.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
...
+	protected void createTypeSpecificXMLEntries() throws XMLStreamException {
+
+		for (final VarDeclaration varDecl : resourceType.getVarDeclaration()) {
+			addVarDeclaration(varDecl);
+		}
+
+		exportFBNetwork();
+
+		final CompilerInfo compilerInfo = resourceType.getCompilerInfo();
+		if (compilerInfo != null) {
```
azoitl: this null check is performed in the addCompilerInfo method is not needed here.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,81 @@
...
+			addCompilerInfo(compilerInfo);
```
azoitl: please move addCompilerInfo to the top

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
anurag-sisodiya: What should be written in the copyright line?

 * Copyright (c)  What to write here?
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0.
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *   ANURAG SISODIYA - Initial implementation of ResourceTypeExporter
 *   Alois Zoitl - Code review and improvements

azoitl: the first line of the copyright statement sould for new files contain the year the file was created and the legal name of the copyright holder. Typically there are two options:

1.  If you are doing this privately (e.g., spare time, university student) then this is year name
2. If you are doing this as part of your work contract with a company then normally the company you are working for. Please make sure that in this case you have the permission of your company. With signing the Eclipse contributor Agreement you ensured that you did that. But just in case I wanted to repeat that. 


For the names on the bottom you don't need to add me. My support is not really copyright worthy.

anurag-sisodiya: Got it, thank you!

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/ResourceTypeExporter.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	protected String getRootTag() {
+		return LibraryElementTags.RESOURCETYPE_ELEMENT;
+	}
+
+	@Override
+	protected void createXMLEntries() throws XMLStreamException {
+		getWriter().writeDTD(LINE_END
```
azoitl: I just noticed that Github lost my comment to this line. For all the other elements we are normally not exporting the `DOCTYPE` entry anymore. Therefore I would find it better to not have it for resources as well.

azoitl (changes requested):

azoitl (approved):

PR #1121, opened 2025-03-03
Implement project specific settings

pazi146:
copied relevant parts from Eclipse JDT
added preferences cache to use in viewers


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractStructManipulatorEditPart.java
```diff
@@ -36,7 +37,8 @@ protected AbstractStructManipulatorEditPart() {
 
 	@Override
 	protected IFigure createFigureForModel() {
-		return new FBNetworkElementFigure(getModel());
+		return new FBNetworkElementFigure(getModel(),
+				((AdvancedScrollingGraphicalViewer) getViewer()).getPreferencesCache().getMaxTypeLabelSize());
```
mx990, line 41: Could we create a helper method for getting the preferences (cache)?

Maybe we could also get the preferences (cache) from the editor via an adapter, similar to the annotation model: https://github.com/eclipse-4diac/4diac-ide/blob/384d58454a68410064c8ff46e839a7ab411c646a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/FordiacAnnotationUtil.java#L71-L78

or we could also copy the relevant preferences to the viewer properties on open, see `org.eclipse.gef.EditPartViewer.getProperty(String)`?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
```diff
@@ -236,7 +237,9 @@ protected void createEditPolicies() {
 
 	@Override
 	protected IFigure createFigure() {
-		final FBNetworkConnection connectionFigure = new FBNetworkConnection(this);
+		final var cache = ((AdvancedScrollingGraphicalViewer) getViewer()).getPreferencesCache();
+		final FBNetworkConnection connectionFigure = new FBNetworkConnection(this,
+				cache.getMaxHiddenConnectionLabelSize(), cache.getPinLabelStyle());
```
mx990, line 250: Maybe we should pass the combined preferences directly to the various figures, in order to make adding additional preferences easier in the future?

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/editparts/WatchValueDirectEditPolicy.java
```diff
@@ -41,6 +43,7 @@ public class WatchValueDirectEditPolicy extends DirectEditPolicy {
 	@Override
 	protected Command getDirectEditCommand(final DirectEditRequest request) {
 		final String value = (String) request.getCellEditor().getValue();
+		// TODO
```
mx990: Please fix or remove this todo note.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
```diff
@@ -40,11 +46,26 @@ public class ToggleBoolValueHandler extends AbstractHandler {
...
+		final IEditorInput input = editor.getEditorInput();
+		if (input instanceof final IContentEditorInput contentInput) {
+			if (EcoreUtil.getRootContainer(contentInput.getContent()) instanceof final LibraryElement libElement) {
+				project = libElement.getTypeEntry().getFile().getProject();
+			}
+		} else if (input instanceof final IFileEditorInput fileInput) {
+			project = fileInput.getFile().getProject();
+		}
+		final boolean monitoringValueWriteThrough = DeploymentDebugPreferences.INSTANCE.getStore(project)
+				.getBoolean(DeploymentDebugPreferences.MONITORING_VALUE_WRITE_THROUGH);
```
mx990: Please extract this into a separate method.

plugins/org.eclipse.fordiac.ide.deployment.debug/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,7 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.fordiac.ide.ui,
  org.eclipse.debug.core,
  org.eclipse.jface,
+ org.eclipse.ui,
```
mx990: Please do not add a UI dependency to a non-UI plugin.

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferenceInitializer.java
```diff
@@ -13,13 +13,13 @@
...
+import org.eclipse.jface.preference.IPreferenceStore;
 
 public class DeploymentDebugPreferenceInitializer extends AbstractPreferenceInitializer {
 
 	@Override
 	public void initializeDefaultPreferences() {
-		DefaultScope.INSTANCE.getNode(DeploymentDebugPreferences.QUALIFIER)
-				.putBoolean(DeploymentDebugPreferences.MONITORING_VALUE_WRITE_THROUGH, false);
+		final IPreferenceStore store = DeploymentDebugPreferences.STORE;
+		store.setDefault(DeploymentDebugPreferences.MONITORING_VALUE_WRITE_THROUGH, false);
```
mx990: Why do we need to use the preference store here, since this is just setting a workspace default?

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
```diff
@@ -12,18 +12,52 @@
...
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.ui.preferences.ScopedPreferenceStore;
 
-public final class DeploymentDebugPreferences {
+public enum DeploymentDebugPreferences implements IPreferenceStoreProvider {
+	INSTANCE;
 
 	public static final String QUALIFIER = "org.eclipse.fordiac.ide.deployment.debug"; //$NON-NLS-1$
+
+	public static final IPreferenceStore STORE = new ScopedPreferenceStore(InstanceScope.INSTANCE, QUALIFIER);
```
mx990: Please use the preferences service from the platform to get project-specific preferences in a non-UI context, instead of the JFace preference store. There is an example on how to implement this in:
https://github.com/eclipse-4diac/4diac-ide/blob/384d58454a68410064c8ff46e839a7ab411c646a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/validation/ValidationPreferences.java#L63-L69

I would also suggest moving the code to set preferences to the respective preference page.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
```diff
@@ -1075,4 +1075,22 @@
...
 	</extension>
+ <extension
+       point="org.eclipse.ui.propertyPages">
+    <page
+		category="org.eclipse.fordiac.ide.debug.ui.preferences.FordiacDebugPreferencePage"
+		class="org.eclipse.fordiac.ide.deployment.debug.ui.preferences.DeploymentDebugPreferencePage"
+		id="org.eclipse.fordiac.ide.deployment.debug.ui.properties"
+		name="Deployment">
+	</page>
+	<enabledWhen>
```
mx990: It seems the `enabledWhen` is outside the `page` element?

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
```diff
@@ -1075,4 +1075,22 @@
 			modelId="org.eclipse.fordiac.ide.deployment.debug.model">
 		</variableValueEditor>
 	</extension>
+ <extension
+       point="org.eclipse.ui.propertyPages">
+    <page
+		category="org.eclipse.fordiac.ide.debug.ui.preferences.FordiacDebugPreferencePage"
```
mx990: Please make sure the category points to a valid *property* page, same in the other plugins.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/ValueEditPart.java
```diff
@@ -57,24 +56,15 @@
...
 	private InitialValueRefreshJob refreshJob;
 
-	private static int maxWidth = -1;
-
-	public static int getMaxWidth() {
-		if (-1 == maxWidth) {
-			final int maxLabelSize = GefPreferenceConstants.STORE.getInt(GefPreferenceConstants.MAX_VALUE_LABEL_SIZE);
-			final FontMetrics fm = FigureUtilities.getFontMetrics(JFaceResources.getFontRegistry()
-					.get(org.eclipse.fordiac.ide.ui.preferences.UIPreferenceConstants.DIAGRAM_FONT));
-			maxWidth = (int) ((maxLabelSize + 2) * fm.getAverageCharacterWidth());
```
mx990, line 75: Are you sure this is no longer needed?

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,8 @@ Require-Bundle: org.eclipse.emf.ecore.editor,
  org.eclipse.ui.views.properties.tabbed,
  org.eclipse.ui.navigator,
  org.eclipse.fordiac.ide.ui,
- org.eclipse.fordiac.ide.model.commands
+ org.eclipse.fordiac.ide.model.commands,
+ org.eclipse.xtext.ui
```
mx990: Why do we need an Xtext dependency here?

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
```diff
@@ -0,0 +1,298 @@
...
+	private SelectionButtonDialogField fUseProjectSettings;
+	private Composite fParentComposite;
+	protected IProject fProject;
+	private Map<String, Object> fData; // page data
+
+	public static final String DATA_NO_LINK = "PropertyAndPreferencePage.nolink"; //$NON-NLS-1$
+
+	protected IPreferenceStoreProvider provider;
+
+	protected FordiacPropertyPreferencePage(final IPreferenceStoreProvider provider) {
```
mx990: I would suggest to remove the unused constructors for now. We can always add them later, should the need arise.

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
```diff
@@ -0,0 +1,298 @@
...
+						Messages.PropertyAndPreferencePage_useworkspacesettings_change);
+				fChangeWorkspaceSettings.setLayoutData(new GridData(SWT.END, SWT.CENTER, false, false));
+			} else {
+				LayoutUtil.setHorizontalSpan(fUseProjectSettings.getSelectionButton(null), 2);
+			}
+
+			final Label horizontalLine = new Label(composite, SWT.SEPARATOR | SWT.HORIZONTAL);
+			horizontalLine.setLayoutData(new GridData(GridData.FILL, GridData.FILL, true, false, 2, 1));
+			horizontalLine.setFont(composite.getFont());
+		}
```
mx990: Please also add the hyperlink to navigate from the workspace to the project settings, see [org.eclipse.jdt.internal.ui.preferences.PropertyAndPreferencePage:134-137](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7d7dab5c6599dd6901fb14ba5e088023c0405fe6/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/PropertyAndPreferencePage.java#L134-L137).

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/GefPreferenceConstants.java
```diff
@@ -12,17 +12,23 @@
...
+public enum GefPreferenceConstants implements IPreferenceStoreProvider {
+	INSTANCE;
+
 	public static final String GEF_PREFERENCES_ID = "org.eclipse.fordiac.ide.gef"; //$NON-NLS-1$
 
 	public static final IPreferenceStore STORE = new ScopedPreferenceStore(InstanceScope.INSTANCE, GEF_PREFERENCES_ID);
-
-	private GefPreferenceConstants() {
-	}
+	private static final Map<IProject, IPreferenceStore> storeMap = new HashMap<>();
```
mx990: The `IPreferenceStore` is not intended to be shared by multiple users/threads, since changes have to be `org.eclipse.jface.preference.IPersistentPreferenceStore.save()`ed, among other things.

Maybe we should copy the `org.eclipse.xtext.ui.editor.preferences.FixedScopedPreferenceStore` and use that instead, which would solve the memory leak issue of the original `ScopedPreferenceStore`, see also https://bugs.eclipse.org/bugs/show_bug.cgi?id=239033 ? That would also make it a fairly lightweight object and we could avoid caching them altogether.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractStructManipulatorEditPart.java
azoitl: When we discussed that feature we found the PreferencesCache class as the better solution for future extensions we had in mind. Also it gives a nicer API then `getProperty`. 

But I like the idea with the helper method and the getAdapter.

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
pazi146: I'm using some UI elements in Xtext that got copied from Eclipse JDT to avoid more copying

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
pazi146: These constructors exist to make it a mostly transparent replacement for `FieldEditorPreferencePage`

Should I still remove them?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/GefPreferenceConstants.java
pazi146: I'll have a look at that

There still needs to be some caching as there are ChangeListeners attached - should I move that back to each class then?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
pazi146: I didn't put `IGraphicalPreferencesCache` as a parameter to avoid coupling and doing it with a map or something similar is also clunky.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/editparts/WatchValueDirectEditPolicy.java
pazi146: will remove

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferenceInitializer.java
pazi146: that was done to standardise everything as it doesn't make a difference in our use case

plugins/org.eclipse.fordiac.ide.deployment.debug/META-INF/MANIFEST.MF
pazi146: this is the same as in the `org.eclipse.fordia.ide.model` package to gain access to `org.eclipse.ui.preferences.ScopedPreferenceStore`

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
pazi146: I'll look into that

> I would also suggest moving the code to set preferences to the respective preference page.

what do you mean by this? Moving the preference store access out of the preference page was the intention

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/ValueEditPart.java
pazi146: thanks for catching that - it got lost in all that refactoring

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
pazi146: Left that out previously but can add it in

Do you know if there exists an equivalent project picker already or if I need to port that over too?

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
pazi146: done

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
pazi146: fixed

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
pazi146: Should I make an empty 4diac property page to provide a category for all property pages or just remove the category from them?

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
mx990: I see the danger of more and more Xtext dependencies creeping into our core plugins over time, when we originally wanted to keep Xtext entirely separate. However, I also don't like copying and we already have a lot of Xtext dependencies in other core plugins anyhow.

What do you think @azoitl?

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
mx990: We seem to currently use only `FordiacPropertyPreferencePage(int, IPreferenceStoreProvider)` and I am not sure if we will ever need the others. In the meantime, this is lots of duplicated code. As such, I would still remove them, yes.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/GefPreferenceConstants.java
mx990: I believe we have to keep a separate instance of the store in each class that registers a change listener, so we can remove it later, if that is what you mean.

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferenceInitializer.java
mx990: It does make a difference, because you are introducing a UI dependency in a non-UI plugin. Please change it back to the way it was.

plugins/org.eclipse.fordiac.ide.deployment.debug/META-INF/MANIFEST.MF
mx990: I realize this may seem overly pedantic on my part, but the separation between UI and non-UI functionality is a basic tenet of good software architecture.

As such, we should not use the JFace preference store in the `org.eclipse.fordiac.ide.model` plugin either. Unfortunately, that plugin still needs a UI dependency to convert between screen and 61499 corrdinates. This makes it no less problematic but less noticeable there.

In any case, we should definitely not add more UI dependencies in non-UI plugins, particularly when there are proper alternatives that could and should be used instead, see `Platform.getPreferencesService()`.

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
mx990: Please remember that you have to *save* the preferences after changing them, whichever approach you use. 

If you would like to keep this here, please use `IEclipsePreferences` and make sure to `flush()` the preference afterwards.

However, I believe it would be more consistent to change the preferences only in the page and then `IPersistentPreferenceStore.save()` them all there together.

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
mx990: I believe you may be able to use the `ElementListSelectionDialog` for that purpose, see also `NewSourceFolderWizardPage.chooseProject()` for an example on how it is used by JDT for the same purpose.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
mx990: I believe we should make empty pages if needed so that we have the same overall structure in the project properties as in the workspace preferences.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
azoitl: I'm also not a big fan of having xtext everywhere. Copying is also not the super cool. I see two options:

1. As we need that kind of code in several places we could provide our (copied) implementation in ide.ui
2. We use the xtext implementation but not import the whole xtext.ui plugin via require bundle but use `Import-Package` and only fetch the minimum packages that we need.

mx990: I had another look at we are actually using from Xtext here:
- `org.eclipse.xtext.ui.editor.tasks.dialogfields.DialogField` is a small base class for the already copied `SelectionButtonDialogField`,
- `org.eclipse.xtext.ui.editor.tasks.dialogfields.IDialogFieldListener` is a functional interface used in `DialogField`,
- `org.eclipse.xtext.ui.editor.tasks.dialogfields.LayoutUtil.setHorizontalGrabbing(Control)` and `org.eclipse.xtext.ui.editor.tasks.dialogfields.LayoutUtil.setHorizontalSpan(Control, int)` are each used only once in `FordiacPropertyPreferencePage` and could probably be inlined.

In light of this, I believe we should go with option 1. We should then probably also move the generic `SelectionButtonDialogField` to `org.eclipse.fordiac.ide.ui`.

plugins/org.eclipse.fordiac.ide.deployment.debug/META-INF/MANIFEST.MF
pazi146: removed the dependency

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferenceInitializer.java
pazi146: reverted

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
pazi146: removed static stores - will now get instanced wherever needed

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/plugin.xml
pazi146: added an empty property page

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
pazi146: copied/moved these classes into `org.eclipse.fordiac.ide.ui`

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
pazi146: removed unused constructors

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/FordiacPropertyPreferencePage.java
pazi146: added the project select - thanks for the hint

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/GefPreferenceConstants.java
pazi146: replaced `ScopedPreferenceStore` with a copied `FixedScopedPreferenceStore` (in `org.eclipse.fordiac.ide.ui`)

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideDataConnections.java
```diff
@@ -29,7 +30,8 @@ public class HideDataConnections extends AbstractHandler {
 	public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final Command command = event.getCommand();
 		final boolean oldValue = HandlerUtil.toggleCommandState(command);
-		UIPreferenceConstants.STORE.setValue(UIPreferenceConstants.P_HIDE_DATA_CON, !oldValue);
+		InstanceScope.INSTANCE.getNode(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID)
+				.putBoolean(UIPreferenceConstants.P_HIDE_DATA_CON, !oldValue);
```
mx990: Please remember to `flush()` the preference node after changing a preference.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideEventConnections.java
```diff
@@ -32,7 +33,8 @@ public class HideEventConnections extends AbstractHandler {
 	public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final Command command = event.getCommand();
 		final boolean oldValue = HandlerUtil.toggleCommandState(command);
-		UIPreferenceConstants.STORE.setValue(UIPreferenceConstants.P_HIDE_EVENT_CON, !oldValue);
+		InstanceScope.INSTANCE.getNode(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID)
+				.putBoolean(UIPreferenceConstants.P_HIDE_EVENT_CON, !oldValue);
```
mx990: Please remember to `flush()` the preference node after changing a preference.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/ToggleJumpStepHandler.java
```diff
@@ -27,7 +28,8 @@ public class ToggleJumpStepHandler extends AbstractHandler {
 	public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final Command command = event.getCommand();
 		final boolean oldValue = HandlerUtil.toggleCommandState(command);
-		UIPreferenceConstants.STORE.setValue(UIPreferenceConstants.P_TOGGLE_JUMP_STEP, !oldValue);
+		InstanceScope.INSTANCE.getNode(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID)
+				.putBoolean(UIPreferenceConstants.P_TOGGLE_JUMP_STEP, !oldValue);
```
mx990: Please remember to `flush()` the preference node after changing a preference.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
```diff
@@ -40,11 +47,14 @@ public class ToggleBoolValueHandler extends AbstractHandler {
 	public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final IEditorPart editor = HandlerUtil.getActiveEditorChecked(event);
 		final CommandStack stack = editor.getAdapter(CommandStack.class);
-		getValues(HandlerUtil.getCurrentStructuredSelection(event)).forEachOrdered(value -> toggleValue(value, stack));
+
+		getValues(HandlerUtil.getCurrentStructuredSelection(event))
+				.forEachOrdered(value -> toggleValue(value, stack, getMonitoringValueWriteThrough(editor)));
```
mx990: Please extract the result of `getMonitoringValueWriteThrough(editor)` out of the lambda into a variable, so it is not executed for each value.

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
```diff
@@ -77,4 +87,18 @@ private static Stream<WatchValueAnnotation> getValues(final IStructuredSelection
...
+	private static boolean getMonitoringValueWriteThrough(final IEditorPart editor) {
+		IProject project = null;
+		final IEditorInput input = editor.getEditorInput();
+		if (input instanceof final IContentEditorInput contentInput) {
+			if (EcoreUtil.getRootContainer(contentInput.getContent()) instanceof final LibraryElement libElement) {
+				project = libElement.getTypeEntry().getFile().getProject();
+			}
+		} else if (input instanceof final IFileEditorInput fileInput) {
+			project = fileInput.getFile().getProject();
+		}
```
mx990, line 102: Please also handle the case when it is neither kind of input, which would currently throw an exception because the project would be null. I would suggest to just return false then.

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
```diff
@@ -12,18 +12,13 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.deployment.debug.preferences;
 
-import org.eclipse.core.runtime.preferences.InstanceScope;
-
-public final class DeploymentDebugPreferences {
+public class DeploymentDebugPreferences {
```
mx990: Why did you make this non-final?

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
```diff
@@ -12,18 +12,13 @@
...
-
-	public static boolean isMonitoringValueWriteThrough() {
-		return InstanceScope.INSTANCE.getNode(QUALIFIER).getBoolean(MONITORING_VALUE_WRITE_THROUGH, false);
-	}
 
+	public static final String MONITORING_VALUE_WRITE_THROUGH = "monitoringValueWriteThrough"; //$NON-NLS-1$
+	
 	private DeploymentDebugPreferences() {
-		throw new UnsupportedOperationException();
+
```
mx990: Why did you remove the throw?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editors/InitialValueEditor.java
```diff
@@ -81,7 +82,10 @@ protected void updateInitialValue(final String value) {
 				&& FordiacMessages.ComputingPlaceholderValue.equals(getModelAccess().getEditablePart())) {
 			final var commandExecutorCache = getCommandExecutor();
 			setCommandExecutor(null);
-			if (value.length() <= DiagramPreferencePage.getMaxDefaultValueLength()) {
+			if (value.length() <= PreferenceStoreProvider
+					.getStore(GefPreferenceConstants.GEF_PREFERENCES_ID,
+							getInterfaceElement().getFBNetworkElement().getTypeLibrary().getProject())
```
mx990: I believe `getFBNetworkElement()` may return null if it is not an instance pin and `getTypeLibrary()` may return null if an instance has no type (e.g., untyped subapps). I would suggest to get the project via `EcoreUtil.getRootContainer(getInterfaceElement()) instanceof LibraryElement` instead.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/PreferenceStoreProvider.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	public void removePropertyChangeListener(final IPropertyChangeListener listener) {
+		instanceStore.removePropertyChangeListener(listener);
+		if (projectStore != null) {
+			projectStore.removePropertyChangeListener(listener);
+		}
+	}
+
+	public static IPreferenceStore getStore(final String qualifier, final IProject project) {
```
mx990, line 133: I would suggest to also add a convenience method that accepts an `EObject` as context and determines the project similar to `TypeLibraryManager.getTypeLibraryFromContext(EObject)`.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/PaletteFlyoutPreferences.java
```diff
@@ -36,37 +37,40 @@ public PaletteFlyoutPreferences(final String paletteDockLocationID, final String
...
 	public int getPaletteWidth() {
-		return GefPreferenceConstants.STORE.getInt(paletteSizeID);
+		return InstanceScope.INSTANCE.getNode(GefPreferenceConstants.GEF_PREFERENCES_ID).getInt(paletteSizeID, 0);
 	}
 
 	@Override
 	public void setDockLocation(final int location) {
-		GefPreferenceConstants.STORE.setValue(paletteDockLocationID, location);
+		InstanceScope.INSTANCE.getNode(GefPreferenceConstants.GEF_PREFERENCES_ID).putInt(paletteDockLocationID,
+				location);
```
mx990: Please remember to `flush()` the preference node after changing a preference.

plugins/org.eclipse.fordiac.ide.model.ui/META-INF/MANIFEST.MF
mx990: Please also remove the dependency.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/preferences/TypeManagementPreferenceConstants.java
```diff
@@ -16,13 +16,10 @@
 /**
  * Constant definitions for plug-in preferences.
  */
-public final class TypeManagementPreferenceConstants {
+public class TypeManagementPreferenceConstants {
```
mx990: Why did you make this class non-final?

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/FordiacPropertyPreferencePage.java
```diff
@@ -0,0 +1,318 @@
...
+		}
+		data.put(DATA_NO_LINK, Boolean.TRUE);
+
+		if (isProjectPreferencePage()) {
+			openWorkspacePreferences(data);
+		} else {
+			final List<IProject> projects = new LinkedList<>();
+			for (final IProject proj : ResourcesPlugin.getWorkspace().getRoot().getProjects()) {
+				try {
+					if (proj.getNature("org.eclipse.fordiac.ide.systemmanagement.FordiacNature") != null) { //$NON-NLS-1$
```
mx990:
```suggestion
					if (proj.hasNature("org.eclipse.fordiac.ide.systemmanagement.FordiacNature")) { //$NON-NLS-1$
```

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/FordiacPropertyPreferencePage.java
```diff
@@ -0,0 +1,318 @@
...
+						projects.add(proj);
+					}
+				} catch (final CoreException e) {
+					// empty
+				}
+			}
+			final ElementListSelectionDialog dialog = new ElementListSelectionDialog(getShell(),
+					new FordiacProjectLabelProvider());
```
mx990: I believe you may be able to use the existing `org.eclipse.ui.model.WorkbenchLabelProvider` here.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/ToggleJumpStepHandler.java
pazi146: done

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
pazi146: yeah, that's better

plugins/org.eclipse.fordiac.ide.deployment.debug.ui/src/org/eclipse/fordiac/ide/deployment/debug/ui/handler/ToggleBoolValueHandler.java
pazi146: Shouldn't ever be anything else, but it's better to be safe

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
pazi146: got lost as this was changed into an enum singleton in between

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideDataConnections.java
pazi146: done

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideEventConnections.java
pazi146: done

plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/preferences/DeploymentDebugPreferences.java
pazi146: same as above

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/PaletteFlyoutPreferences.java
pazi146: thanks

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/FordiacPropertyPreferencePage.java
pazi146: Thanky you for finding that - it fits perfectly

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/preferences/TypeManagementPreferenceConstants.java
pazi146: same as above - got lost in transition

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/FordiacPropertyPreferencePage.java
pazi146: changed

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/preferences/PreferenceStoreProvider.java
pazi146: added

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editors/InitialValueEditor.java
pazi146: rewrote the if

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideDataConnections.java
```diff
@@ -19,17 +19,29 @@
...
 		final Command command = event.getCommand();
 		final boolean oldValue = HandlerUtil.toggleCommandState(command);
-		UIPreferenceConstants.STORE.setValue(UIPreferenceConstants.P_HIDE_DATA_CON, !oldValue);
+		final IEclipsePreferences prefs = InstanceScope.INSTANCE
+				.getNode(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID);
+		prefs.putBoolean(UIPreferenceConstants.P_HIDE_DATA_CON, !oldValue);
+		try {
+			prefs.flush();
+		} catch (final BackingStoreException e) {
+			return new Status(IStatus.ERROR, getClass(), Messages.HandlerPreferenceSafeError);
```
mx990: The return value of handlers is reserved for future use and should always be null, see the JavaDoc of `org.eclipse.core.commands.IHandler.execute(ExecutionEvent)`.

I would suggest to use `org.eclipse.fordiac.ide.ui.FordiacLogHelper.logError(String, Throwable)` with the exception instead.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/PaletteFlyoutPreferences.java
```diff
@@ -36,37 +39,49 @@ public PaletteFlyoutPreferences(final String paletteDockLocationID, final String
...
+		setValue(paletteSizeID, width);
+	}
+
+	private static void setValue(final String valueID, final int value) {
+		final IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode(GefPreferenceConstants.GEF_PREFERENCES_ID);
+		prefs.putInt(valueID, value);
+		try {
+			prefs.flush();
+		} catch (final BackingStoreException e) {
+			// empty
```
mx990: I would suggest to log the exception with `org.eclipse.fordiac.ide.ui.FordiacLogHelper.logError(String, Throwable)`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/HideDataConnections.java
pazi146: done

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/preferences/PaletteFlyoutPreferences.java
pazi146: done

mx990 (approved):

PR #1127, opened 2025-03-04
Define 4diac IDE preferences export

pazi146:


plugins/org.eclipse.fordiac.ide.model/plugin.xml
```diff
@@ -291,4 +291,41 @@
...
+          <entry node="org.eclipse.fordiac.ide.gef"/>
+          <entry node="org.eclipse.fordiac.ide.gitlab"/>
+          <entry node="org.eclipse.fordiac.ide.globalconstantseditor.GlobalConstants"/>
+          <entry node="org.eclipse.fordiac.ide.globalconstantseditor.ui"/>
+          <entry node="org.eclipse.fordiac.ide.model"/>
+          <entry node="org.eclipse.fordiac.ide.structuredtextalgorithm.STAlgorithm"/>
+          <entry node="org.eclipse.fordiac.ide.structuredtextalgorithm.ui"/>
+          <entry node="org.eclipse.fordiac.ide.structuredtextcore.STCore"/>
+          <entry node="org.eclipse.fordiac.ide.structuredtextcore.ui"/>
+          <entry node="rg.eclipse.fordiac.ide.structuredtextfunctioneditor.STFunction"/>
```
mx990: I believe you are missing an `o` at the beginning?

plugins/org.eclipse.fordiac.ide.model/plugin.xml
```diff
@@ -291,4 +291,41 @@
           name="LIB">
     </attribute>
  </extension>
+ <extension
+       point="org.eclipse.ui.preferenceTransfer">
```
mx990: I would suggest to move this extension to `org.eclipse.fordiac.ide.ui`.

plugins/org.eclipse.fordiac.ide.model/plugin.xml
```diff
@@ -291,4 +291,41 @@
...
+    <transfer
+          id="org.eclipse.fordiac.ide.model.projectTransfer"
+          name="%exportName">
```
mx990: Could we please also add the 4diac icon for the transfer?

plugins/org.eclipse.fordiac.ide.model/plugin.properties
```diff
@@ -18,3 +18,4 @@
 
 pluginName = 4diac IDE Model Plug-in
 providerName = Eclipse 4diac
+exportName=4diac IDE Properties
```
mx990: Please rename this to `4diac IDE Preferences` to conform with the naming of the other transfers.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/plugin.xml
pazi146: yes ^^

plugins/org.eclipse.fordiac.ide.model/plugin.xml
pazi146: makes sense - though I don't know the proper link as this icon is in `.ide` and not in `.ide.images`

plugins/org.eclipse.fordiac.ide.model/plugin.xml
pazi146: will move it there

plugins/org.eclipse.fordiac.ide.model/plugin.xml
azoitl: you can try if our icon url is working: `fordiacimage://ICON_4DIAC_16`

pazi146: This doesn't show the 4diac logo but a white X on red background

azoitl: I just noticed that the 4diac 16 px icon get lost. could you add the attached icon to the images and update the `fordiacimages.properties` accordingly.

If it then still not working it can be that our url is not working for that extension point we have to use the platform url: `platform:/plugin/org.eclipse.fordiac.ide.images/images/...` 


4diac16

pazi146: Works

Note: the 4diac 32 px icon is also missing

azoitl (approved):

plugins/org.eclipse.fordiac.ide.model/plugin.xml
azoitl: Thx for pointing that out. Need to check if this is even needed anymore.

PR #1140, opened 2025-03-07
Changed Storage of Struct_Demux configuration to use % Separator

azoitl:
In order to treat a struct child pin of a struct demuxer as single entity the internal string and configuration string uses now the % separator. In the ui still the . is shown.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
+++ b/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
@@ -1773,13 +1773,24 @@
   <eClassifiers xsi:type="ecore:EClass" name="MemberVarDeclaration" eSuperTypes="#//VarDeclaration">
     <eOperations name="getName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
-        <details key="body" value="return ConfigurableFBManagement.getMemberVarName(this);"/>
+        <details key="body" value="return ConfigurableFBManagement.getMemberVarName(this, ConfigurableFBManagement.MEMBER_VAR_SEPARATOR);"/>
+      </eAnnotations>
+    </eOperations>
+    <eOperations name="getDisplayName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="return ConfigurableFBManagement.getMemberVarName(this, &quot;.&quot;);   //$NON-NLS-1$"/>
       </eAnnotations>
     </eOperations>
     <eOperations name="getVarName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
         <details key="body" value="return super.getName();"/>
       </eAnnotations>
+    </eOperations>
+    <eOperations name="setName">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="// ensure that we are always only setting the last segment as name&#xA;final String[] subNames = ConfigurableFBManagement.splitMemberVarName(name);&#xA;super.setName(subNames[subNames.length - 1]);"/>
+      </eAnnotations>
+      <eParameters name="name" lowerBound="1" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
     </eOperations>
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="parentNames" upperBound="-1"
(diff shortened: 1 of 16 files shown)


plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -1773,14 +1773,25 @@
   <eClassifiers xsi:type="ecore:EClass" name="MemberVarDeclaration" eSuperTypes="#//VarDeclaration">
     <eOperations name="getName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
-        <details key="body" value="return ConfigurableFBManagement.getMemberVarName(this);"/>
+        <details key="body" value="return ConfigurableFBManagement.getMemberVarName(this, ConfigurableFBManagement.MEMBER_VAR_SEPARATOR);"/>
+      </eAnnotations>
+    </eOperations>
+    <eOperations name="getUIName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
```
mx990: I would suggest renaming this to `getDisplayName` instead. Maybe we could also add it directly in `INamedElement`, which would delegate to `getName` by default and we could then override it here?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CheckableStructTree.java
```diff
@@ -24,12 +24,13 @@
 public class CheckableStructTree extends AbstractStructTree<CheckableStructTreeNode> {
 
 	public CheckableStructTree(final StructManipulator manipulator, final StructuredType struct,
-			final TreeViewer viewer) {
-		super(manipulator, struct, viewer);
+			final TreeViewer viewer, final String childSeparator) {
```
mx990: Does the separator actually need to be configurable? It seems it is always set to `%` anyhow?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: I find this idea interesting. However this would require to check all getNames. Therefore I would like to do this in a separate PR.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CheckableStructTree.java
azoitl: I was not sure. Currently this model is only used for the property sheet of muxers. But that may change in the future. Just changing the separtor would be much easier.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CheckableStructTreeNode.java
```diff
@@ -44,7 +45,8 @@ public CheckableStructTree getTree() {
 	@Override
 	public CheckableStructTreeNode addChild(final EObject memberVariable) {
 		if (memberVariable instanceof final VarDeclaration vardeclaration) {
-			final CheckableStructTreeNode treeNode = new CheckableStructTreeNode(vardeclaration, this, getTree());
+			final CheckableStructTreeNode treeNode = new CheckableStructTreeNode(vardeclaration, this, getTree(),
```
oberlehner: can you extract the getChildSepartor in a factory method, so that we don't have to propate the standard value through the whole tree data structure.

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: Changed it to `getDisplayName`. Had a quick look into moving it up the hierarchy. We use `getName` more then 1800 times. Just a quick look already showed  not clearly decidable situations. I think this is something for another time.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CheckableStructTree.java
azoitl: Based on @oberlehner I reworked the tree a second time. Now it is easier to hook an own separator and I made the default now the %.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CheckableStructTreeNode.java
azoitl: Good idea. I reworked the tree nodes such that it is easier to customize it later.

mx990 (approved):

m-meingast (approved):
Looks good to me

bwiesmayr (approved):

PR #1141, opened 2025-03-08
Fix SubApp expanded view sizing issues

### LINKED ISSUE #402: Size of fresh expanded subapp
After creating a new subapp and expanding it (empty) it looks like this:
image
After increasing the size to the bottom, it looks like this:
image
This is two base problems:

- The default settings for the side bar are _huge_ sidebar
- The size calculation is incorrect

anurag-sisodiya:
Fixes #402 

### Problem:

- Expanded SubApps had overly wide sidebars by default.

- Empty SubApps collapsed into a thin line due to incorrect size calculation.

### Solution:

- Reduced the minimum width of input/output sidebars in SubAppForFbNetworkFigure.

- Enforced minimum dimensions for empty SubApps in SubAppForFBNetworkEditPart.

### Testing:

- Create a new SubApp → Expand it. Verify sidebars are narrower.

- Resize an empty SubApp → Ensure it retains a usable size (min 100x60).

https://github.com/user-attachments/assets/ce240467-72cf-4dae-ad55-f0177ca4ddd3

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/SubAppForFBNetworkEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/SubAppForFBNetworkEditPart.java
@@ -384,8 +384,12 @@ public class SubAppForFBNetworkEditPart extends AbstractFBNElementEditPart imple
 
 	private Dimension getSubappSize() {
 		if (getModel().isUnfolded()) {
-			return new Dimension(CoordinateConverter.INSTANCE.iec61499ToScreen(getModel().getWidth()),
+			final int minWidth = 100;
+			final int minHeight = 60;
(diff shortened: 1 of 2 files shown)


azoitl:
thank your for finding this issue and the proposed fix. However the size of the side bar is based on preferences given by the user (or the according default values). With your change this preferences are not reflected in the sidebar size anymore. 

However I like the idea of giving the expanded subapp the first time it is expanded a better size. As the size is stored in the model and when you press save on the disk you should not perform any size corrections in the editpart but better calculate the right size in the `ToggleSubAppRepresentationCommand`. 

Please be also aware that the properties framework is changing: https://github.com/eclipse-4diac/4diac-ide/pull/1121
This may affect where you get the sizes for the interfacebar of expanded subapps.

bwiesmayr:
Hi, I agree that there is some problems with the size calculation. I have also noticed that sometimes (esp. if there is just one function block inside) a contained block is half-hidden behind the interface bar. But Alois' comments are of course still valid!

azoitl:
closing in favor of #1928

PR #1150, opened 2025-03-12
preserve resource mapping when creating contracts

lindhuber:


bwiesmayr (approved):
thanks a lot that was on the roadmap already

plugins/org.eclipse.fordiac.ide.contracts/src/org/eclipse/fordiac/ide/contracts/UpdateContractCommand.java
```diff
@@ -83,10 +88,14 @@ public void redo() {
...
 			toggle.redo();
 		}
+		if (mappingcmd != null) {
+			mappingcmd.redo();
+		}
 		ccc.redo();
 	}
 
 	private SubApp createNewSubapp() {
+		final var resource = fbNetworkElement.getResource();
```
azoitl, line 98: @lindhuber unfortunately using getResource here can be dangerous. When you are inside of a mapped subapp then you will also get a resource here and map the subapp that is inside of the subapp. It would be better to use the `isMapped` is method for checking if you need to map and cache the resource here. Could you please provide a second PR.

bwiesmayr: ... and sorry for merging too fast, it was exactly what I had planned to do so I accepted it quickly

lindhuber: Sure!

azoitl: thx

PR #1154, opened 2025-03-12
Updated exporter to new StringId handling

azoitl:
The FORTE exporter is creating the new USE_STRING_ID and STRID() for all string ids.

@mx990 as string ids are only used for external accessibility I miss the imagination why and how languages can provide string ids and have left that out. 

I hope I have not missed any ids.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteLibraryElementTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteLibraryElementTemplate.xtend
@@ -70,6 +70,13 @@ abstract class ForteLibraryElementTemplate<T extends LibraryElement> extends For
 
 	def protected generateIncludeGuardEnd() '''
 	'''
+	
+	def protected generateImplIncludes() '''
+		#include "«fileBasename».h"
+		
+		«getDependencies(emptyMap).generateDependencyIncludes»
+		«type.compilerInfo?.header»
+	'''
 
 	def protected generateVariableDeclarations(List<VarDeclaration> variables, boolean const) '''
 		«FOR variable : variables AFTER '\n'»
@@ -146,6 +153,14 @@ abstract class ForteLibraryElementTemplate<T extends LibraryElement> extends For
 	def CharSequence generateVariableTypeSpec(VarDeclaration decl) {
 		variableLanguageSupport.get(decl)?.generate(#{ForteNgExportFilter.OPTION_TYPE_SPEC -> Boolean.TRUE})
 	}
+	
+	def protected generateUseStringId() '''
+		«getUsedStrings(emptyMap).generateUseStringIdDecls»
+	'''
+	
+	def generateUseStringIdDecls(Set<String> strings) '''
+		«FOR str : strings.sort»USE_STRING_ID(«str»);«ENDFOR»
+	'''
 
 	def protected getFORTENameList(List<? extends INamedElement> elements) {
 		elements.map[name.FORTEStringId].join(", ")
@@ -173,5 +188,9 @@ abstract class ForteLibraryElementTemplate<T extends LibraryElement> extends For
 
 	def Set<INamedElement> getDependencies(Map<?, ?> options) {
 		variableLanguageSupport.values.filterNull.flatMap[getDependencies(options)].toSet
+	}
+	
+	def protected Set<String> getUsedStrings(Map<?, ?> options) {
+		newHashSet(type.generateTypeNamePlain)
 	}
 }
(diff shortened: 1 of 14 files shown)


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteLibraryElementTemplate.xtend
```diff
@@ -146,6 +153,14 @@ abstract class ForteLibraryElementTemplate<T extends LibraryElement> extends For
 	def CharSequence generateVariableTypeSpec(VarDeclaration decl) {
 		variableLanguageSupport.get(decl)?.generate(#{ForteNgExportFilter.OPTION_TYPE_SPEC -> Boolean.TRUE})
 	}
+	
+	def protected generateUseStringId() '''
+		«getUsedStrings(emptyMap).generateUseStringIdDecls»
+	'''
+	
+	def generateUseStringIdDecls(Set<CharSequence> strings) '''
+		«FOR str : strings»USE_STRING_ID(«str»);«ENDFOR»
```
mx990: I would suggest to also sort the set of string ids, similar to the includes.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/base/BaseFBImplTemplate.xtend
```diff
@@ -140,4 +142,12 @@ abstract class BaseFBImplTemplate<T extends BaseFBType> extends ForteFBTemplate<
 			]
 		).toSet
 	}
+	
+	override Set<CharSequence> getUsedStrings(Map<?, ?> options) {
+		val strings = super.getUsedStrings(options)
+		type.internalVars.forEach[getUsedIEStrings(it, strings)]
+		type.internalConstVars.forEach[getUsedIEStrings(it, strings)]
```
mx990: I believe we do not use any string ids for `internalConstVars`?

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/composite/CompositeFBImplTemplate.xtend
```diff
@@ -314,4 +317,19 @@ class CompositeFBImplTemplate extends ForteFBTemplate<CompositeFBType> {
 			getDependencies(options)
 		]).toSet
 	}
+	
+	override Set<CharSequence> getUsedStrings(Map<?, ?> options) {
+		val strings = super.getUsedStrings(options)
+		type.FBNetwork.networkElements.forEach[getUsedFBStrings(it, strings)]		
+		type.FBNetwork.eventConnections.forEach[getUsedConStrings(it, strings)]		
+		type.FBNetwork.dataConnections.forEach[getUsedConStrings(it, strings)]		
+		type.FBNetwork.adapterConnections.forEach[getUsedConStrings(it, strings)]
```
mx990, line 326: I believe we also need the names of instance variables with initial values, see https://github.com/eclipse-4diac/4diac-ide/blob/98c5cc1f73180174231372e948b14dc8e3699767/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/composite/CompositeFBImplTemplate.xtend#L284

mx990 (changes requested):
I believe we are missing a `generateUseStringId` in `FunctionFBImplTemplate`?

mx990:
> @mx990 as string ids are only used for external accessibility I miss the imagination why and how languages can provide string ids and have left that out.

The `VarDeclarationSupport` also uses string ids when generating the type spec for interface variables. You already handle those in the FB templates, so I agree we can leave that out.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/base/BaseFBImplTemplate.xtend
azoitl: true, will remove

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/composite/CompositeFBImplTemplate.xtend
azoitl: Thx for noting. To be sure I checked a second time where `FORTEStringId` is used and I think now we have everything.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteLibraryElementTemplate.xtend
azoitl: good idea this makes the code better readable.

azoitl:
> I believe we are missing a `generateUseStringId` in `FunctionFBImplTemplate`?

Thx for noting. Also fixed. We have here quite some code duplication but I was not creative enough to find a solution for it.

mx990 (approved):

PR #1155, opened 2025-03-13
Improve contract creation workflow

Felix-Schmid:
* add new dialog to choose from contract rules based on selected pins
* the selected pins will be added and connected to the subapp if not part of it already
* use CellEditor instead of dialog for contract editing


plugins/org.eclipse.fordiac.ide.contracts/src/org/eclipse/fordiac/ide/contracts/model/helpers/ContractElementBuilder.java
```diff
@@ -13,7 +13,10 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.contracts.model.helpers;
```
bwiesmayr, line 16: In this class, I'd think you should add yourself to the copyright header

bwiesmayr (changes requested):

plugins/org.eclipse.fordiac.ide.contracts/src/org/eclipse/fordiac/ide/contracts/model/helpers/ContractElementBuilder.java
Felix-Schmid: Ok, I added myself.

bwiesmayr (approved):

PR #1158, opened 2025-03-17
Introducing a new error marker for negated connections.

PStmmr:


azoitl (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -133,7 +134,7 @@ public static boolean validateTypeMismatch(final Connection connection, final Di
 			return true; // ignore incomplete connections
 		}
 		// basic type check
-		if (!LinkConstraints.typeCheck(src, dest)) {
+		if (!LinkConstraints.typeCheck(src, dest) && !isNegated(connection)) {
```
mx990: I would suggest to move this to the condition in line 133 above to entirely ignore type checks for negated connections.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -352,6 +353,21 @@ private ConnectionAnnotations() {
 		throw new UnsupportedOperationException("Helper class must not be instantiated"); //$NON-NLS-1$
 	}
 
+	public static boolean validateNegatedConnection(final Connection connection, final DiagnosticChain diagnostics,
+			final Map<Object, Object> context) {
+		if (connection.isNegated() && (!connection.getSource().getType().equals(IecTypes.ElementaryTypes.BOOL)
+				|| !connection.getDestination().getType().equals(IecTypes.ElementaryTypes.BOOL))) {
```
mx990: Please also check that the source and destination are not VarInOut pins and also add null checks for the source and destination, see also line 131-135 above.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
mx990: Also please use `connection.isNegated()` instead of calling the annotation implementation directly.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -352,6 +353,35 @@ private ConnectionAnnotations() {
...
+		}
+		return true;
+	}
+
+	private static boolean hasInOutVar(final Connection connection) {
+		if ((connection.getSource() instanceof final VarDeclaration vd && vd.isInOutVar())
+				|| (connection.getDestination() instanceof final VarDeclaration vd2 && vd2.isInOutVar())) {
+			return true;
+		}
+		return false;
```
mx990: I would suggest to simplify this:
```suggestion
		return (connection.getSource() instanceof final VarDeclaration vd && vd.isInOutVar())
				|| (connection.getDestination() instanceof final VarDeclaration vd2 && vd2.isInOutVar());
```

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -352,6 +353,35 @@ private ConnectionAnnotations() {
...
+	public static boolean validateNegatedConnection(final Connection connection, final DiagnosticChain diagnostics,
+			final Map<Object, Object> context) {
+		if (connection == null || connection.getSource() == null || connection.getDestination() == null) {
+			return false;
+		}
+
+		if (connection.isNegated()
+				&& (!connection.getSource().getType().equals(IecTypes.ElementaryTypes.BOOL)
+						|| !connection.getDestination().getType().equals(IecTypes.ElementaryTypes.BOOL))
+				&& !hasInOutVar(connection)) {
```
mx990: I believe this should be:
```suggestion
				&& (!connection.getSource().getType().equals(IecTypes.ElementaryTypes.BOOL)
						|| !connection.getDestination().getType().equals(IecTypes.ElementaryTypes.BOOL)
						|| hasInOutVar(connection))) {
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
PStmmr: Shouldn't all three conditions be met for the error marker to appear? And there **shouldn't** be any inout vars since you removed the '!', or have I misunderstood that?

mx990: My original comment may have been misleading. Negated connections are *not supported* for inout vars, so there *should* be an error marker if either source or destination is an inout var.

mx990 (approved):

PR #1168, opened 2025-03-18
Temporarily disable RichTextEditor

pazi146:
RichTextEditor causes a stack overflow on windows with Eclipse 2025-03

Reinstate once this issue is fixed:
https://github.com/EclipseNebula/nebula/issues/631

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.doc/plugin.xml
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.doc/plugin.xml
@@ -1,11 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
-	<extension
+	<!-- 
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.fbtypeeditor.doc/plugin.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
-	<extension
+	<!--extension
```
oberlehner, line 10: please add your comment here too: 
Otherwise we might forget about that
```
RichTextEditor causes a stack overflow on windows with Eclipse 2025-03

Reinstate once this issue is fixed:
https://github.com/EclipseNebula/nebula/issues/631
```

pazi146: done

azoitl (approved):

PR #1173, opened 2025-03-19
fix potential null pointer in LibraryManager

pazi146:
also use getRawLocation to avoid unnecessarily resolving links


azoitl:
Is the problem that the path is null or is the problem somewhere else and this null check is just curing the symptom and not the cause.

pazi146:
`getRawLocation()` should now always produce a non-null path, as it's always called on a linked folder.

An internal error might still cause it to return null - which would then mean no accessible path and as such no preferred version

azoitl (approved):
Thx for the clarification. As this is a bug fix a cherry picked it to the freeze branch

PR #1186, opened 2025-03-26
Make libraries read only

pazi146:
External library contents are set read only on extraction 
Standard library contents are set read only through decorate job 
Delete participant stops deletion of library contents 
Move strings into correct Messages class
Add read only behaviour to editor
> Future PR: implement doSaveAs()


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -331,6 +340,77 @@ private static Path newPath(final Path destinationDir, final ZipEntry zipEntry)
...
+					PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ);
+			try {
+				posixView.setPermissions(permissions);
+			} catch (final IOException e) {
+				// empty
+			}
+		}
+	}
+
+	private void setStandardLibsReadOnly() {
```
oberlehner, line 364: why only standard libs ? Shouldnt external libs be read only too?

plugins/org.eclipse.fordiac.ide.library.ui/plugin.properties
```diff
@@ -1,4 +1,32 @@
-#
+ ################################################################################
+ # Copyright (c) 2024 Primetals Technologies Austria GmbH
```
oberlehner: 2025

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
pazi146: external libs files get set read only on extraction

plugins/org.eclipse.fordiac.ide.library.ui/plugin.properties
pazi146: fixed

oberlehner (approved):

PR #1190, opened 2025-03-27
Check if Demultiplexer is configured for Forte Type Name Creation

m-meingast:


plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/interactors/ForteTypeNameCreator.java
```diff
@@ -53,6 +56,14 @@ private static String getConfigureFBType(final ConfigurableFB confFB) {
 		return typeName;
 	}
 
+	private static boolean isConfigured(final Demultiplexer demux) {
+		final EObject eObj = demux.eContainer().eContainer();
+		if (eObj instanceof SubAppType || eObj instanceof UntypedSubApp) {
```
azoitl: I would prefere if you could switch the if and check if the eObj is a Resource and only if it is contained in a resource get the opposite. Then the code would better capture the intent.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/interactors/ForteTypeNameCreator.java
m-meingast: Done

azoitl (approved):

PR #1194, opened 2025-03-31
Add property page for model validation severity

pazi146:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
```diff
@@ -30,6 +39,51 @@ public final class ModelPreferenceConstants {
...
+	private static int getDiagnosticSeverity(final String value, final int defaultValue) {
+		return switch (value) {
+		case SEVERITY_IGNORE -> Diagnostic.OK;
+		case SEVERITY_INFO -> Diagnostic.INFO;
+		case SEVERITY_WARNING -> Diagnostic.WARNING;
+		case SEVERITY_ERROR -> Diagnostic.ERROR;
+		case null, default -> defaultValue;
+		};
+	}
+
```
mx990: Why did you merge the validation preferences with the other model preferences?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
```diff
@@ -13,6 +13,15 @@
...
 package org.eclipse.fordiac.ide.model.preferences;
 
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.emf.common.util.Diagnostic;
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.fordiac.ide.ui.preferences.PreferenceStoreProvider;
```
mx990: Please do not use UI dependencies in the model plugin.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
pazi146: Because they need to use the same ID - so might as well be the same class
Accessing preferences for the first causes a lookup of the defaults which triggers the execution of the preference initializers.
The calls lead through `PreferenceServiceRegistryHelper.applyRuntimeDefaults()` which checks the store name (our qualifier) against the name of plugin that contains the initializer.
And as we the plugin isn't named `org.eclipse.fordiac.ide.model.validation` the default values won't get initialized at the correct time

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
pazi146: I didn't want to reimplement all the instance/project specific handling and the model plugin already uses ide.ui anyways

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
mx990: If we did use the same ID, it would mean we mix preferences from different pages and thus also could only select whether to use project settings for the entire group.

I would suggest to create a sub-node `validation` under `org.eclipse.fordiac.ide.model`, that is, changing the path from `org.eclipse.fordiac.ide.model.validation` to `org.eclipse.fordiac.ide.model/validation`. That should both allow us to use the preference initializer mechanism and also use a different node to store the preferences.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
mx990: The appropriate handling should already be implemented in `ValidationPreferences.getDiagnosticSeverity(String, int, IProject)`, except to check whether project properties are enabled. That should be solvable by a simple if to check for a boolean preference and only query the `InstanceScope` otherwise.

The model plugin unfortunately currently has to use some UI dependencies for coordinate conversion, which also makes additional UI dependencies easy to creep in unnoticed. However, that does not mean we should knowingly add more and thus only make it harder to get rid of those dependencies in the future.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
pazi146: I can do that for more clarity.

But you can absolutely split the behaviour (as is currently done in the `gef` plugin). Preferences are only saved if it is necessary - so if you activate project specific settings but make no changes the instance (or even default) settings would be used instead.
Also if you only access through the `Instance` scope you ignore any project specific ones - this already happens when you have project specific settings saved and then untick the check box -> the other settings don't get deleted, they simply no longer get accessed

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
pazi146: I don't want to use that handling only in the `ValidationPreferences`. What would be the appropriate package to move this to? Other plugins should also be able touse the instance/project settings switching without dependency on UI plugins
I'd also move the `PreferenceStoreProvider.PROJECT_STORE_ACTIVE` String there if possible

mx990: That makes sense. Ideally, I would say this should go into a separate plugin for generic, non-UI functionality. As we currently do not seem to have one, I believe this could go into `org.eclipse.fordiac.ide.model.preferences` for now instead.

@azoitl Maybe we could finally clean up the `org.eclipse.fordiac.ide.util` plugin and later move it there? We could then also move the `FordiacLogHelper` and the generic error marker handling there, as well?

pazi146: I'll move the functionality there but will need to duplicate the string (as `model` currently depends on `ide.ui`)

There is merit to a utility plugin with no 4diac dependencies so it can be used wherever needed

pazi146: done

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/preferences/ModelPreferenceConstants.java
azoitl: cleaning up util for such generic functions sounds like a very good idea.

azoitl (approved):

PR #1195, opened 2025-04-01
Add startTime & duration to OpSem model

Felix-Schmid:
Most of the changes happen because of regenerating the src-gen directory.
The new start time and durations are also given to the Evaluator.


bwiesmayr (changes requested):
the generated code is in some places edited manually, and those changes are now overridden. It seems that the generated NOT flag was not in all cases set correctly, sorry for this.

bwiesmayr (approved):

PR #1204, opened 2025-04-02
LibraryElement Hash Generation

azoitl:
Added infrastructure for getting hashes on a types content. Only type relevant information is used as basis for the hash. Comments, positions (e.g., CFB internals, connections), algorithm formating are ignored.

Also-by: Felix Roithmayr <felix.roithmayr@jku.at>

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
+++ b/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
@@ -735,6 +735,9 @@
         defaultValueLiteral=""/>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="ConnectionRoutingData">
+    <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+      <details key="ignored" value="true"/>
+    </eAnnotations>
     <eOperations name="is1SegementData" lowerBound="1" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EBoolean">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
         <details key="body" value="return ConnectionRoutingDataAnnotations.connectionRoutingDataValueIsNull(getDx1());"/>
@@ -937,6 +940,9 @@
       <eAnnotations source="http:///org/eclipse/emf/ecore/util/ExtendedMetaData">
         <details key="kind" value="attribute"/>
         <details key="name" value="Condition"/>
+      </eAnnotations>
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="transformer" value="org.eclipse.fordiac.ide.model.util.StringTransformer"/>
       </eAnnotations>
     </eStructuralFeatures>
     <eStructuralFeatures xsi:type="ecore:EReference" name="source" lowerBound="1"
@@ -1551,7 +1557,11 @@
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="name" lowerBound="1" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//String"
         defaultValueLiteral=""/>
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="comment" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//String"
-        defaultValueLiteral=""/>
+        defaultValueLiteral="">
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="ignored" value="true"/>
+      </eAnnotations>
+    </eStructuralFeatures>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="InputPrimitive" eSuperTypes="#//Primitive"/>
   <eClassifiers xsi:type="ecore:EClass" name="InterfaceList">
@@ -1842,6 +1852,9 @@
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="OutputPrimitive" eSuperTypes="#//Primitive"/>
   <eClassifiers xsi:type="ecore:EClass" name="Position">
+    <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+      <details key="ignored" value="true"/>
+    </eAnnotations>
     <eOperations name="toScreenPoint" lowerBound="1" eType="#//Point">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
         <details key="body" value="return PositionAnnotation.toScreenPoint(this);"/>
@@ -2008,7 +2021,11 @@
       </eAnnotations>
     </eStructuralFeatures>
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="comment" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//String"
-        defaultValueLiteral=""/>
+        defaultValueLiteral="">
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="ignored" value="true"/>
+      </eAnnotations>
+    </eStructuralFeatures>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="ServiceSequence" eSuperTypes="#//INamedElement #//ConfigurableObject">
     <eOperations name="getService" lowerBound="1" eType="#//Service">
@@ -2180,7 +2197,11 @@
     </eStructuralFeatures>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="TextAlgorithm" abstract="true" eSuperTypes="#//Algorithm">
-    <eStructuralFeatures xsi:type="ecore:EAttribute" name="text" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="text" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="transformer" value="org.eclipse.fordiac.ide.model.util.StringTransformer"/>
+      </eAnnotations>
+    </eStructuralFeatures>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="TextFunction" abstract="true" eSuperTypes="#//Function">
     <eStructuralFeatures xsi:type="ecore:EReference" name="inputParameters" upperBound="-1"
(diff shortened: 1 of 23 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
```diff
@@ -0,0 +1,23 @@
...
+ * Contributors:
+ *   Felix Roithmayr - initial API and implementation and/or initial documentation
+ ******************************************************************************/
+package org.eclipse.fordiac.ide.model.util;
+
+public class StringTransformer implements ITransformer {
+	@Override
+	public Object transform(final Object obj) {
+		if (obj instanceof final String str) {
+			return str.replaceAll("\s+", " "); //$NON-NLS-1$ //$NON-NLS-2$
```
ernstblechaPT: This will also reduce whitespace inside of strings/wstrings to a single space

ernstblechaPT (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
azoitl: @ernstblechaPT do you have an idea how to make it better. Should we also filter ST comments?

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -2180,7 +2194,11 @@
     </eStructuralFeatures>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="TextAlgorithm" abstract="true" eSuperTypes="#//Algorithm">
-    <eStructuralFeatures xsi:type="ecore:EAttribute" name="text" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="text" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="transformer" value="org.eclipse.fordiac.ide.model.util.StringTransformer"/>
```
mx990, line 2202: I believe the `text` attributes in `TextFunction`, `TextFunctionBody`, and `TextMethod` need a similar annotation? The same also for `conditionExpression` in `ECTransition`?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AbstractTypeEntryImpl.java
```diff
@@ -420,6 +424,22 @@ public Set<TypeEntry> getDependencies() {
 		return Collections.emptySet();
 	}
 
+	@Override
+	public String getTypeHash() {
+		final SoftReference<String> typeHashRefCached = typeHashRef;
+
+		if (typeHashRefCached != null) {
+			final String typeHash = typeHashRefCached.get();
+			if (typeHash != null) {
```
mx990, line 434: I believe this also needs to check if the type has changed on disk, just in case this is the first access after the change and before someone has called `getType()`, which would normally invalidate the hash.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
```diff
@@ -0,0 +1,134 @@
...
+		return FULL_HASH_STRING;
+	}
+
+	public static String getLibraryElementHash(final INamedElement libelem, final String version) {
+		final String[] versionParts = version.split(":"); //$NON-NLS-1$
+		if (versionParts.length != 2) {
+			return ""; // TODO raise exception maybe //$NON-NLS-1$
+		}
+		if (!versionParts[0].equals(HASH_VERSION)) {
+			return ""; // TODO raise exception maybe //$NON-NLS-1$
```
mx990: Please address the TODOs.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
```diff
@@ -0,0 +1,134 @@
...
+		return hash(libelem, versionParts[1]);
+	}
+
+	private static String hash(final EObject eobj, final String version) {
+		MessageDigest digest = null;
+		try {
+			digest = MessageDigest.getInstance(version);
+		} catch (final NoSuchAlgorithmException e) {
+			FordiacLogHelper.logError("could not aquire hashing algorithm", e); //$NON-NLS-1$
+			return ""; //$NON-NLS-1$
```
mx990: I believe this should throw an exception instead of failing silently.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		final ResourceSetImpl xmiResourceSet = new ResourceSetImpl();
+		Resource.Factory.Registry.INSTANCE.getExtensionToFactoryMap().putIfAbsent("xmi", new XMIResourceFactoryImpl()); //$NON-NLS-1$
+		final Resource xmiResource = xmiResourceSet.createResource(URI.createFileURI("tohash.xmi")); //$NON-NLS-1$
```
mx990: I would suggest extracting those strings into constants.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
```diff
@@ -0,0 +1,134 @@
...
+		xmiResource.getContents().add(copyLibraryElementForHashing(eobj));
+
+		final StringBuilder sb = new StringBuilder();
+
+		try (final ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
mx990: I would suggest using a `java.security.DigestOutputStream` (with a `null` underlying stream), in order to avoid having to cache everything in memory.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
mx990: It also produces a different result whether there is whitespace or not. For example, the string `17 + 4` will remain different from `17+4`, although they are semantically equivalent.

I believe we will need a rudimentary lexer for ST strings here, that can properly handle whitespace, strings, and ignore comments.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
ernstblechaPT: I do not have a better solution than what @mx990 proposes

But the example that creates a false positive (i.e. that `17 + 4` is different from `17+4`) seems less relevant to me - as `ADD(17, 4)` is also (somewhat) semantically equivalent.

ernstblechaPT: Please excuse the multiposting: An option using existing code might be to run the autoformatter on the ST-code to be hashed. That should also handle whitespace around operators. But I am not happy about the performance implications of that solution.

azoitl: This would then mark formatter changes as changed hashes. I was thinking about using the parser and serialize the AST model, which maybe has the comments removed. But again this would be interesting performance wise.

ernstblechaPT: Do I understand that right: you would just run the xmi-export on a filtered version of the AST model and generate the hash from that - just like it is done for the LibraryElements?

azoitl: Yes I would parse here the ST code, maybe filter, and return the root entry to the ast to be inserted into model. And while I'm writing this I noticed this will not work. So I would need to serialize at this point into a string and return this string.

azoitl: After thinking a bit I could write do a very simply hand written "lexer" that would just identify comments (single, multiline), STRING literals, and WSTRING literals and remove outside of these ALL whitespace. This would maybe not be parseable ST but it would lead to a consisten hash and the performance would be better. Also the maintainence would not be that hard as it would be at max 150 lines of code.

oberlehner: At which locations is the StringTransformer called ?
Is it not possible to check whether we pass a STRING/WSTRING before calling the `str.replaceAll("\s+", " ");`

azoitl: Currently for all Text algorithm like elements (i.e., Algorithms, Methods, Functions, Transitions). I'm working on a solution handling that all.

azoitl: I hope this new implementation is better fitting. Please check also the tests for cases that I may have missed. More concrete ST examples would be good.

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AbstractTypeEntryImpl.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
azoitl: I was not sure. But I think you are right. Done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		return result.toString();
+	}
+
+	private static int skipWhitespace(final String str, int i, final StringBuilder result) {
+		final boolean startsWithIdentifier = (i > 0 && isIdentifierChar(str.charAt(i - 1)));
+		while (i < str.length() && Character.isWhitespace(str.charAt(i))) {
+			i++;
+			if (i < str.length() && isIdentifierChar(str.charAt(i))) {
```
mx990: I believe you can move this `if` out of the `while` loop?

tests/org.eclipse.fordiac.ide.test.model/src/org/eclipse/fordiac/ide/test/model/util/StringTransformTest.java
```diff
@@ -0,0 +1,104 @@
...
+				Arguments.of("""
+						test unclosed multine comment
+						/*
+								Multi-line comment
+
+						""", "test unclosed multine comment"), //$NON-NLS-1$ //$NON-NLS-2$
+				Arguments.of("Test unclosed string \"Hello   World! ", "Test unclosed string\"Hello   World! "), //$NON-NLS-1$ //$NON-NLS-2$
+				Arguments.of("Test unclosed string 'Hello   World! ", "Test unclosed string'Hello   World! "), //$NON-NLS-1$ //$NON-NLS-2$
+				Arguments.of("\"Hello   World!  // Not a comment\";", "\"Hello   World!  // Not a comment\";"), //$NON-NLS-1$ //$NON-NLS-2$
+				Arguments.of("'Hello   World!  // Not a comment';", "'Hello   World!  // Not a comment';") //$NON-NLS-1$ //$NON-NLS-2$
```
mx990: I would suggest also adding a few tests where spaces around operators are removed as intended, for example:
```java
Arguments.of("  Test  :=  'Hello   World! $\' '  ;  ", "Test:='Hello   World! $' ';"), //$NON-NLS-1$ //$NON-NLS-2$
Arguments.of("testVar & varTest2", "testVar&varTest2"), //$NON-NLS-1$ //$NON-NLS-2$
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/StringTransformer.java
azoitl: ah yes, this is much cleaner.

tests/org.eclipse.fordiac.ide.test.model/src/org/eclipse/fordiac/ide/test/model/util/StringTransformTest.java
azoitl: added

mx990 (approved):

ernstblechaPT (approved):

PR #1218, opened 2025-04-08
Search algorithm jump to

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeXtextEditor.java
```diff
@@ -152,10 +157,13 @@ public void selectionChanged(final IWorkbenchPart part, final ISelection selecti
 
 	@Override
 	public boolean outlineSelectionChanged(final Object selectedElement) {
-		// do not react to selection changes from the FB type editor, until we can
-		// determine the text position for the selected element without:
-		// - blocking the UI thread waiting for a lock on the Xtext document,
-		// - interrupting the Xtext validation job when executed with priority.
```
mx990: Why did you remove this comment?

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeXtextEditor.java
```diff
@@ -152,10 +157,13 @@ public void selectionChanged(final IWorkbenchPart part, final ISelection selecti
...
+		if (selectedElement instanceof final Object[] objects
+				&& this.getInternalSourceViewer() instanceof XtextSourceViewer && objects.length > 0
+				&& objects[0] instanceof final TextMatch textMatch) {
+			final EObject eobject = getType().eResource().getEObject(textMatch.getUri().fragment());
+			selectAndReveal(textMatch.getOffset(), textMatch.getLength());
+			return eobject instanceof Algorithm || eobject instanceof Method;
```
mx990: This is a generic base class for all Xtext editor tabs in an FB type editor, including ST functions and even non-ST editors. As such, I believe it is unwise to specifically target `Algorithm`s and `Method`s here.

The original implementation did further check if the source actually contained the corresponding element, but that was causing problems, hence the comment you removed.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/editors/FBTypeXtextEditor.java
```diff
@@ -152,10 +157,13 @@ public void selectionChanged(final IWorkbenchPart part, final ISelection selecti
...
 	public boolean outlineSelectionChanged(final Object selectedElement) {
-		// do not react to selection changes from the FB type editor, until we can
-		// determine the text position for the selected element without:
-		// - blocking the UI thread waiting for a lock on the Xtext document,
-		// - interrupting the Xtext validation job when executed with priority.
+		if (selectedElement instanceof final Object[] objects
+				&& this.getInternalSourceViewer() instanceof XtextSourceViewer && objects.length > 0
+				&& objects[0] instanceof final TextMatch textMatch) {
+			final EObject eobject = getType().eResource().getEObject(textMatch.getUri().fragment());
+			selectAndReveal(textMatch.getOffset(), textMatch.getLength());
```
mx990: Why do you change the selection even though the match may not correspond to this editor, that is, the check in the next line may return false?

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchResult.java
```diff
@@ -32,6 +38,8 @@ public class ModelSearchResult extends AbstractTextSearchResult {
 
 	private final SearchNameDictionary dictionary;
 
+	private final Map<Object, Set<Match>> matches = new HashMap<>();
```
mx990:
```suggestion
	private final Map<URI, Set<Match>> matches = new HashMap<>();
```

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchResult.java
```diff
@@ -73,6 +81,25 @@ public void addResults(final List<? extends EObject> res) {
...
+		if (targetMatches == null) {
+			return new Match[0];
+		}
+		final Match[] sortingCopy = targetMatches.toArray(new Match[targetMatches.size()]);
+		Arrays.sort(sortingCopy, (match1, match2) -> match1.getLocation().compareTo(match2.getLocation()));
+		return sortingCopy;
+	}
+
+	public void addFordiacMatch(final Match match) {
+		matches.computeIfAbsent(match.getUri(), k -> ConcurrentHashMap.newKeySet()).add(match);
```
mx990: I would suggest using a `ConcurrentSkipListSet` with an appropriate comparator instead to always provide sorted matches. You can also use `Comparator.comparing(Match::getLocation)` to compare by the location.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchResultPage.java
```diff
@@ -237,11 +240,17 @@ private static void jumpToBlock(final OpenEvent doubleClick) {
 		}
 	}
 
-	private static void jumpHelper(final EObject jumpingTo) {
+	private void jumpHelper(final EObject jumpingTo) {
```
mx990, line 249: I would suggest using the `IGotoMarker` facility instead, which should offer a more genric way to open a search result than working with and integrating the search into different editors directly.

See also `org.eclipse.search.internal.ui.text.EditorOpener.showWithMarker(...)` for an example on how this is done by the regular text search.

oberlehner (approved):

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/ModelSearchResultPage.java
sebHollersbacher: I changed it, but I'm not sure how to correctly set the type of the marker.
Do you know how to best do that?

mx990: I have created a PR https://github.com/eclipse-4diac/4diac-ide/pull/1247 that adds the expected marker type to the search match.

sebHollersbacher:
Depends on #1247

mx990 (approved):

oberlehner (approved):

PR #1220, opened 2025-04-09
Made the ResizeGroupOrSubappCommand a QualNameAffected Command

oberlehner:


azoitl (changes requested):
I think there is no reason that the MoveAndREconnectCommand is inside of an compound command. Therefore I would recommend to also change the creation in SubAppContentLayoutEditPolicy. This would make the code there and here easier. Also I would introduce a getter for the MoveAndREconnectCommand so that we don't have duplicated code in all three methods.

azoitl (approved):

PR #1229, opened 2025-04-12
Handle Dependency Update during Editor Closing

azoitl:
When during closing several editors a dependency update was forwarded to the editor an NPE could occur.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
+++ b/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
@@ -78,7 +78,7 @@ public class TypeEntryAdapter extends AdapterImpl {
 			break;
 		case TypeEntry.TYPE_ENTRY_FILE_FEATURE:
 			Display.getDefault().asyncExec(() -> {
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -136,6 +136,10 @@ private void handleDependencyUpdate(final TypeEntry typeEntry) {
 		final LibraryElement editedElement = editor.getAdapter(LibraryElement.class);
 		if (editedElement != null) {
 			Display.getDefault().asyncExec(() -> {
+				if (editor.getSite().getPage() == null) {
```
mx990: We should probably add the same check in `handleFileContentChange()` above?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: good point. Added now to all display asyncs here. Just to be on the save side.

pazi146 (approved):

mx990 (approved):

PR #1230, opened 2025-04-13
Creating Types in Non-Existing Folders was Broken

azoitl:
When creating types (i.e., new types, extracting subapps and structs) an core exception was thrown because the workspace job was not set to the right parent directory.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/wizards/AbstractSaveAsWizard.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/wizards/AbstractSaveAsWizard.java
@@ -15,6 +15,7 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.wizards;
 
+import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.fordiac.ide.ui.providers.DialogSettingsProvider;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/wizards/AbstractSaveAsWizard.java
```diff
@@ -52,6 +53,16 @@ private void setupDiagramSettings() {
 		setDialogSettings(settings);
 	}
 
+	protected static IContainer getFirstExistingParent(final IFile targetFile) {
```
pazi146, line 56: why don't we just use `targetFile.getProject()`?
This will just go up the hierarchy and always return the workspace root

azoitl: We normally try to lock as little as possible as maybe several things happen in parallel that could block each other.

sebHollersbacher (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/wizards/AbstractSaveAsWizard.java
pazi146: I misread the while loop, specifically the `!parent.exists()`

pazi146 (approved):

PR #1231, opened 2025-04-14
create standard libraries with version 3.0

pazi146:
combined libraries devices, resources, segments into hardware

added packages:
> org::iec61131 -> iec61131-3
> org::iec61499 -> events, hardware, net
> org::eclipse::fordiac -> everthing else


mx990:
I would suggest to remove the `org` prefix for both `iec61131` and `iec61499`, since those are standard language features, just like `java.*` in Java or `std::` in C++.

I would further suggest to add the respective library name in the package (e.g., `iec61499::events` for the `events` library), in order to make it more obvious which types belong to which library.

@azoitl What do you think?

azoitl:
The idea with org prefix was coming from my but after a second thought I agree that IEC61131 and IEC61499 alone are better. I also think we should have sub packages for the individual libs.  Not sure though about IEC 61131 which is curently for us one big lib. 

For the Eclipse 4diac specific FBs I would use now `Eclipse4diac`.

mx990:
> Not sure though about IEC 61131 which is curently for us one big lib.

If we still want to have them in different sub-folders in forte, we would need to also put them into corresponding sub-packages or the include paths would not match.

> For the Eclipse 4diac specific FBs I would use now `Eclipse4diac`.

I realize we currently do not have a naming convention for packages, but I would personally prefer if packages were to always start with a lowercase letter (e.g., `iec61131`, `iec61499`, and `eclipse4diac`). I also believe this would look nicer in the resulting folder structure, include paths, and C++ namespaces.

azoitl:
> > Not sure though about IEC 61131 which is curently for us one big lib.
> 
> If we still want to have them in different sub-folders in forte, we would need to also put them into corresponding sub-packages or the include paths would not match.

I don't think we would like to move all IEC 61131-3 blocks into one big folder. So I guess we need to add packages for the sub folders.

mx990:
I would suggest to shorten the following package names:
- `iec61131::bistableelements` to `iec61131::bistable`,
- `iec61131::bitwiseoperators` to `iec61131::bitwise`,
- `iec61131::charstring` to `iec61131::string`,
- `iec61131::edgedetection` to `iec61131::edge`.

We could similarly shorten the corresponding folder names, such as `bistableElements` to `bistable`?

azoitl:
> I would suggest to shorten the following package names:
> 
>     * `iec61131::bistableelements` to `iec61131::bistable`,
> 
>     * `iec61131::bitwiseoperators` to `iec61131::bitwise`,
> 
>     * `iec61131::charstring` to `iec61131::string`,
> 
>     * `iec61131::edgedetection` to `iec61131::edge`.
> 
> 
> We could similarly shorten the corresponding folder names, such as `bistableElements` to `bistable`?

I don't care to much about the packages. But I don't want the folder names shown in the IDE be shortened. This looses meaning and these folder names follow the sections in IEC 61131-3.

oberlehner:
> > I would suggest to shorten the following package names:
> > ```
> > * `iec61131::bistableelements` to `iec61131::bistable`,
> > 
> > * `iec61131::bitwiseoperators` to `iec61131::bitwise`,
> > 
> > * `iec61131::charstring` to `iec61131::string`,
> > 
> > * `iec61131::edgedetection` to `iec61131::edge`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We could similarly shorten the corresponding folder names, such as `bistableElements` to `bistable`?
> 
> I don't care to much about the packages. But I don't want the folder names shown in the IDE be shortened. This looses meaning and these folder names follow the sections in IEC 61131-3.

i would go with alois in that discussion. i think that there is no benefit in shortening the package names in that case

oberlehner:
Aren't we violating our own conventions by putting the blocks into package namespaces that do not correspond with the folder name ?
I understand why the prefixes `iec61499`, `iec61499` and `eclipse4diac` are used, but i am not sure if this is a good idea or not when thinking about exchanging these packages with other tools.

in my opinion it would be better to add these `iec61499`, `iec61499` and `eclipse4diac`  somewhere else , maybe as attributes.
Furthermore, it does not correspond to the manifest file.
Maybe another option would be to simple add  `iec61499`, `iec61499` and `eclipse4diac` as manifest attribute.

mx990:
> > > We could similarly shorten the corresponding folder names, such as `bistableElements` to `bistable`?
> > 
> > I don't care to much about the packages. But I don't want the folder names shown in the IDE be shortened. This looses meaning and these folder names follow the sections in IEC 61131-3.
> 
> i would go with alois in that discussion. i think that there is no benefit in shortening the package names in that case

My main concern was to avoid package names such as `bistableelements`, since that is somewhat hard to read. Maybe we could then just use lower camel case for the packages (e.g., `iec61131::bistableElements`), which would also match the current folder names?

> Aren't we violating our own conventions by putting the blocks into package namespaces that do not correspond with the folder name ? I understand why the prefixes `iec61499`, `iec61499` and `eclipse4diac` are used, but i am not sure if this is a good idea or not when thinking about exchanging these packages with other tools.

I agree that we are currently not setting a particularly good example here. However, I would rather add an intermediate folder in each of the libraries than remove the package prefix.

The standard should probably ultimately define the `iec61499` and `iec61131` package hierarchies, which would also solve interoperability issues with other tools. As for the `eclipse4diac` prefix, I believe the name should make it clear that those types are vendor specific and will not be present or work with other tools.

> in my opinion it would be better to add these `iec61499`, `iec61499` and `eclipse4diac` somewhere else , maybe as attributes. Furthermore, it does not correspond to the manifest file. Maybe another option would be to simple add `iec61499`, `iec61499` and `eclipse4diac` as manifest attribute.

If we hide these prefixes in some attribute, we may as well not have them altogether. I believe the main purpose of having those prefixes is to immediately recognize standard and vendor-specific types and also to avoid name collisions if someone decides to name their package just `utils`. We should have probably included similar prefixes in the names of standard libraries for the same reasons, but that may be too late now.

mx990 (approved):

azoitl (approved):

PR #1237, opened 2025-04-15
Add ECState handling for SimpleFB

Felix-Schmid:
Interpreting SimpleFBs only returned the first output event so far.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
+++ b/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
@@ -28,6 +28,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.eclipse.emf.common.util.BasicEList;
+import org.eclipse.emf.common.util.ECollections;
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.emf.common.util.EMap;
 import org.eclipse.emf.ecore.EObject;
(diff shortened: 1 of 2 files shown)


bwiesmayr (commented):
I think this leads to some undesired behavior in the RecordSequenceDialog. I believe this should not be changed.

Felix-Schmid:
> I think this leads to some undesired behavior in the RecordSequenceDialog. I believe this should not be changed.

The changes in RecordSequenceDialog were done, because it otherwise leads to NullPointerExceptions when the type is something other than BasicFBType. The combobox was only created for BasicFBTypes, but still accessed in other cases.
For non-BasicFBTypes, the combobox is already disabled by existing code. However, I could change it so the combobox does not show up at all for e.g. SimpleFBTypes if this is preferred.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -316,12 +317,21 @@ public EList<EventOccurrence> runSimpleFBType(final SimpleFBTypeRuntime simpleFB
...
-		return outputEvents;
+
+		// if we don't have ECStates, we just return the first output event as fallback
+		if (simpleFBType.getSimpleECStates() == null || simpleFBType.getSimpleECStates().isEmpty()) {
+			final Event outputEvent = simpleFBType.getInterfaceList().getEventOutputs().get(0);
+			return ECollections.asEList(createOutputEventOccurrence(simpleFBTypeRuntime, outputEvent, simpleFBType));
+		}
+
+		// find ECStates matching the input event and return their outputs
+		final String inEvent = this.eventOccurrence.getEvent().getName();
```
azoitl: While this code nicely searches for the right outputEvents to be sent it misses to execute the right algorithms. I guess the better solution is to have a method getActions for getting the actions for the current event. And then for each action entry execute the algorithm if present and send the output event if present. This could be done similar to the action execution in the BasicFB. 

And if there are no simple actions the fall back would be the first algorithm and the first output event. Could even be returend by the get actions method in a freshly created action.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
Felix-Schmid: Thanks for the review! I fixed it as suggested.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -316,12 +319,31 @@ public EList<EventOccurrence> runSimpleFBType(final SimpleFBTypeRuntime simpleFB
 		// Initialization of variables
 		final SimpleFBType simpleFBType = simpleFBTypeRuntime.getSimpleFBType();
 		VariableUtils.fBVariableInitialization(simpleFBType);
-		final var outputEvents = new BasicEList<EventOccurrence>();
-		processAlgorithmWithEvaluator(simpleFBType, simpleFBType.getAlgorithm().get(0), eventOccurrence);
+
+		final List<SimpleECAction> actions = getActions(simpleFBType, eventOccurrence.getEvent().getName());
+
+		for (final SimpleECAction action : actions) {
+			simpleFBType.getAlgorithm().stream().filter(a -> a.getName().equals(action.getAlgorithm())).findAny()
```
azoitl, line 326: instead of the stream it should be possible that you just take the algorithm that is refered to by the action, if not null.

the event of the action could be put into a return list. This would reduce to have the stream at the bottom

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
Felix-Schmid: I addressed the second point.
For the first point, that is what I wanted to do initially. However, the `SimpleECAction` from `simpleFBType.getSimpleECStates()` doesn’t store the Algorithm but only the String of the algorithm name, which requires me to search for the actual algorithm based on the name.

azoitl: ah sorry I forgot that. Thx for the clarfication

azoitl (approved):

PR #1239, opened 2025-04-16
Add source suffix for negated connections in deployment

mx990:
Add `.NOT` source suffix for negated connections in deployment.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/util/DeploymentHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/util/DeploymentHelper.java
@@ -12,6 +12,7 @@
  *   Alois Zoitl - initial API and implementation and/or initial documentation
  *   Martin Jobst - adopt new ST editor for values
  *                - rework initial value handling
+ *                - add connection source suffix for delegate connections
  *******************************************************************************/
 package org.eclipse.fordiac.ide.deployment.util;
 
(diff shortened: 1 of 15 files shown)


plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/util/DeploymentHelper.java
```diff
@@ -84,6 +86,13 @@ private static boolean hasInitalValue(final VarDeclaration varDecl) {
 				&& !varDecl.getValue().getValue().isEmpty();
 	}
 
+	static String getSourceSuffix(final Connection connection) {
+		if (connection.isNegated()) {
```
azoitl: unfortunately this is not sufficient if source is a subapp interface we need to backtrack the connection and count the number of negations. Not sure if this can be more efficiently done during creating the connection data where we do the forward propagation.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/util/DeploymentHelper.java
mx990: You are right, I had completely forgotten about that. After thinking about this some more, we would also have to transform initial values on subapp interfaces, for example:
```text
         _____________________
        |       SubApp1       |
        |        ___________  |
        |       |    FB1    | |
TRUE - BOOL ---o| BOOL      | |
        |       |___________| |
        |_____________________|
```
The input of `SubApp1` has the default value `TRUE`. With the negated connection, this means the input of `FB1` must then receive a value of `FALSE`.

We would even have to consider unconnected subapp inputs without an initial value, for example:
```text
  _____________________
 |       SubApp1       |
 |        ___________  |
 |       |    FB1    | |
BOOL ---o| BOOL      | |
 |       |___________| |
 |_____________________|
```
The input of `SubApp1` has no default value, so it is `FALSE` by default. With the negated connection, this means the input of `FB1` must then receive a value of `TRUE`.

In order to properly handle all of that, I believe we should reverse the direction of resolving connections during deployment, that is, search backwards from FB inputs. We could then handle both connections and initial values at the same time. This should also simplify only deploying the outermost initial values for nested subapps and type conversions when going through different levels of subapp hierarchies. What do you think?

azoitl: Oh yes I have forgotten about the initial values. 

When I wrote the "new" deployment code for subapps I had a reason for tracing outputs and not inputs. I'm not perfectly sure what the reason was. But I think it was that we can handle fan outs about more efficiently. But given your arguments I also think tracing backwards is the better option now.

azoitl (approved):

Issue #1243, opened 2025-04-16
update typed subapp used in other typed subapp created multiple instances


moniwe:
when the user creates a typed subapp A and uses this within another typed subapp B, also creating connections from typed subapp A to the interface of typed subapp B, there seems to occure an NPE when the user saves changes within the typed subapp A. Maybe this NPE occures due to newly created instances of typed subapp A within typed subapp B laying over each other but now not having the connections.

[.log](https://github.com/user-attachments/files/19778421/default.log)

azoitl:
Thx for reporting. I could not reproduce the problem and after a longer investigation I think it can only happen when there is already an issue in the model. Nonetheless I did some clean-up in the problematic code. This not only makes it cleaner it should also better handle your situation.

PR #1244, opened 2025-04-16
Cleaned UpdateFBNElementCommand and protect against multi deletes #1243

azoitl:
The connection handling code in AbstractUpdateFBNElementCommand did a rather complicate check for self loops. Especially the creation code could result in mismatches to the deletion. With this clean-up it should not happen anymore.

Fixes https://github.com/eclipse-4diac/4diac-ide/issues/1243

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractReconnectConnectionCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractReconnectConnectionCommand.java
@@ -1,6 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2017 fortiss GmbH
- * 				 2019 Johannes Keppler University Linz
+ * Copyright (c) 2016, 2025 fortiss GmbH, Johannes Keppler University Linz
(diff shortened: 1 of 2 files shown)


bwiesmayr (approved):
I missed at first that there was a "distinct()" added to the stream, but now it makes sense.

azoitl:
> I missed at first that there was a "distinct()" added to the stream, but now it makes sense.

@bwiesmayr do you think I should add a comment to `distinct()`?

bwiesmayr:
> > I missed at first that there was a "distinct()" added to the stream, but now it makes sense.
> 
> @bwiesmayr do you think I should add a comment to `distinct()`?

at first I thought it wouldn't be necessary, but for safety it could be wise to annotate that we are returning a set of distinct connections without self-loops. I think that it is now "more correct" based on the method name than it was before, so a small comment should be sufficient for sure

azoitl:
> > > I missed at first that there was a "distinct()" added to the stream, but now it makes sense.
> > 
> > 
> > @bwiesmayr do you think I should add a comment to `distinct()`?
> 
> at first I thought it wouldn't be necessary, but for safety it could be wise to annotate that we are returning a set of distinct connections without self-loops. I think that it is now "more correct" based on the method name than it was before, so a small comment should be sufficient for sure

Done

oberlehner (approved):

PR #1245, opened 2025-04-17
Fixed BulkEditor search for Types to include package-name

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/SearchHelper.java
```diff
@@ -192,7 +192,7 @@ public boolean apply(final EObject searchCandidate) {
 				final ITypedElement typedElement = (ITypedElement) searchCandidate;
 
 				return matchesString(typedElement.getName(), filters.get(0), namePattern)
-						&& matchesString(typedElement.getTypeName(), filters.get(1), typePattern)
+						&& matchesString(typedElement.getFullTypeName(), filters.get(1), typePattern)
```
azoitl, line 195: have not tested but does this affect the full word search? Can I search for a type name without package?

oberlehner: we did not specifed how we deal with packages.
 i think we should discuss about handling packages in general, like searching for specific packages.

in my opinion it would make sense to provide an extra option somewhere to search for packages.
But i don't know how and where we should put that.

Issue #1248, opened 2025-04-19
custom type export issue - "Overwrite all" button does not work!


ashiwalv:
When I try to export a custom type FB (via 4diac IDE Type Export Wizard), if the same type already exists in my EXTERNAL_FB folder, then the "Overwrite all" button does not export the updated type. Just overwriting to the same export file works with the "Overwrite" button!

4diac IDE Version: 3.0.0
Build id: 2025-02-09_1716

PR #1249, opened 2025-04-21
Overwrite All didn't write the first file #1248

azoitl:
When selecting Overwrite All for exporting files the first file was not written. This is fixed now.

Fixes: https://github.com/eclipse-4diac/4diac-ide/issues/1248

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/TemplateExportFilter.java
+++ b/plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/TemplateExportFilter.java
@@ -164,6 +164,8 @@ public abstract class TemplateExportFilter extends ExportFilter {
 					openMergeEditor(writtenFiles);
 				}
 			} else if (res == BUTTON_OVERWRITE_ALL) {
+				// write the files that were prepared
+				files.write(true);
 				throw (new ExportException.OverwriteAll());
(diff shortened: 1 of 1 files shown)


ernstblechaPT (approved):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/TemplateExportFilter.java
```diff
@@ -164,6 +164,8 @@ public void export(final IFile typeFile, final String destination, final boolean
 					openMergeEditor(writtenFiles);
 				}
 			} else if (res == BUTTON_OVERWRITE_ALL) {
+				// write the files that were prepared
+				files.write(overwrite);
```
ernstblechaPT: in this case `overwrite` might not have the right value - in this place `true` should be passed to ensure that the file is always overwritten

ernstblechaPT (changes requested):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/TemplateExportFilter.java
azoitl: thx for noting. Fixed.

ernstblechaPT (approved):

pazi146 (approved):

PR #1253, opened 2025-04-22
Added list for VarConfig entries for TypedSubapps

PStmmr:


plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -2299,6 +2299,8 @@
     </eOperations>
     <eStructuralFeatures xsi:type="ecore:EReference" name="subAppNetwork" eType="#//FBNetwork"
         transient="true" derived="true" containment="true"/>
+    <eStructuralFeatures xsi:type="ecore:EReference" name="varConfigParams" upperBound="-1"
+        eType="#//VarDeclaration" transient="true" derived="true" containment="true"/>
```
mx990: I believe these should be neither transient nor derived?

mx990 (changes requested):

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: I totally agree they shouldn't be transient nor derived

azoitl (approved):

PR #1259, opened 2025-04-23
Automatic event connection creation

MandKastner:
Inserting a FB into a SubApp will automatically wire it into an existing execution chain if possible. By default the block is inserted at the end of the chain. If a predecessor FB has been marked via the context menu, the new block is inserted after it and will become the new predecessor.


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkPredecessorHandler.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.commands.IElementUpdater;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.ui.menus.UIElement;
+
+public class MarkPredecessorHandler extends AbstractHandler implements IElementUpdater {
+
+	private static final String MARKER_COLOR = "org.eclipse.fordiac.ide.ui.PredecessorMarkerColor"; //$NON-NLS-1$
+	private static Map<FBNetworkRootEditPart, AbstractFBNElementEditPart> predecessorMap = new HashMap<>();
```
oberlehner: i think you should use emf uris as key here.
Otherwise the reference is lost, if the editor refreshes. 
For example i have a certain block type as predesseor, then i make a non-breaking  change of that type  (removing an unused interface pin), this results in reloading the editor and the predessor is lost then.

oberlehner: Before:
image
After i removed the input pin DI9
image

oberlehner:
This case should also be considered:
Moving a marked block out of the subapp:
image
Now we have an invalid predessor here:
image


I think in that case the map should be cleared.

oberlehner:
The same applies for 
image

oberlehner:
I think you can have a look at the org.eclipse.fordiac.ide.model.commands.QualNameChangeListener extension point.
This might help you. because it will be triggered every time a selected element changes.

oberlehner (approved):

PR #1261, opened 2025-04-24
Added feature to make Connection use 5 segments

sebHollersbacher:


azoitl:
do we have any chance to get this feature without model change?
would make it easier to also have the manual 3 segment connection also we wouldn't persist this information.

sebHollersbacher:
@azoitl 
I don't think manual 3 segment connection would add anything. It should just be 1 <-> 5 and 3 <-> 5
It could be done without a model change, but moving the block would then reset this setting.
Another option would be an Internal attribute, which would then be persistable.

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ToggleConnectionBendpointCommand.java
```diff
@@ -0,0 +1,51 @@
...
+	private final ConnectionRoutingData connectionRoutingData;
+
+	public ToggleConnectionBendpointCommand(final Connection connection, final org.eclipse.draw2d.Connection figure) {
+		this.connectionRoutingData = connection.getRoutingData();
+		this.figure = figure;
+	}
+
+	@Override
+	public void execute() {
+		toggle5Segments();
```
azoitl: As you are not storing the original connection routing data undo is not working. Also I would split it into two commands:
  - ConvertToThreeSegmentsCon
  - ConvertToFiveSegmentsCon
  
  Most probably with a common base class with the common functionality. 
  The advantage is that you could in both commands create a new connectionrouting data suitable for the different kind of connection.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ToggleConnectionBendpointCommand.java
```diff
@@ -0,0 +1,71 @@
...
+	protected void createInitalNewRoutingData() {
+		newRoutingData = LibraryElementFactory.eINSTANCE.createConnectionRoutingData();
+		newRoutingData.setDx1(oldRoutingData.getDx1());
+		newRoutingData.setDx2(oldRoutingData.getDx2());
+		newRoutingData.setDy(oldRoutingData.getDy());
+	}
+
+	protected void changeRoutingData(final ConnectionRoutingData routingData) {
+		this.connection.setRoutingData(routingData);
+		this.figure.revalidate();
```
azoitl: this should not be needed, at least in my tests I got the same result with and witout it. not haveing ui elements in commands is normally better.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ToggleConnectionBendpointCommand.java
sebHollersbacher: Thank you for noticing, this makes it much cleaner

azoitl (approved):

PR #1268, opened 2025-04-29
Append inserted FBs to execution chain

MandKastner:
A FB inserted into a untyped subapp will be automatically inserted (and connected) at the end of a given execution chain.


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/InsertFBIntoExecutionChainCommand.java
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	@Override
+	public void undo() {
+		commands.undo();
+	}
+
+	@Override
+	public void redo() {
+		execute();
```
oberlehner: This will create duplicate connections, since execute adds new commands to the compound command on execution and the old ones are not cleaned

MandKastner: thank you i missed that

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/DeleteElementInExecutionChainCommand.java
```diff
@@ -0,0 +1,87 @@
+package org.eclipse.fordiac.ide.application.commands;
```
oberlehner: the copyright header is missing here. that is the reason why the build fails

PR #1273, opened 2025-05-02
Add CBasicFB::setInitialValues call in Basic FB 4diac FORTE Exporter

azoitl:
In order that the ECC State is also correctly reset a call to CBasicFB::setInitialValues shall be done for all basic fbs.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteFBTemplate.xtend
@@ -388,27 +388,20 @@ abstract class ForteFBTemplate<T extends FBType> extends ForteLibraryElementTemp
 	'''
 
 	def protected generateSetInitialValuesDeclaration(Iterable<VarDeclaration> variables) '''
-		«IF containsNonRetainedVariable(variables)»
-			void setInitialValues() override;
-		«ENDIF»
+		void setInitialValues() override;
 	'''
 
 	def protected generateSetInitialValuesDefinition(Iterable<VarDeclaration> variables) '''
-		«IF (containsNonRetainedVariable(variables))»
-			void «className»::setInitialValues() {
-			  «generateVariableDefaultAssignment(variables.filter[!isRetainedVariable(it)])»
-			}
-			
-		«ENDIF»	
+		void «className»::setInitialValues() {
+		  «baseClass»::setInitialValues();
+		  «generateVariableDefaultAssignment(variables.filter[!isRetainedVariable(it)])»
+		}
+		
 	'''
 
 	def private boolean isRetainedVariable(VarDeclaration variable) {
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/basic/BasicFBImplTemplate.xtend
```diff
@@ -44,6 +45,16 @@ class BasicFBImplTemplate extends BaseFBImplTemplate<BasicFBType> {
 			ILanguageSupportFactory.createLanguageSupport("forte_ng", it, options)
 		]
 	}
+	
+	override generateSetInitialValuesDefinition(Iterable<VarDeclaration> variables) '''
+		«IF (containsNonRetainedVariable(variables))»
```
mx990: We now always need a `setInitialValues()`, since it is pure virtual in `CBasicFB`. The same goes for the method this one is overriding, since it is also pure virtual in `CFunctionBlock`.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/basic/BasicFBImplTemplate.xtend
```diff
@@ -44,6 +45,16 @@ class BasicFBImplTemplate extends BaseFBImplTemplate<BasicFBType> {
...
+			void «className»::setInitialValues() {
+			  CBasicFB::setInitialValues();
```
mx990: If we were to provide an empty pure virtual default implementation in `CFunctionBlock`, we could just always call `baseClass::setInitialValues();` in the generic case. That would also allow us to extend the behavior in `CFunctionBlock` in the future if need be.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/basic/BasicFBImplTemplate.xtend
azoitl: I like that a lot. it allows us to also add later functionality into this empty methods.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/basic/BasicFBImplTemplate.xtend
azoitl: done

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/basic/BasicFBImplTemplate.xtend
azoitl: done

mx990 (approved):

PR #1277, opened 2025-05-05
adding eIO FBs to the typelibrary

scharf-m:
This pull request introduces four new service interface function blocks and one adapter. 
The `eIX` and `eIW` function blocks are similar to the standard `IX` and `IW` FBs but include an additional adapter port. Through this adapter (`eGenAdapter`), it becomes possible to configure external triggers for these function blocks using the `eIXconfig` and `eIWconfig` FBs.

### DIFF
--- a/data/typelibrary/io-1.0.0/typelib/eIO/eIWconfig.fbt
+++ b/data/typelibrary/io-1.0.0/typelib/eIO/eIWconfig.fbt
@@ -1,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FBType Name="eIWconfig" Comment="Service Interface Function Block Type">
+	<Identification Standard="61499-2">
+	</Identification>
+	<VersionInfo Version="1.0" Author="Maximilian Scharf" Date="2025-05-05">
+	</VersionInfo>
+	<CompilerInfo>
+	</CompilerInfo>
+	<InterfaceList>
+		<EventInputs>
+			<Event Name="CONF" Type="Event" Comment="Configure Request">
+				<With Var="QI"/>
+				<With Var="ST"/>
+				<With Var="BT"/>
+				<With Var="GRAD"/>
+			</Event>
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="CNF" Type="Event" Comment="Confirmation of Requested Service">
+				<With Var="QO"/>
+				<With Var="STATUS"/>
+			</Event>
+		</EventOutputs>
+		<InputVars>
+			<VarDeclaration Name="QI" Type="BOOL" Comment="Event Input Qualifier"/>
+			<VarDeclaration Name="ST" Type="WORD" Comment="Smaller than"/>
+			<VarDeclaration Name="BT" Type="WORD" Comment="Bigger than"/>
+			<VarDeclaration Name="GRAD" Type="WORD" Comment="Gradient in %"/>
+		</InputVars>
+		<OutputVars>
+			<VarDeclaration Name="QO" Type="BOOL" Comment="Event Output Qualifier"/>
+			<VarDeclaration Name="STATUS" Type="WSTRING" Comment="Service Status"/>
+		</OutputVars>
+		<Sockets>
+			<AdapterDeclaration Name="eIW" Type="eGenAdapter"/>
+		</Sockets>
+	</InterfaceList>
+	<Service RightInterface="RESOURCE" LeftInterface="APPLICATION">
+		<ServiceSequence Name="normal_establishment">
+			<ServiceTransaction>
+				<InputPrimitive Interface="APPLICATION" Event="INIT+" Parameters="PARAMS"/>
+				<OutputPrimitive Interface="RESOURCE" Event="initialize" Parameters="PARAMS"/>
+				<OutputPrimitive Interface="APPLICATION" Event="INITO+"/>
+			</ServiceTransaction>
+		</ServiceSequence>
+		<ServiceSequence Name="unsuccessful_establishment">
+			<ServiceTransaction>
+				<InputPrimitive Interface="APPLICATION" Event="INIT+" Parameters="PARAMS"/>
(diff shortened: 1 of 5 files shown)


franz-ms-muc:
> `eIXconfig`

BTW: 
you need also an `eIE` Block and a `eIEconfig` e.g. if you want "Raising Edge" this is only a Event so a IE Block.

data/typelibrary/io-1.0.0/typelib/eIO/eIXconfig.fbt
franz-ms-muc, line 1: here it would be easier: 
make a Input of a ENUM, and define in the Enum the cases, like Double Click, Falling Edge, Raising Edge, long Press etc.. ! makes cleaner Block and easier extension without ending up in say 10..15 BOOL inputs which are then needed to config.

data/typelibrary/io-1.0.0/typelib/eIO/eIXconfig.fbt
```diff
@@ -0,0 +1,106 @@
...
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="CNF" Type="Event" Comment="Confirmation of Requested Service">
+				<With Var="QO"/>
+				<With Var="STATUS"/>
+			</Event>
+		</EventOutputs>
+		<InputVars>
+			<VarDeclaration Name="QI" Type="BOOL" Comment="Event Input Qualifier"/>
+			<VarDeclaration Name="FE" Type="BOOL" Comment="Falling Edge Trigger Enable"/>
```
franz-ms-muc, line 22: here it would be easier:
make a Input of a ENUM, and define in the Enum the cases, like Double Click, Falling Edge, Raising Edge, long Press etc.. ! makes cleaner Block and easier extension without ending up in say 10..15 BOOL inputs which are then needed to config.

data/typelibrary/io-1.0.0/typelib/eIO/eIWconfig.fbt
```diff
@@ -0,0 +1,108 @@
...
+			<Event Name="CNF" Type="Event" Comment="Confirmation of Requested Service">
+				<With Var="QO"/>
+				<With Var="STATUS"/>
+			</Event>
+		</EventOutputs>
+		<InputVars>
+			<VarDeclaration Name="QI" Type="BOOL" Comment="Event Input Qualifier"/>
+			<VarDeclaration Name="ST" Type="WORD" Comment="Smaller than"/>
+			<VarDeclaration Name="BT" Type="WORD" Comment="Bigger than"/>
+			<VarDeclaration Name="GRAD" Type="WORD" Comment="Gradient in %"/>
```
franz-ms-muc, line 25: wouldn't a Hysteresis here be better ? 
`IW` is a general Block, can be used for Analog Input. 

but `IW` does not tell one if it is a 10bit, 14bit or 24bit Analog input 
(OK, 24 bit would require a ID)
so % is a bit complicated. 


I would recommend either a Hysteresis in Bits (same as Smaller, Bigger etc..) or if we want to stick in % at least 0.1% resolution better 0.01% Resolution. % is sure to raw. 

also we need to think about Error Handling. 

MATCH from HYDAC might be a good Template for it. 

Think of a Sensor: https://elo.store/Winkelsensor-0700-CCW-424A17A070-0-5-4-5V/424A17A070

it has 0,5-4,5V

i want Values from 0,5-4,5V, but i want an Error Flag "SHORT CIRCUIT" if 5V, and a "OPEN CIRCUIT" if 0V. 
how could this be done ?

data/typelibrary/io-1.0.0/typelib/eIO/eIWconfig.fbt
```diff
@@ -0,0 +1,108 @@
...
+	<Identification Standard="61499-2">
+	</Identification>
+	<VersionInfo Version="1.0" Author="Maximilian Scharf" Date="2025-05-05">
+	</VersionInfo>
+	<CompilerInfo>
+	</CompilerInfo>
+	<InterfaceList>
+		<EventInputs>
+			<Event Name="CONF" Type="Event" Comment="Configure Request">
+				<With Var="QI"/>
```
azoitl: I don't think that in your config blocks QI and QO is needed. Please remove them.

data/typelibrary/io-1.0.0/typelib/eIO/eIWconfig.fbt
azoitl: Hysterisis is something else and may be an own event config block. Here we wanted to experiment with rate change as event generation criteria.

The other elements that you mention should IMHO not be here but at IO config side as they are sensor specific. I don't think sensor specific information should be at application side. I also know that we are not there yet.

data/typelibrary/io-1.0.0/typelib/eIO/eIXconfig.fbt
azoitl: as we are currently not supporting eNUMS in FORTE this is hard to do.

azoitl (changes requested):
Please remove the service sequences from the template. As they are not suitable for your blocks.

Please add copyright headers to your blocks. You can take the other event blocks as reference.

azoitl (approved):

franz-hoepfinger-4diac:
@azoitl @scharf-m 

the eIO was not ported to 4diac-ide\typelibrary\io-3.0.0

is this by intention ?

see: 

https://github.com/eclipse-4diac/4diac-ide/tree/develop/data/typelibrary/io-3.0.0/typelib
https://github.com/eclipse-4diac/4diac-ide/tree/develop/data/typelibrary/io-1.0.0/typelib

Thanks, 
Franz

franz-hoepfinger-4diac:
Question: 

how does this work: 

image

with an "Empty" Adapter ? 


image

this is very interesting.
is it a new Concept ?

azoitl:
@franz-hoepfinger-4diac here both sides are SIFBs. SIFBs can utilize implementation specific features. For performance reasons the design makes use of this and uses the adapter connection mainly to find the implementation of the event generation.

PR #1289, opened 2025-05-06
FIX: Var config typed sub app sheet

PStmmr:
FB Network Editor


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -168,10 +172,8 @@ private List<VarDeclaration> collectVarConfigs() {
 		if (getType() instanceof final TypedSubApp tsa) {
 			return tsa.getVarConfigParams();
 		}
-		final Stream<VarDeclaration> varConfStream = StreamSupport
-				.stream(Spliterators.spliteratorUnknownSize(EcoreUtil.getAllProperContents(getType(), true), 0), false)
-				.filter(VarDeclaration.class::isInstance).map(VarDeclaration.class::cast)
```
oberlehner, line 173: i think it would be sufficient if you just filter the varConfig children of typed subapps here.
something like  `!FBNetworkElementHelper.isContainedInTypedInstance(var)`

The `FBNetworkElementHelper.isContainedInTypedInstance()` can be also used for any eObject now.
that is changed in https://github.com/eclipse-4diac/4diac-ide/pull/1284/ .

oberlehner (approved):

PR #1293, opened 2025-05-07
Use default scopes with preferences

pazi146:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
```diff
@@ -269,13 +272,15 @@ protected IFigure createFigure() {
 
 	private void performConnTypeConfiguration(final FBNetworkConnection connectionFigure) {
 		if (getModel() instanceof EventConnection) {
-			connectionFigure.setVisible(!InstanceScope.INSTANCE.getNode(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID)
-					.getBoolean(UIPreferenceConstants.P_HIDE_EVENT_CON, false));
+			connectionFigure.setVisible(!Platform.getPreferencesService().getBoolean(
+					UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID, UIPreferenceConstants.P_HIDE_EVENT_CON, false,
+					new IScopeContext[] { InstanceScope.INSTANCE, DefaultScope.INSTANCE }));
```
mx990: It should not be necessary to explicitly specify the scopes each time. Simply passing `null` should suffice, see also the documentation for `IPreferencesService.getBoolean(...)`

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/ConnectionEditPart.java
pazi146: thanks for pointing that out - changed it to `null`

mx990 (approved):

PR #1298, opened 2025-05-07
Refactoring: Moved isContainedInTypedInstance() to super class

oberlehner:
We might want to check all elements which can be part of an FBnetwork if they are inside a TypedSubappInstance

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
+++ b/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
@@ -1260,11 +1260,6 @@
         <details key="body" value="&#x9;&#x9;if ((this.eContainer() == null) || !(this.eContainer().eContainer() instanceof FBNetworkElement)) {&#xA;&#x9;&#x9;&#x9;return null;&#xA;&#x9;&#x9;}&#xA;&#x9;&#x9;return (FBNetworkElement) this.eContainer().eContainer();"/>
       </eAnnotations>
     </eOperations>
-    <eOperations name="isContainedInTypedInstance" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean">
-      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
-        <details key="body" value="return org.eclipse.fordiac.ide.model.helpers.FBNetworkElementHelper.isContainedInTypedInstance(this);"/>
-      </eAnnotations>
-    </eOperations>
     <eOperations name="isInGroup" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean">
       <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
         <details key="body" value="return org.eclipse.fordiac.ide.model.annotations.FBNetworkElementAnnotations.isInGroup(this);"/>
@@ -1736,6 +1731,11 @@
           <eTypeArguments eClassifier="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EJavaObject"/>
         </eGenericType>
       </eParameters>
+    </eOperations>
+    <eOperations name="isContainedInTypedInstance" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="return org.eclipse.fordiac.ide.model.helpers.FBNetworkElementHelper.isContainedInTypedInstance(this);"/>
+      </eAnnotations>
     </eOperations>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="IVarElement" abstract="true" interface="true">
--- a/plugins/org.eclipse.fordiac.ide.model/model/fordiac.genmodel
+++ b/plugins/org.eclipse.fordiac.ide.model/model/fordiac.genmodel
@@ -478,8 +478,6 @@
           body="final org.eclipse.emf.ecore.EObject parent = this.eContainer();&#xA;if (parent == null) {&#xA;&#x9;return false;&#xA;}&#xA;final org.eclipse.emf.ecore.EObject grandParent = parent.eContainer();&#xA;if (grandParent == null) {&#xA;&#x9;return false;&#xA;}&#xA;return grandParent instanceof org.eclipse.fordiac.ide.model.libraryElement.SubApp;"/>
       <genOperations ecoreOperation="lib.ecore#//FBNetworkElement/getOuterFBNetworkElement"
           body="&#x9;&#x9;if ((this.eContainer() == null) || !(this.eContainer().eContainer() instanceof FBNetworkElement)) {&#xA;&#x9;&#x9;&#x9;return null;&#xA;&#x9;&#x9;}&#xA;&#x9;&#x9;return (FBNetworkElement) this.eContainer().eContainer();"/>
-      <genOperations ecoreOperation="lib.ecore#//FBNetworkElement/isContainedInTypedInstance"
-          body="return org.eclipse.fordiac.ide.model.helpers.FBNetworkElementHelper.isContainedInTypedInstance(this);"/>
       <genOperations ecoreOperation="lib.ecore#//FBNetworkElement/isInGroup" body="return org.eclipse.fordiac.ide.model.annotations.FBNetworkElementAnnotations.isInGroup(this);"/>
       <genOperations ecoreOperation="lib.ecore#//FBNetworkElement/validateName" body="return org.eclipse.fordiac.ide.model.libraryElement.impl.FBNetworkElementAnnotations.validateName(this, diagnostics, context);">
         <genParameters ecoreParameter="lib.ecore#//FBNetworkElement/validateName/diagnostics"/>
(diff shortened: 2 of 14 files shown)


plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -1737,6 +1732,11 @@
         </eGenericType>
       </eParameters>
     </eOperations>
+    <eOperations name="isContainedInTypedInstance" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="return org.eclipse.fordiac.ide.model.helpers.FBNetworkElementHelper.isContainedInTypedInstance(this);"/>
```
mx990, line 1737: Please then also move the implementation to `org.eclipse.fordiac.ide.model.libraryElement.impl.TypedElementAnnotations`.

mx990 (changes requested):
If we want to bring this to `ITypedElement`, I believe we also have to consider additional possibilities for being *contained in a typed instance*, such as variables in an FB instance.

Maybe we could consider any container that is an `ITypedElement` where `getType() != null` a typed instance? Would that still fit with the current and intended use cases?

oberlehner:
> If we want to bring this to `ITypedElement`, I believe we also have to consider additional possibilities for being _contained in a typed instance_, such as variables in an FB instance.
> 
> Maybe we could consider any container that is an `ITypedElement` where `getType() != null` a typed instance? Would that still fit with the current and intended use cases?

Can you give me a more concrete example on that, i am not sure if i understand that correclty ?
Do you mean internal blocks for example or did i misunderstand that ?

The intention for that is to easily find elements which are contained in instances of typed subapps or CFBs as can be seen in the code of the helper:

```
	public static boolean isContainedInTypedInstance(final EObject element) {
		EObject obj = element;
		while (obj.eContainer() != null) {
			obj = obj.eContainer();
			if ((obj instanceof final SubApp subApp && subApp.isTyped()) || (obj instanceof CFBInstance)) {
				return true;
			}
		}
		return false;
	}
```

 For example, blocks  inside TypedSubapp Instance viewers.
 Currently this was only possible in the fbNetworkElements. Other classes needed to call the helper method.
 And now i want to extend that, to also detect attributes, varConfigs, InterfaceElements, etc.

mx990:
> > If we want to bring this to `ITypedElement`, I believe we also have to consider additional possibilities for being _contained in a typed instance_, such as variables in an FB instance.
> > Maybe we could consider any container that is an `ITypedElement` where `getType() != null` a typed instance? Would that still fit with the current and intended use cases?
> 
> Can you give me a more concrete example on that, i am not sure if i understand that correclty ? Do you mean internal blocks for example or did i misunderstand that ?

I basically meant any `IInterfaceElement` in an `FB`, whether it is an internal FB or contained in a network. I am not sure if that fits with your use case, of course.

> For example, blocks inside TypedSubapp Instance viewers. Currently this was only possible in the fbNetworkElements. Other classes needed to call the helper method. And now i want to extend that, to also detect attributes, varConfigs, InterfaceElements, etc.

If I understand the current usage correctly, it is mainly used to check for elements that are *derived* from somewhere else (and should thus not be modified by themselves). Maybe we could generalize this into a method `isDerived()`, which would include:
- interface and network of CFB or typed subapp instances,
- interface of FB instances (both internal FBs and in networks),
- FB network elements mapped from an application network.

What do you think? Would that fit with your extended use case?

oberlehner:
> > > If we want to bring this to `ITypedElement`, I believe we also have to consider additional possibilities for being _contained in a typed instance_, such as variables in an FB instance.
> > > Maybe we could consider any container that is an `ITypedElement` where `getType() != null` a typed instance? Would that still fit with the current and intended use cases?
> > 
> > 
> > Can you give me a more concrete example on that, i am not sure if i understand that correclty ? Do you mean internal blocks for example or did i misunderstand that ?
> 
> I basically meant any `IInterfaceElement` in an `FB`, whether it is an internal FB or contained in a network. I am not sure if that fits with your use case, of course.
> 
> > For example, blocks inside TypedSubapp Instance viewers. Currently this was only possible in the fbNetworkElements. Other classes needed to call the helper method. And now i want to extend that, to also detect attributes, varConfigs, InterfaceElements, etc.
> 
> If I understand the current usage correctly, it is mainly used to check for elements that are _derived_ from somewhere else (and should thus not be modified by themselves). Maybe we could generalize this into a method `isDerived()`, which would include:
> 
> * interface and network of CFB or typed subapp instances,
> * interface of FB instances (both internal FBs and in networks),
> * FB network elements mapped from an application network.
> 
> What do you think? Would that fit with your extended use case?

I think you are describing a different use case - which also sounds interesting. 
Would make sense to ensure, that derived elements can not be edited.

But it would not fit my use case.
For example i want to address only interface pins, fbNetworkElements,etc,.. which are inside the fbNetwork of TypedSubapps. 

**Concrete Use Case:** an interface pin  of an FBNetworkElement inside a typedSubapp should not be allowed to have an attribute or pin instance comment.
In contrast the interface pin of the typed subapp itself would be capable of doing that.

If I understood your use case correctly, this would result in `TypedSubapp.Pin1.isDerived() == TRUE ` 
 and `TypedSubapp.InnerFbNetworkElement.Pin2.isDervied() == TRUE`

Whereas my code would, then return 
`TypedSubapp.Pin1.isContainedInTypedInstance()== FALSE` 
 and `TypedSubapp.InnerFbNetworkElement.Pin2.isContainedInTypedInstance() == TRUE`

mx990:
> **Concrete Use Case:** an interface pin of an FBNetworkElement inside a typedSubapp should not be allowed to have an attribute or pin instance comment. In contrast the interface pin of the typed subapp itself would be capable of doing that.
> 
> If I understood your use case correctly, this would result in `TypedSubapp.Pin1.isDerived() == TRUE ` and `TypedSubapp.InnerFbNetworkElement.Pin2.isDervied() == TRUE`
> 
> Whereas my code would, then return `TypedSubapp.Pin1.isContainedInTypedInstance()== FALSE` and `TypedSubapp.InnerFbNetworkElement.Pin2.isContainedInTypedInstance() == TRUE`

If I understand the current implementation of `isContainedInTypedInstance` correctly, `TypedSubapp.Pin1.isContainedInTypedInstance()` would return `true`, because the pin is, in fact, contained in a typed subapp instance.

Aside from that, the attributes of interface elements in `FunctionFBTypes` should also not be editable. Maybe it would be better to add a specific method `isAttributesEditable()` in `ConfigurableObject` to handle all of these cases?

oberlehner:
no `TypedSubapp.Pin1.isContainedInTypedInstance` would return false, because it should only return true to elements which are inside the network of typedSubapp. 
Sry i described that bad. maybe the naming of the method is misleading. Maybe it should be `isContaintainedInReadOnlyNetwork() `or something similar.

mx990:
> no `TypedSubapp.Pin1.isContainedInTypedInstance` would return false, because it should only return true to elements which are inside the network of typedSubapp.

That may be the intent, but it does not seem to be the current behavior.

Feel free to try out the following snippet for yourself:
```java
final TypedSubApp subApp = LibraryElementFactory.eINSTANCE.createTypedSubApp();
subApp.setName("TypedSubapp");

final InterfaceList interfaceList = LibraryElementFactory.eINSTANCE.createInterfaceList();
subApp.setInterface(interfaceList);

final VarDeclaration varDeclaration = LibraryElementFactory.eINSTANCE.createVarDeclaration();
varDeclaration.setName("Pin1");
interfaceList.getInputVars().add(varDeclaration);

System.out.println("TypedSubapp.Pin1.isContainedInTypedInstance(): " + varDeclaration.isContainedInTypedInstance());
```

I get the following output based on your PR:
```
TypedSubapp.Pin1.isContainedInTypedInstance(): true
```

It does not seem `isContainedInTypedInstance()` actually checks anywhere that the element is contained in a *network* before it is contained in a typed instance.

> Sry i described that bad. maybe the naming of the method is misleading. Maybe it should be `isContaintainedInReadOnlyNetwork() `or something similar.

I would then suggest `isContainedInTypedInstanceNetwork()`.

oberlehner:
> > no `TypedSubapp.Pin1.isContainedInTypedInstance` would return false, because it should only return true to elements which are inside the network of typedSubapp.
> 
> That may be the intent, but it does not seem to be the current behavior.
> 
> Feel free to try out the following snippet for yourself:
> 
> ```java
> final TypedSubApp subApp = LibraryElementFactory.eINSTANCE.createTypedSubApp();
> subApp.setName("TypedSubapp");
> 
> final InterfaceList interfaceList = LibraryElementFactory.eINSTANCE.createInterfaceList();
> subApp.setInterface(interfaceList);
> 
> final VarDeclaration varDeclaration = LibraryElementFactory.eINSTANCE.createVarDeclaration();
> varDeclaration.setName("Pin1");
> interfaceList.getInputVars().add(varDeclaration);
> 
> System.out.println("TypedSubapp.Pin1.isContainedInTypedInstance(): " + varDeclaration.isContainedInTypedInstance());
> ```
> 
> I get the following output based on your PR:
> 
> ```
> TypedSubapp.Pin1.isContainedInTypedInstance(): true
> ```
> 
> It does not seem `isContainedInTypedInstance()` actually checks anywhere that the element is contained in a _network_ before it is contained in a typed instance.
> 
> > Sry i described that bad. maybe the naming of the method is misleading. Maybe it should be `isContaintainedInReadOnlyNetwork() `or something similar.
> 
> I would then suggest `isContainedInTypedInstanceNetwork()`.

Thank you for noticing. You are totally right. The code is not correct. It will most probably work indeed for most cases. Thats why no one noticed it yet.
I will fix that.

PR #1302, opened 2025-05-12
Application EditPart VarConfig XML Export

PStmmr:
Feature adds adds the function by which elements from the subapps are saved to the xml when updates are made in the application edit part.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
@@ -21,9 +21,12 @@ import java.util.List;
 
 import javax.xml.stream.XMLStreamException;
 
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.model.LibraryElementTags;
 import org.eclipse.fordiac.ide.model.datatype.helper.IecTypes;
 import org.eclipse.fordiac.ide.model.libraryElement.AdapterType;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
```diff
@@ -86,6 +88,10 @@ protected void addFBNetworkElement(final FBNetworkElement fbnElement) throws XML
 			addStartElement(nodeName);
 			addFBNetworkElementXMLAttributes(fbnElement);
 			addFBNetworkElementChildren(fbnElement);
+			if (fbnElement.eContainer().eContainer() instanceof Application
```
oberlehner: i think we dont need this check. it should also be stored in SubappTypes

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
```diff
@@ -86,6 +89,12 @@ protected void addFBNetworkElement(final FBNetworkElement fbnElement) throws XML
 			addStartElement(nodeName);
 			addFBNetworkElementXMLAttributes(fbnElement);
 			addFBNetworkElementChildren(fbnElement);
+			if (fbnElement instanceof final TypedSubApp tsa && !tsa.getVarConfigParams().isEmpty()) {
+				final EObject container = fbnElement.eContainer().eContainer();
```
oberlehner: Please use EcoreUtil.getRootContainer.
Then check for AutomationSystem and SubAppType

oberlehner (approved):

PR #1314, opened 2025-05-20
Moved Block Margin preferences into separate preference page

pazi146:


azoitl (changes requested):
Similar how it was for the diagram preferences now also for the block size preferences a message dialog that tells you that you either need to restart IDE or reopen your editors is needed so that the changes take effect should be added.

pazi146:
added the dialog

azoitl (approved):

PR #1331, opened 2025-05-26
Add builders for external libraries and type library

pazi146:
added library builder
removed library change listener
removed library plugin activator

added type library builder
removed TypeLibraryStarter extension point
removed event broker

added marker resolution to add new builders
adjusted library tests

cleaned up FordiacChangeListener


oberlehner (approved):

pazi146:
Replaced by #1354 - decision was made to preserve `FordiacResourceChangeListener`

PR #1333, opened 2025-05-26
Fix block margin project specific settings

pazi146:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
```diff
@@ -69,22 +72,26 @@ static boolean validateCollisions(final FBNetwork network, final DiagnosticChain
...
-				Predicate.not(FBNetworkElement::isInGroup), diagnostics, context);
+				Predicate.not(FBNetworkElement::isInGroup), diagnostics, context,
+				((LibraryElement) EcoreUtil.getRootContainer(network)).getTypeLibrary().getProject());
 	}
 
 	static boolean validateCollisions(final Group group, final DiagnosticChain diagnostics,
 			final Map<Object, Object> context) {
-		return validateCollisions(Optional.of(group), group.getGroupElements(), unused -> true, diagnostics, context);
+		return validateCollisions(Optional.of(group), group.getGroupElements(), unused -> true, diagnostics, context,
+				group.getTypeLibrary().getProject());
```
mx990: I believe this will always trigger an NPE, since a `Group` has no type entry and so `getTypeLibrary()` will always return `null`.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
```diff
@@ -69,22 +72,26 @@ static boolean validateCollisions(final FBNetwork network, final DiagnosticChain
 		if (parent.filter(p -> p.getTypeEntry() != null).isPresent()) {
 			return true; // ignore typed parent
 		}
+
 		return validateCollisions(parent.filter(FBNetworkAnnotations::isUnfoldedSubapp), network.getNetworkElements(),
-				Predicate.not(FBNetworkElement::isInGroup), diagnostics, context);
+				Predicate.not(FBNetworkElement::isInGroup), diagnostics, context,
+				((LibraryElement) EcoreUtil.getRootContainer(network)).getTypeLibrary().getProject());
```
mx990: Please extract this into a separate method and add both instanceof and null checks.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
pazi146: how would I get a project reference based on `Group`?

azoitl: as with FBNetworks via its root.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
pazi146: extracted into separate method

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/FBNetworkAnnotations.java
```diff
@@ -181,13 +188,24 @@ private static Diagnostic createInterfaceBarCollisionDiagnostic(final FBNetworkE
...
 	}
 
-	private static int getIntPreference(final Map<Object, Object> context, final String name) {
-		return ((Integer) context.computeIfAbsent(name, FBNetworkAnnotations::internalGetIntPreference)).intValue();
+	private static IProject getRootProject(final EObject eObject) {
+		if (eObject == null) {
+			return null;
+		}
+		final EObject root = EcoreUtil.getRootContainer(eObject);
+		if (root != null && root instanceof final LibraryElement libElem && libElem.getTypeLibrary() != null) {
```
mx990: Both `EcoreUtil.getRootContainer` and `instanceof` are null-safe, so we can safely simplify to:
```suggestion
		if (EcoreUtil.getRootContainer(eObject) instanceof final LibraryElement libElem && libElem.getTypeLibrary() != null) {
```

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #1353, opened 2025-05-30
Removes obsolete function block type files

franz-hoepfinger-4diac:
Deletes F_MUX_2_1.fbt and F_MUX_2_2.fbt from the typelibrary as these files are no longer needed in the project, reducing project size and maintenance overhead.


azoitl (changes requested):
Please do this on the 3.0.0 version of the typelib. Because this is the version for the latest 4diac FORTE

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-ide/pull/1358

PR #1354, opened 2025-06-02
Add builder for external libraries

pazi146:
added library builder
removed library change listener
removed library plugin activator

removed TypeLibraryStarter extension point
removed event broker

added marker resolution to add new builder
adjusted library tests


plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -466,19 +466,28 @@ public static boolean saveManifest(final Manifest manifest) {
...
-		if (manifest.getDependencies() != null) {
-			// ensure dependencies are sorted (can't use EList.sort())
-			final var dependencies = new LinkedList<>(manifest.getDependencies().getRequired());
-			manifest.getDependencies().getRequired().clear();
-			dependencies.forEach(d -> ManifestHelper.addDependency(manifest, d));
+	public static boolean sortManifestDependencies(final Manifest manifest) {
+		if (manifest.getDependencies() == null) {
+			return false; // nothing to do
+		}
+		// ensure dependencies are sorted (can't use EList.sort())
```
mx990: Could you use `ECollections.sort(EList<T>, Comparator<? super T>)` instead?

plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -466,19 +466,28 @@ public static boolean saveManifest(final Manifest manifest) {
...
+		final var dependencies = new ArrayList<>(manifest.getDependencies().getRequired());
+		manifest.getDependencies().getRequired().clear();
+		dependencies.forEach(d -> ManifestHelper.addDependency(manifest, d));
+		int index = 0;
+		for (final Required req : manifest.getDependencies().getRequired()) {
+			if (req != dependencies.get(index)) {
+				return true;
+			}
+			index++;
 		}
```
mx990: Could you not simply use:
```suggestion
		return !dependencies.equals(manifest.getDependencies().getRequired());
```

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -816,32 +612,39 @@ public boolean belongsTo(final Object family) {
...
+
 		final Manifest projectManifest = ManifestHelper.getContainerManifest(project);
 
 		if (projectManifest == null) {
+			SubMonitor.done(monitor);
 			return;
 		}
 
+		progress.split(4);
 		checkLibChanges();
```
mx990: I would suggest passing the splitted `SubMonitor` to each callee, so that it may also report progress and handle cancellation.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -850,7 +653,10 @@ public void resolveDependencies(final IProject project, final TypeLibrary typeLi
 			queue.add(req.getSymbolicName());
 		});
 
+		final var loopProgress = progress.split(70);
+
 		while (!queue.isEmpty()) {
```
mx990: Please handle cancellation inside the loop. I would also suggest extracting the loop into a separate method.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -901,6 +708,7 @@ public void resolveDependencies(final IProject project, final TypeLibrary typeLi
 
 		final List<ErrorMarkerBuilder> markerList = new LinkedList<>();
 
+		progress.split(15);
 		// import valid nodes
 		for (final var dnode : deps.values()) {
```
mx990: same here

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -923,6 +731,7 @@ public void resolveDependencies(final IProject project, final TypeLibrary typeLi
 			}
 		}
 
+		progress.split(5);
 		linked.values().forEach(f -> {
```
mx990: and here

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+		case IResource.FILE:
+			// check manifest files
+			// information on previous linked status is not available on delete
+			if (delta.getResource() instanceof final IFile file && LibraryManager.MANIFEST.equals(file.getName())) {
+				// project manifest:
+				if (file.getParent() instanceof IProject
+						&& (delta.getKind() & (IResourceDelta.ADDED | IResourceDelta.CHANGED)) != 0) {
+					// ignore manifest changes caused by full build
+					projectManifestChanged = checkProjectManifest;
+				}
```
mx990: I believe it may be easier to directly look for the manifest in the delta beforehand by using `IResourceDelta.findMember(IPath)`, instead of updating member flags in the visitor.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -933,6 +742,8 @@ public void resolveDependencies(final IProject project, final TypeLibrary typeLi
 
 		FordiacMarkerHelper.updateMarkers(project.getFile(MANIFEST), FordiacErrorMarker.LIBRARY_MARKER, markerList,
 				true);
+
+		SubMonitor.done(monitor);
```
mx990: It is not usually necessary to call `done()` in the callee anymore. Are you sure this is necessary in this case?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
pazi146: SubMonitor doc explicitly states that we're responsible for calling `done()` on the caller`s `IProgressMonitor` when we can't guarantee it being a `SubMonitor`

pazi146: Ignore my previous comment, will remove that `done()`

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+import org.eclipse.fordiac.ide.library.LibraryManager;
+import org.eclipse.fordiac.ide.library.model.library.Manifest;
+import org.eclipse.fordiac.ide.library.model.util.ManifestHelper;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacMarkerHelper;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
+
+public class LibraryBuilder extends IncrementalProjectBuilder {
+	private static final int MASK = IResourceDelta.ADDED | IResourceDelta.REMOVED | IResourceDelta.CHANGED;
+	private boolean projectManifestChanged = false;
```
mx990: > Remove the "projectManifestChanged" field and declare it as a local variable in the relevant methods.

from SonarQube

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+	private boolean checkProjectManifest = true;
```
mx990: I am not sure I fully understand the purpose of `checkProjectManifest`. It seems to suppress reacting to manifest changes in the next incremental build following a full build? What if the user changes the manifest?

plugins/org.eclipse.fordiac.ide.systemmanagement/src/org/eclipse/fordiac/ide/systemmanagement/nature/FordiacNature.java
```diff
@@ -107,6 +109,30 @@ public static boolean deconfigureExportBuilder(final IProjectDescription descrip
...
+			final ICommand command = description.newCommand();
+			command.setBuilderName(SystemManager.FORDIAC_LIBRARY_BUILDER_ID);
+			newCommands[commands.length] = command;
+			description.setBuildSpec(newCommands);
+			return true;
+		}
+		return false;
+	}
+
+	public static boolean deconfigureLibraryBuilder(final IProjectDescription description) {
```
mx990, line 125: I believe you also need to call this in `deconfigure()`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
pazi146: done

plugins/org.eclipse.fordiac.ide.systemmanagement/src/org/eclipse/fordiac/ide/systemmanagement/nature/FordiacNature.java
pazi146: added

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
pazi146: `ManifestHelper.saveManifest(manifest)` in `fullBuild` causes the builder to run again (with the changed manifest file as the delta)

Removing it would cause the resolver to run again (this should only be caused by the user changing the manifest file manually)

mx990: Why do you change the manifest in the builder in the first place? As the manifest will commonly be under version control, that inevitably leads to recurring conflicts if the builder keeps changing it automatically.

I believe such changes are better suited to be part of the editor, such as with a format handler. That would ensure the manifest is only changed due to an explicit user interaction.

pazi146: The manifest only changes if it needs to - in this case to ensure the dependencies are sorted (which is important for version control)

we don't have a fully functional editor for manifests at this point in time

oberlehner: > The manifest only changes if it needs to - in this case to ensure the dependencies are sorted (which is important for version control)
> 
> we don't have a fully functional editor for manifests at this point in time

@mx990 i think we discussed that issue today. please clarifiy if it is ok now

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+	public ISchedulingRule getRule(final int kind, final Map<String, String> args) {
+		return getProject();
+	}
+
+	private void fullBuild(final IProject project, final Manifest manifest, final IProgressMonitor monitor) {
+		final SubMonitor progress = SubMonitor.convert(monitor, "Library build", 1); //$NON-NLS-1$
+		if (ManifestHelper.sortManifestDependencies(manifest)) {
+			ManifestHelper.saveManifest(manifest);
+			checkProjectManifest = false;
+		}
```
mx990: Please remove the automatic resorting of the manifest in the builder.

pazi146: Ok, will remove it

mx990 (approved):

oberlehner (approved):

PR #1356, opened 2025-06-02
Not wanted Attribute for VarConfigs will not be added anymore

PStmmr:
Not wanted Attribute for VarConfigs will not be added anymore when changes are made in the property sheet.


azoitl (commented):
Is this need because you are copying the var config pins to the list of your typed subapps? If yes I think it would be better that you remove the attribute when copying.

azoitl (approved):

PR #1366, opened 2025-06-04
Improved search for VarConfigs

PStmmr:
Varconfigs are searched for in two ways.
First, based on existing saved varconfigs, and second, dynamically through the project.
This is the first step to later modifying and saving unsaved varconfigs.
Changes to a varconfig are now also visible in other entries that point to the same pin.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
@@ -15,11 +15,14 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.properties;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Spliterators;
-import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
+import java.util.Map;
+import java.util.Set;
 
+import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.application.Messages;
 import org.eclipse.fordiac.ide.application.editparts.FBNetworkEditPart;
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -141,21 +150,76 @@ protected void performRefreshAnnotations() {
...
+		final List<VarDeclaration> results = new ArrayList<>(resultMap.values());
+		return results;
+	}
+
+	private void collectVarConfigsRecursive(final INamedElement type, final Map<String, VarDeclaration> result,
+			final Set<INamedElement> visited, final String currentPrefix) {
+		if (!visited.add(type)) {
+			return;
+		}
+		if (type instanceof final Application app) {
```
azoitl: please turn this into a switch case with pattern matching.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -141,21 +150,76 @@ protected void performRefreshAnnotations() {
...
+			traverseFBNetwork(app.getFBNetwork().getNetworkElements(), result, visited, currentPrefix);
+		}
+		if ((type instanceof final SubApp sa) && (sa.getType() instanceof final SubAppType sat)) {
```
azoitl: you need to differentiate between typed and untyped subapp here. Better check for TypedSubApp and UntypedSubapp

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -141,21 +150,76 @@ protected void performRefreshAnnotations() {
...
+			}
+
+		}
+	}
+
+	private static void addPossibleVarConfigs(final FBNetworkElement fbne, final Map<String, VarDeclaration> result,
+			final String prefix) {
+		for (final IInterfaceElement elem : fbne.getInterface().getAllInterfaceElements()) {
+			if (elem instanceof final VarDeclaration vd && vd.isVarConfig()) {
+				addAndCopyElement(vd, prefix, result);
```
azoitl: you must only copy elements that are coming from a typed subapp instance. I guess you need a different call path for var configs originating from the application and its direct children (e.g., untyped subapp) and from var configs coming from typed subapp instances.

azoitl (changes requested):

azoitl (approved):

PR #1367, opened 2025-06-04
Add untyped SubApp support in interpreter

Felix-Schmid:
Also contains some minor code cleanup.


bwiesmayr (changes requested):
It seems to me that simply taking the references from the old transactions (eventoccurrences) could sometimes point to elements that are outdated (i..e, have been copied).

bwiesmayr (approved):

PR #1252, opened 2025-04-22
VarConfiguration Property sheet has been added for typed subapps in the FB Network Editor

PStmmr:

PR #1397, opened 2025-06-18
Fix IdentifierVerifier allowing dots

Felix-Schmid:
The current IdentifierVerifier allowed names with dots ("."). I believe this was a mistake and names with dots are not IEC 61499 compliant identifiers.
I also added such an identifier to the test cases.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
@@ -63,7 +63,7 @@ public final class IdentifierVerifier {
 	 *         message is contained in the Optional
 	 */
 	public static Optional<String> verifyIdentifier(final String identifier, final Object context) {
-		if (identifier == null || (!IDENTIFIER_PATTERN.matcher(identifier).matches() && !identifier.contains("."))) { //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


azoitl (commented):
can't we add this check into the regex? Or shouldn't the identifier regex be adjusted?

Felix-Schmid:
> can't we add this check into the regex? Or shouldn't the identifier regex be adjusted?

If the dot _should_ be allowed we could also adjust the regex, yes. But I believe the dot shouldn’t be allowed, in which case the regex is already correct – I just removed the additional contains check.

mx990:
I have checked the history and this was changed just recently with https://github.com/eclipse-4diac/4diac-ide/pull/1252, seemingly to allow VarConfig pins with dots in their name.

That said, I believe we should definitely *not* allow dots in identifiers globally like that, since this has great potential to cause problems in other areas.

If we need an exception to this rule for VarConfig pins during validation, I believe it should be added by overriding the `validateName` model invariant for `VarDeclaration`s and checking for `isVarConfig()` instead.

Aside from that, VarConfig pins should probably use `$` instead of `.` as a path separator anyhow, similar to member pins on struct (de)muxers.

azoitl:
@mx990 thx for finding the root cause. Yes it should definitely not be allowed. @Felix-Schmid I missread the PR. Sorry for that.
As I know where this is coming from I can find better solutions. But this PR should be accepted.

mx990 (approved):

azoitl (approved):

PR #1402, opened 2025-06-23
TypeEditor Warning Dialog for dirty type in BulkEditor

sebHollersbacher:


oberlehner:
Unfortunately, i cannot reproduce the situation but i managed to get an status "Other Editor is Dirty". But i didnt had any dirty  editors opened as can be seen in the screenshot
image

sebHollersbacher:
@oberlehner 
Has been fixed
I think this situation was created by having the other editor dirty so the message shows and then saving the Editor

oberlehner (approved):

oberlehner (approved):

Issue #1406, opened 2025-06-24
Broken system configuration with standard libraries in version 3.0.0


mx990:
There are multiple problems in the system configuration with standard libraries in version 3.0.0, since the latter moves several core types into packages.

1. When creating a new device, the `EMB_RES` resource is not added automatically:
  - with version 1.0.0:
    ![Image](https://github.com/user-attachments/assets/3fa2c413-34c0-4f57-a615-33241ca90ba7)
  - with version 3.0.0:
    ![Image](https://github.com/user-attachments/assets/61681fb3-eb3c-4612-aa31-c2b13da45d9a)

2. After manually adding the `EMB_RES` resource, the type for the default `E_RESTART` block is not found:
  - with version 1.0.0:
    ![Image](https://github.com/user-attachments/assets/08efc4e0-423f-4a9a-8b3b-295846cfc59a)
  - with version 3.0.0:
    ![Image](https://github.com/user-attachments/assets/dbae84b8-1a4d-4f04-8ae6-e6560ede231d)

mx990:
It seems the first problem is caused by hard-coded resource type names in https://github.com/eclipse-4diac/4diac-ide/blob/bf0e8df5108cb6298e21f4aa22cf2fa9aaf0b592/plugins/org.eclipse.fordiac.ide.systemconfiguration/src/org/eclipse/fordiac/ide/systemconfiguration/commands/DeviceCreateCommand.java#L131-L136

mx990:
A quick search also revealed a related problem in https://github.com/eclipse-4diac/4diac-ide/blob/bf0e8df5108cb6298e21f4aa22cf2fa9aaf0b592/plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/DeploymentEvaluatorConfiguration.java#L25-L29

mx990:
The second problem seems to be caused by https://github.com/eclipse-4diac/4diac-ide/blob/bf0e8df5108cb6298e21f4aa22cf2fa9aaf0b592/data/typelibrary/hardware-3.0.0/typelib/resources/EMB_RES.res#L10
and also https://github.com/eclipse-4diac/4diac-ide/blob/bf0e8df5108cb6298e21f4aa22cf2fa9aaf0b592/data/typelibrary/hardware-3.0.0/typelib/resources/Config_EMB_RES.res#L11

azoitl:
[...]

PR #1407, opened 2025-06-24
Fix package name of E_RESTART in EMB_RES version 3.0.0

mx990:
Fixes https://github.com/eclipse-4diac/4diac-ide/issues/1406


pazi146 (approved):

azoitl:
@mx990  didn't we say we wanted to do it like java lang that for these FBs no package names are needed.

PR #1410, opened 2025-06-24
Remove distributed system listener

azoitl:


oberlehner (approved):

azoitl:
I reworked and cleaned several of the label providers. I have the impression to get a more consistent update behavior of error marker changes. What would be needed if this work also under windows is that a filter for 4diac Projects is added to the system explorer.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/navigator/FBTypeContentProvider.java
```diff
@@ -38,4 +52,35 @@ public boolean hasChildren(final Object element) {
...
+					for (final IMarkerDelta markerDelta : delta.getMarkerDeltas()) {
+						IResource resource = markerDelta.getResource();
+						while (resource != null) {
+							list.add(resource);
+							resource = resource.getParent();
+						}
+					}
+					return true;
+				});
+				if (!list.isEmpty()
```
pazi146: I suggest moving the `Display.getDefault().asyncExec(() -> {` here
No real reason to handle the delta on the display thread

azoitl: Thx for pointing that out. done.

pazi146 (approved):

PR #1412, opened 2025-06-26
Merge Branch Freeze into Develop

azoitl:


mx990 (changes requested):
It looks to me as if the actual merge commit is from develop into freeze?

azoitl:
Then github ui messed with me. Will do a dedicated merge commit locally.

PR #1419, opened 2025-06-30
Removed FixedScopedPreferenceStore

pazi146:
No longer necessary as Bug was fixed in ScopedPreferenceStore


ernstblechaPT:
As it took me a moment to find the Bug (Memory Leak) that was fixed:

First discussed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2878 and fixed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2886

mx990 (approved):

azoitl (approved):

PR #1431, opened 2025-07-03
Add providers for resource, device, and segment extensions

mx990:


azoitl:
@mx990 with the fix in PR #1442 this is not needed anymore or?

mx990:
I believe we should still add this to enable proper validation in resource, device, and segment types via the builder.

azoitl (approved):

PR #1433, opened 2025-07-04
Save and build before rename refactoring

mx990:


oberlehner:
Do we really want that behaviour? 
I don't know if we want to save everything before we start a rename. 
What if i have an dirty editor? Then everthing will be overwritten.
Wouldn't it be better to inform the user before we start a refactoring ?

mx990:
We definitely need to save and build everything before refactoring or risk corrupting files in a way that cannot be undone. When a refactoring is triggered inside the ST editor, it automatically saves all dirty editors beforehand via the Xtext framework. This PR simply extends the same behavior to when the refactoring is triggered outside of the ST editor.

I agree that it would make sense to give the user the option the cancel the refactoring when there are dirty editors. However, I believe that should be part of a separate PR and then also include the ST editor, in order to ensure the same behavior regardless where the refactoring is triggered from.

oberlehner:
> We definitely need to save and build everything before refactoring or risk corrupting files in a way that cannot be undone. When a refactoring is triggered inside the ST editor, it automatically saves all dirty editors beforehand via the Xtext framework. This PR simply extends the same behavior to when the refactoring is triggered outside of the ST editor.
> 
> I agree that it would make sense to give the user the option the cancel the refactoring when there are dirty editors. However, I believe that should be part of a separate PR and then also include the ST editor, in order to ensure the same behavior regardless where the refactoring is triggered from.

i can go with that

oberlehner (approved):

PR #1445, opened 2025-07-08
Create Global Constants Type Editor

pazi146:
create base editor and GlobalConstantsTypeInstanceSearch 
adapt relevant sections to handle ITextSelection
added LocalLiveSearchContext (ignores linked files)


plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+
+	public GlobalConstantsTypeInstanceSearch(final GlobalConstantsEntry gcEntry) {
+		super(new LocalLiveSearchContext(gcEntry.getTypeLibrary()), createSearchFilter(gcEntry),
+				new GlobalConstantsTypeInstanceSearchChildrenProvider());
+	}
+
+	private static IEC61499SearchFilter createSearchFilter(final GlobalConstantsEntry gcEntry) {
+		return searchCandidate -> ((searchCandidate instanceof final VarDeclaration varDecl && isGlobalConstantValue(
+				varDecl, gcEntry /* , gcEntry.getFullTypeName().split(ImportHelper.PACKAGE_NAME_DELIMITER) */))
+				|| (searchCandidate instanceof final Attribute attr && isGlobalConstantValue(attr, gcEntry)));
```
mx990: Please note that global constants may also be used inside algorithms, methods, and functions. Also, please do not commit commented code.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+//		if (match == gcEntrySegments.length) {
+//			return true; // full match
+//		}
+//
+//		for (final Import imp : ImportHelper.getContainerImports(varDecl)) {
+//			final String namespace = imp.getImportedNamespace();
+//			if (namespace.endsWith(ImportHelper.WILDCARD_IMPORT_SUFFIX)) {
+//
+//			}
+//		}
```
mx990: Please do not commit commented code.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+				|| (searchCandidate instanceof final Attribute attr && isGlobalConstantValue(attr, gcEntry)));
+	}
+
+	private static boolean isGlobalConstantValue(final VarDeclaration varDecl,
+			final GlobalConstantsEntry gcEntry /* , final String[] gcEntrySegments */) {
+		if (varDecl.getValue() == null) {
+			return false;
+		}
+		final String value = varDecl.getValue().getValue();
+		if (value == null || value.isBlank() || !value.contains(ImportHelper.PACKAGE_NAME_DELIMITER)) {
```
mx990: I am not sure why you are testing for a package name delimiter here. It is perfectly valid, albeit discouraged, to import a global constant individually (e.g., `math::Constants::PI`) and then use only the name of the constant (e.g., `PI`) in code or (initial) values.

I would suggest to test for potential matches by checking for a letter character and that it is not a `VariableOperations.isSimpleInitialValue(...)` (you will need to make it public) instead.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+//			if (namespace.endsWith(ImportHelper.WILDCARD_IMPORT_SUFFIX)) {
+//
+//			}
+//		}
+		return isGlobalConstantValue(value, varDecl, gcEntry);
+	}
+
+	private static boolean isGlobalConstantValue(final Attribute attr, final GlobalConstantsEntry gcEntry) {
+		final String value = attr.getValue();
+		if (value == null || value.isBlank() || !value.contains(ImportHelper.PACKAGE_NAME_DELIMITER)) {
```
mx990: same here

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		return isGlobalConstantValue(value, attr, gcEntry);
+	}
+
+	private static boolean isGlobalConstantValue(final String value, final EObject context,
+			final GlobalConstantsEntry gcEntry) {
+		final String name = value.substring(0, value.lastIndexOf(ImportHelper.PACKAGE_NAME_DELIMITER));
+		final TypeLibrary typeLibrary = gcEntry.getTypeLibrary();
+		return gcEntry == ImportHelper.resolveImport(name, context, typeLibrary::getGlobalConstantsEntry, n -> null);
```
mx990: This will not give you proper results when a constant is used without qualifier. I would suggest to obtain an `ISearchSupport` for the `context` and use a custom `IModelMatcher` to test for the FQN of the referenced feature (minus the last segment).

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
```diff
@@ -12,33 +12,40 @@
...
-import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.IEditorSite;
+import org.eclipse.ui.IWorkbenchPart;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.views.properties.tabbed.ITabbedPropertySheetPageContributor;
 import org.eclipse.xtext.ui.editor.XtextEditor;
 
 public class GlobalConstantsEditor extends XtextEditor
-		implements CommandStackEventListener, ITabbedPropertySheetPageContributor {
+		implements CommandStackEventListener, ITabbedPropertySheetPageContributor, ITypeEditorPage {
```
mx990: I have extracted a common Xtext type editor page in https://github.com/eclipse-4diac/4diac-ide/pull/1450. Please use the new `XtextTypeEditorPage` as the base class for this type editor page. That should already handle almost everything of what is needed for this to work properly inside a type editor. There are also some additional required methods: https://github.com/eclipse-4diac/4diac-ide/blob/243f50d0161cbc1d4beab881ffd0dd09473d4271/plugins/org.eclipse.fordiac.ide.fbtypeeditor.st/src/org/eclipse/fordiac/ide/fbtypeeditor/st/StructuredTextFBTypeEditor.java#L97-L105

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
```diff
@@ -12,33 +12,40 @@
...
+	private CommandStack commandStack;
 	private final ActionRegistry actionRegistry = new ActionRegistry();
 
 	@Override
 	public void init(final IEditorSite site, final IEditorInput input) throws PartInitException {
 		super.init(site, input);
-		commandStack.addCommandStackEventListener(this);
+		setPartName(Messages.GlobalConstantsEditor);
+		setTitleImage(FordiacImage.ICON_DATA_TYPE.getImage());
 	}
```
mx990: I would suggest overriding `doSetInput` instead, see https://github.com/eclipse-4diac/4diac-ide/blob/243f50d0161cbc1d4beab881ffd0dd09473d4271/plugins/org.eclipse.fordiac.ide.fbtypeeditor.st/src/org/eclipse/fordiac/ide/fbtypeeditor/st/StructuredTextFBTypeEditor.java#L57-L62

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/messages.properties
```diff
@@ -14,3 +14,4 @@ GlobalConstantsOutlineTreeProvider_VarGlobalDeclarationBlockText=VAR_GLOBAL
 GlobalConstantsOutlineTreeProvider_VarGlobalDeclarationBlockConstantText=VAR_GLOBAL CONSTANT
 GlobalConstQuickFix_AddConstantKeyword=Add CONSTANT modifier
 GlobalConstQuickFix_RenameGlobalConstants=Rename global constants
+GlobalConstantsEditor=Global Constants Editor
```
mx990: I would suggest a shorter name for the editor tab, such as `Constants`.

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
```diff
@@ -12,8 +12,7 @@
    <extension
 		point="org.eclipse.ui.editors">
 		<editor
-			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.fordiac.ide.globalconstantseditor.ui.editor.GlobalConstantsEditor"
-			contributorClass="org.eclipse.ui.editors.text.TextEditorActionContributor"
+			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.editor.GlobalConstantsBaseEditor"
 			default="true"
 			extensions="globalconsts,gcf"
```
mx990: Please create a separate editor extension for the `globalconsts` file extension, since that is a plain-text file. There is an example on how this should look like in https://github.com/eclipse-4diac/4diac-ide/blob/243f50d0161cbc1d4beab881ffd0dd09473d4271/plugins/org.eclipse.fordiac.ide.structuredtextfunctioneditor.ui/plugin.xml#L4-L15

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsBaseEditor.java
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	@Override
+	protected GlobalConstantsEntry getTypeEntry() {
+		return (GlobalConstantsEntry) super.getTypeEntry();
+	}
+
+	public GlobalConstantsEntry getGlobalConstants() {
+		return getTypeEntry();
+	}
```
mx990: I believe this separate method is unnecessary?

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsBaseEditor.java
```diff
@@ -0,0 +1,46 @@
...
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.globalconstantseditor.ui.editor;
+
+import org.eclipse.fordiac.ide.model.search.dialog.AbstractTypeEntryDataHandler;
+import org.eclipse.fordiac.ide.model.search.dialog.GlobalConstantsEntryDataHandler;
+import org.eclipse.fordiac.ide.model.typelibrary.GlobalConstantsEntry;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
+import org.eclipse.fordiac.ide.typeeditor.AbstractTypeEditor;
+
+public class GlobalConstantsBaseEditor extends AbstractTypeEditor {
```
mx990: I would suggest renaming this to `GlobalConstantsTypeEditor`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsBaseEditor.java
pazi146: removed

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/messages.properties
pazi146: I'll rename it to `Global Constants` to stay consistent with other uses

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
pazi146: What would be the difference?
Using `class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.fordiac.ide.globalconstantseditor.ui.editor.GlobalConstantsEditor"
			contributorClass="org.eclipse.ui.editors.text.TextEditorActionContributor` for `globalconsts` file ending?

mx990: I meant to use a plain Xtext editor for `globalconst` files:
```xml
class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.xtext.ui.editor.XtextEditor"
```

The type editor cannot open plain-text files, since it expects a library element in XML format.

pazi146: Makes sense, thank you for the explanation

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,153 @@
...
+			if (obj instanceof final StructuredType structType) {
+				return SearchChildrenProviderHelper.getStructChildren(structType);
+			}
+			if (obj instanceof final AttributeDeclaration attrdecl) {
+				return SearchChildrenProviderHelper.getAttributeDeclChildren(attrdecl);
+			}
+
+			if (obj instanceof final FBNetworkElement elem) {
+				return Stream.concat(elem.getAttributes().stream(),
+						SearchChildrenProviderHelper.getInterfaceListChildren(elem.getInterface()));
```
oberlehner, line 132: are you sure we need to search outputVars too ?
I think this is not necessary for global constants.

mx990: I believe there may be output variables with attributes that use global constants.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
pazi146: Ran into the following issue:
- Global constant `gc` with VAR `A`
- Import in application: `gc::A`
- use `A` as initial value

`ImportHelper.resolveImport()` doesn't provide the correct result as it searches for `gc::A` in the typelibry (where only `gc` is registered)
How do I properly use it in this case?

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
```diff
@@ -12,17 +12,35 @@
    <extension
 		point="org.eclipse.ui.editors">
 		<editor
-			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.GlobalConstantsExecutableExtensionFactory:org.eclipse.fordiac.ide.globalconstantseditor.ui.editor.GlobalConstantsEditor"
-			contributorClass="org.eclipse.ui.editors.text.TextEditorActionContributor"
+			class="org.eclipse.fordiac.ide.globalconstantseditor.ui.editor.GlobalConstantsTypeEditor"
```
mx990, line 15: Please also add the missing `contributorClass`, so that action contributions (incl. undo/redo) will work properly:
```xml
contributorClass="org.eclipse.fordiac.ide.typeeditor.TypeEditorContributor"
```

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
```diff
@@ -23,16 +27,19 @@
...
+
+public class GlobalConstantsEditor extends XtextTypeEditorPage
 		implements CommandStackEventListener, ITabbedPropertySheetPageContributor {
 
-	private final CommandStack commandStack = new CommandStack();
+	@Inject
+	private STCoreEditorPreferences editorPreferences;
+
+	private CommandStack commandStack;
 	private final ActionRegistry actionRegistry = new ActionRegistry();
```
mx990: Neither a command stack nor an action registry should be needed here anymore, since both undo/redo and actions are managed by the text editor framework and action contributor.

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
```diff
@@ -23,16 +27,19 @@
...
 import org.eclipse.ui.IEditorSite;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.views.properties.tabbed.ITabbedPropertySheetPageContributor;
-import org.eclipse.xtext.ui.editor.XtextEditor;
 
-public class GlobalConstantsEditor extends XtextEditor
+import com.google.inject.Inject;
+
+public class GlobalConstantsEditor extends XtextTypeEditorPage
 		implements CommandStackEventListener, ITabbedPropertySheetPageContributor {
```
mx990: Implementing these interfaces should no longer be needed. The same with `getAdapter(...)` below.

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,128 @@
...
+				&& isGlobalConstantValue(varDecl, gcEntry))
+				|| (searchCandidate instanceof final Attribute attr && isGlobalConstantValue(attr, gcEntry)));
+	}
+
+	private static boolean isGlobalConstantValue(final VarDeclaration varDecl, final GlobalConstantsEntry gcEntry) {
+		if (varDecl.getValue() == null) {
+			return false;
+		}
+		final String value = varDecl.getValue().getValue();
+		if (value == null || value.isBlank() || !Character.isLetter(value.charAt(0))
```
mx990: Please check that none of the characters is a letter and not just the first. For example, `1 + Math::PI` is also a valid initial value but would currently not be considered.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
mx990: This will still not work when the value is more complex, such as `1 + Math::PI` or `Other::CONST + Math::PI` (when looking for `Math`). A constant may also appear in a string, such as `'The value of $'PI$' is 3.14159...'`, which should not produce false positives.

I would again *strongly* suggest to use `ISearchSupport` to properly parse the value and look for matching references, instead of attempting to handle all of these cases here.

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
pazi146: removed

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/src/org/eclipse/fordiac/ide/globalconstantseditor/ui/editor/GlobalConstantsEditor.java
pazi146: removed

plugins/org.eclipse.fordiac.ide.globalconstantseditor.ui/plugin.xml
pazi146: added

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
pazi146: searching these as well now

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
pazi146: using `ISearchSupport` now

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
pazi146: changed to checking entire string for letters

plugins/org.eclipse.fordiac.ide.model.search/src/org/eclipse/fordiac/ide/model/search/types/GlobalConstantsTypeInstanceSearch.java
```diff
@@ -0,0 +1,173 @@
...
+		case final STFunctionBody body: {
+			searchSupport = ISearchFactory.createSearchSupport(body, STFunctionBody.class);
+			break;
+		}
+		default: {
+			return false;
+		}
+		}
+
+		return searchSupport.search(matcher).count() > 0;
```
mx990: I would suggest to simplify this to:

```suggestion
		final ISearchSupport searchSupport = ISearchFactory.createSearchSupport(context,
				context.eClass().getInstanceClass());
		return searchSupport != null && searchSupport.search(matcher).count() > 0;
```

mx990 (changes requested):
Please make sure your ECA is still valid.

mx990 (approved):

pazi146:
renewed my ECA

PR #1466, opened 2025-07-21
Changed Attribute Target Lock to have more options

sebHollersbacher:


oberlehner:
image

Member of DataType has to be moved to instances

oberlehner (approved):

PR #1467, opened 2025-07-21
adjust ant lib element export

pazi146:
use 4diac exporter instead of standard xml export

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
+++ b/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
@@ -17,35 +17,22 @@ package org.eclipse.fordiac.ide.ant.ant;
 import java.io.File;
 import java.io.IOException;
 
-import javax.xml.XMLConstants;
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.transform.Transformer;
-import javax.xml.transform.TransformerConfigurationException;
-import javax.xml.transform.TransformerException;
-import javax.xml.transform.TransformerFactory;
-import javax.xml.transform.dom.DOMSource;
-import javax.xml.transform.stream.StreamResult;
-
 import org.apache.tools.ant.BuildException;
 import org.eclipse.core.resources.IFile;
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.fordiac.ide.export.ExportFilter;
-import org.w3c.dom.Document;
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-import org.xml.sax.SAXException;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.model.resource.FordiacTypeResourceFactory;
+import org.eclipse.fordiac.ide.model.typelibrary.AttributeTypeEntry;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
+import org.eclipse.fordiac.ide.model.util.LibraryElementHashException;
+import org.eclipse.fordiac.ide.model.util.LibraryElementHasher;
 
 public abstract class AbstractExportLibElements extends AbstractFBTask {
 	private static final String ANT_CONVERT_TASK_DIRECTORY_NAME = "converted_FBs"; //$NON-NLS-1$
-
-	private static final String SIMPLEFB_XML_ATTRIBUTE = "SimpleFB"; //$NON-NLS-1$
-	private static final String BASICFB_XML_ATTRIBUTE = "BasicFB"; //$NON-NLS-1$
-	private static final String COMPOSITEFB_XML_ATTRIBUTE = "FBNetwork"; //$NON-NLS-1$
-	private static final String FUNCTIONBODY_XML_ATTRIBUTE = "FunctionBody"; //$NON-NLS-1$
-
-	private final DocumentBuilder builder = getDocumentBuilder();
-	private final Transformer transformer = getTransformer();
 
 	@Override
 	protected String getExportDirectoryDefault() {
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +70,42 @@ protected ExportFilter getExportFilter() {
...
 		}
-		removeAllAlgorithms(document);
 
-		final DOMSource src = new DOMSource(document);
-		final StreamResult res = new StreamResult(new File(folder + File.separator + file.getName()));
 		try {
-			transformer.transform(src, res);
-		} catch (final TransformerException e) {
+			element.setAttribute(TypeLibraryTags.TYPE_HASH_ATTRIBUTE_NAME, IecTypes.ElementaryTypes.STRING,
+					LibraryElementHasher.hash(element), null);
```
mx990: We need to calculate the hash *before* transforming the FB, since we want the original hash in the attribute.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +70,42 @@ protected ExportFilter getExportFilter() {
...
-			factory.setFeature("http://apache.org/xml/features/nonvalidating/load-dtd-grammar", false); //$NON-NLS-1$
-			factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
-			return factory.newDocumentBuilder();
-		} catch (final ParserConfigurationException e) {
+		try (OutputStream outputStream = new FileOutputStream(new File(folder, file.getName()))) {
+			final AbstractTypeExporter exporter = getTypeExporter(element);
+			try (InputStream inputStream = exporter.getFileContent()) {
+				inputStream.transferTo(outputStream);
+			}
+		} catch (final IOException e) {
```
mx990: I would suggest using a `FordiacTypeResource` to save the type, instead of duplicating `getTypeExporter` below.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +70,42 @@ protected ExportFilter getExportFilter() {
...
 		log(file.toString());
 
-		Document document;
-		try {
-			document = builder.parse(file.getLocation().toFile());
-		} catch (IOException | SAXException e) {
-			throw new BuildException(e);
+		final TypeEntry entry = TypeLibraryManager.INSTANCE.getTypeEntryForFile(file);
+		LibraryElement element = entry.copyType();
+		if (element instanceof final FBType fb) {
```
mx990: I believe a FunctionFBType should still remain as such, except for its body.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/TransformUtil.java
```diff
@@ -0,0 +1,39 @@
...
+ *   Patrick Aigner - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.ant.ant;
+
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.libraryElement.FBType;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementFactory;
+import org.eclipse.fordiac.ide.model.libraryElement.ServiceInterfaceFBType;
+
+public class TransformUtil {
```
mx990: I would suggest using a custom `EcoreUtil.Copier` that always creates an SIFB instance instead. The `Copier` should already skip any EMF attributes that are not present in the new instance.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/TransformUtil.java
```diff
@@ -0,0 +1,39 @@
...
+	public static ServiceInterfaceFBType fbToSIFB(final FBType fb) {
+		final ServiceInterfaceFBType newType = LibraryElementFactory.eINSTANCE.createServiceInterfaceFBType();
+
+		fb.getAttributes().forEach(attr -> newType.setAttribute(EcoreUtil.copy(attr.getAttributeDeclaration()),
+				attr.getName(), attr.getComment()));
```
mx990: Please just copy the `Attribute`s and make sure *not* to copy the declarations.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +70,42 @@ protected ExportFilter getExportFilter() {
...
+			element = TransformUtil.fbToSIFB(fb);
 		}
-		removeAllAlgorithms(document);
 
-		final DOMSource src = new DOMSource(document);
-		final StreamResult res = new StreamResult(new File(folder + File.separator + file.getName()));
 		try {
-			transformer.transform(src, res);
-		} catch (final TransformerException e) {
+			element.setAttribute(TypeLibraryTags.TYPE_HASH_ATTRIBUTE_NAME, IecTypes.ElementaryTypes.STRING,
```
mx990: Please create an attribute with the correct attribute declaration, instead of a free attribute.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
pazi146: Would need to subclass it to deal with the URI check in the `doSave()` method

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
pazi146: yes, that makes more sense

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
mx990: I believe that check is left over from a prior implementation that utilized the type entries for saving. I would suggest to remove it.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
pazi146: should I put a `static final AttributeDeclaration` of the hash attribute somewhere (documentation might also be a candidate)?
Similar to `org.eclipse.fordiac.ide.model.datatype.helper.InternalAttributeDeclarations`

mx990: There already is a `TypeHash` attribute declaration in the standard library[^1][^2]. You should be able to look it up from the respective type library using `TypeLibrary.getAttributeTypeEntry(String)` with either the `TYPE_HASH_ATTRIBUTE_NAME` or `TYPE_HASH_ATTRIBUTE_FULL_NAME`, depending on which version is present.

[^1]: [data/typelibrary/core-1.0.0/typelib/TypeHash.atp](https://github.com/eclipse-4diac/4diac-ide/blob/develop/data/typelibrary/core-1.0.0/typelib/TypeHash.atp)
[^2]: [data/typelibrary/core-3.0.0/typelib/TypeHash.atp](https://github.com/eclipse-4diac/4diac-ide/blob/develop/data/typelibrary/core-3.0.0/typelib/TypeHash.atp)

pazi146: One major issue: the `core` typelib isn't guaranteed to be included

mx990: If neither attribute declaration is present, I would suggest throwing an exception with a descriptive message that the attribute is missing.

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/TransformUtil.java
pazi146: I've overwritten the `getTarget(final EClass eClass)` method in `Copier`, but that leads to exceptions on copying as features are not filtered.
Which means I'd need to filter them manually - which I already do in TransformUtil

pazi146: nevermind, found a simple way to do it

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/SIFBCopier.java
```diff
@@ -0,0 +1,46 @@
...
+				|| eClass == LibraryElementPackage.Literals.SUB_APP_TYPE) {
+			return LibraryElementPackage.Literals.SERVICE_INTERFACE_FB_TYPE;
+		}
+		return super.getTarget(eClass);
+	}
+
+	@Override
+	protected Setting getTarget(final EStructuralFeature eStructuralFeature, final EObject eObject,
+			final EObject copyEObject) {
+		if (ExtendedMetaData.INSTANCE.getAffiliation(copyEObject.eClass(), eStructuralFeature) == null) {
```
mx990: I would suggest to check via the class itself instead:
```suggestion
		if (!eStructuralFeature.getEContainingClass().isInstance(copyEObject)) {
```

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +51,45 @@ protected ExportFilter getExportFilter() {
...
-	}
 
-	private static void removeAllAlgorithms(final Document doc) {
-		NodeList algo = doc.getElementsByTagName(SIMPLEFB_XML_ATTRIBUTE);
-		for (int i = 0; i < algo.getLength(); i++) {
-			final Node parent = algo.item(i).getParentNode();
-			parent.removeChild(algo.item(i));
+		if (element instanceof final FBType fb) {
+			if (fb instanceof final FunctionFBType func) {
+				func.setBody(null);
```
mx990: I would suggest to handle this also in the `SIFBCopier` by excluding the body containment reference. Maybe then rename it to `ExportCopier`?

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/AbstractExportLibElements.java
```diff
@@ -61,72 +51,45 @@ protected ExportFilter getExportFilter() {
...
 
-		Document document;
-		try {
-			document = builder.parse(file.getLocation().toFile());
-		} catch (IOException | SAXException e) {
-			throw new BuildException(e);
-		}
-		removeAllAlgorithms(document);
+		final TypeEntry entry = TypeLibraryManager.INSTANCE.getTypeEntryForFile(file);
+		LibraryElement element = entry.copyType();
```
mx990: We now copy the element twice, once here and once in the `Copier` below. If we also handle function bodies in the `Copier` and thus never modify the original, I would suggest to simply use `getType()` directly.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/ExportCopier.java
```diff
@@ -0,0 +1,50 @@
...
+import org.eclipse.emf.ecore.EStructuralFeature.Setting;
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementPackage;
+
+public class ExportCopier extends EcoreUtil.Copier {
+	private static final long serialVersionUID = 1L;
+
+	@Override
+	protected EClass getTarget(final EClass eClass) {
+		if (eClass == LibraryElementPackage.Literals.FUNCTION_BODY
```
ernstblechaPT, line 27: This seems to leave in the `<FunctionBody>`-Tag that was removed by the old export (new on the left side, old on the right side in the picture)

image

mx990: @pazi146 Please add a check in the `FCTExporter` to skip the tag when the body is null.

ernstblechaPT (approved):

azoitl (approved):

mx990 (approved):

PR #1478, opened 2025-07-24
Use empty value instead of placeholder when editing initial values

mx990:
Use empty value instead of `computing` or `too large` placeholder when editing initial values. This avoids having to manually delete the placeholder when editing and also keeps it from being applied on commit.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/ValueEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/ValueEditPart.java
@@ -376,9 +376,17 @@ public class ValueEditPart extends AbstractGraphicalEditPart implements NodeEdit
 		final IInterfaceElement interfaceElement = getIInterfaceElement();
 		if (interfaceElement instanceof final VarDeclaration varDecl) {
 			return new InitialValueDirectEditManager(this, new FigureCellEditorLocator(getFigure()), varDecl,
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/ValueEditPart.java
```diff
@@ -375,8 +375,11 @@ protected IFigure createFigure() {
 	public DirectEditManager createDirectEditManager() {
 		final IInterfaceElement interfaceElement = getIInterfaceElement();
 		if (interfaceElement instanceof final VarDeclaration varDecl) {
-			return new InitialValueDirectEditManager(this, new FigureCellEditorLocator(getFigure()), varDecl,
-					getFigure().getText());
+			final String text = getFigure().getText();
+			if (FordiacMessages.ComputingPlaceholderValue.equals(text) || FordiacMessages.ValueTooLarge.equals(text)) {
+				return new InitialValueDirectEditManager(this, new FigureCellEditorLocator(getFigure()), varDecl, ""); //$NON-NLS-1$
```
azoitl: how about clearing the value in text instead of duplicating the object creation?

mx990: Good point. I have extracted the text handling into a separate method.

azoitl (approved):

PR #1479, opened 2025-07-24
Only set navigation location of type entry adapter on part activation

azoitl:
when the type entry adapter needs to update typed subapps to have the correct entry it needs to restore the viewer. This lead to wrong editors getting activated. This change postpones this to the activation of the part.

As part of this change also the reload checks for bulk editor synchronisation are moved to the same infrastructure.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/AbstractTypeEntryAdapter.java
+++ b/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/AbstractTypeEntryAdapter.java
@@ -18,26 +18,57 @@ import org.eclipse.emf.common.notify.impl.AdapterImpl;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IPartListener;
+import org.eclipse.ui.IPartService;
+import org.eclipse.ui.IWindowListener;
+import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
 
 public abstract class AbstractTypeEntryAdapter extends AdapterImpl {
+
+	private final IEditorPart editor;
+	private ActivationListener activationListener;
 	private boolean reloadPending;
 
 	protected abstract void reloadEditorType();
 
-	protected abstract IEditorPart getEditor();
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/AbstractTypeEntryAdapter.java
```diff
@@ -64,4 +83,62 @@ protected int openFileChangedDialog() {
...
+
+			@Override
+			public void partVisible(final IWorkbenchPartReference partRef) {
+				// nothing to do
+			}
+
+		};
+
+		final IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+		window.getPartService().addPartListener(partListener);
```
mx990: I would suggest to use individual listeners per editor, so that we may properly dispose of them when the editor is closed. I would also suggest to listen to window activations[^1].

[^1]: see also `org.eclipse.ui.texteditor.AbstractTextEditor.ActivationListener`

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -186,4 +193,11 @@ private static void handleDataTypeEntryUpdate(final LibraryElement editedElement
 		}).filter(Objects::nonNull).forEach(Command::execute);
 	}
 
+	private void performLocationRestore() {
+		if (location != null) {
+			location.restoreLocation();
```
mx990, line 206: Please also `dispose()` the `location` once it is no longer needed.

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/AbstractTypeEntryAdapter.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: good point, done.

oberlehner (approved):

mx990 (approved):

PR #1483, opened 2025-07-28
Add replay debugging plugins

cochicde:
A big change, I know. 

But, it's divided into two plugins: non-gui and gui stuff. 

Non-gui: the model of the replay navigator basically. Classes to represent all the information needed for replay debugging. Then,  the ReplayNavigator calculates the new state when moving forward/backwards.
Also in this plugin, the Debugging targets, which are instantiated when the Debug configuration is launched. The target contain a device which handles the watches, a device contains resources and each resource manages a ReplayNavigator. Existing ReplayNavigators are managed by a singleton.

GUI: Listen to the singleton when ReplayNavigators are registered. And then, creates beautiful widgets to show and interact with each ReplayNavigator. 

And basically, that's it. Want to try? Check the docu http://github.com/cochicde/4diac-documentation/blob/add-replay-ide/src/other_features/replayDebugger.adoc


azoitl (changes requested):
As this is a rather large PR I would prefere not to squash all commits into one. Could you clean-up the commit history a bit so that it consists of sensible steps.

Furthermore, we have now for each of our project a set of project specific java settings that ensure common formatting and common code clean-ups on save. Please add these settings to your new plugins. The best is to copy the two files ' org.eclipse.jdt.core.prefs` and `org.eclipse.jdt.ui.prefs` from an existing 4diac IDE plugin.

cochicde:
all right! separated into commits and nice formatting applied. I think now is ready for review.

azoitl (approved):

PR #1484, opened 2025-07-29
add instance property pages for typed subapps

pazi146:


oberlehner:
image
I think we dont need Events,Data,Adapter any more now, this is duplicate information.
Furthermore, i think this should be also propagated to all other fb instances, like simple, sifb,etc.. 
@ernstblechaPT what do you think ?

pazi146:
@oberlehner You're misunderstanding something here - this is _not_ duplicate information - one is the information of the instance, the other information on the type.

Instance information will of course start out the same as the type, but specific fields can be edited

azoitl (approved):

PR #1486, opened 2025-07-29
Changed size calculation to correctly represent figure

sebHollersbacher:
Model size calculation for Collisions was not calculated correctly depending on the pin ordering

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Martin Erich Jobst
+ * Copyright (c) 2023, 2025 Martin Erich Jobst, Primetals Technologies Austria GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -88,51 +106,154 @@ public static double getWidth(final FBNetworkElement element) {
...
 		if (element instanceof final StructManipulator structManipulator && structManipulator.getDataType() != null) {
-			return structManipulator.getDataType().getName().length();
+			return structManipulator.getDataType().getName().length() * AVARAGE_CHAR_WIDTH + WIDTH_ADJUST_STRUCT_NAME;
 		}
 		return 0;
 	}
 
-	protected static double getHeightAdjust(final FBNetworkElement element) {
+	private static double getMaxEventPinWidth(final FBNetworkElement element, final int minPinLabelSize,
+			final int maxPinLabelSize, final boolean isInput) {
```
azoitl: if you directly provide here the stream you don' need the FBNetworkElement and reduce the complexity of the method a lot.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -88,51 +106,154 @@ public static double getWidth(final FBNetworkElement element) {
...
+
+		int topInputLabelSize = pinStream.filter(IInterfaceElement::isVisible).map(IInterfaceElement::getName)
+				.mapToInt(String::length).max().orElse(0);
+		topInputLabelSize = Math.clamp(topInputLabelSize, minPinLabelSize, maxPinLabelSize);
+
+		return topInputLabelSize * AVARAGE_CHAR_WIDTH + WIDTH_ADJUST_INTERFACE;
+	}
+
+	private static double getMaxPinWidth(final FBNetworkElement element, final int minPinLabelSize,
+			final int maxPinLabelSize, final boolean isInput) {
```
azoitl, line 151: you could directly provide here a stream with the pins to check, this can be done with stream.concat, this make the method much cleaner.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -88,51 +106,154 @@ public static double getWidth(final FBNetworkElement element) {
...
+					}
+					if (ie instanceof VarDeclaration) {
+						return 2;
+					}
+					return 3;
+				});
+
+		final Optional<IInterfaceElement> largestInterface = element.getInterface().getAllInterfaceElements().stream()
+				.filter(Predicate.not(Event.class::isInstance)).filter(IInterfaceElement::isVisible).filter(inputFilter)
+				.sorted(maxPinLengthComparator).findFirst();
```
azoitl: why do you need to sort according to the pin type. Isn't it always the longest? the layout how it is drawn is that the each side is as wide as the one with longest name.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
sebHollersbacher: We also have to consider the symbol of the pin, which are longer for VarInOut and Adapter.
So 2 pins with the same name length can be different sizes

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
sebHollersbacher: Here we need the element anyway as we have a special check for the hidden-pin marker inside `getHiddenWidthAdjust`

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -1,5 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2023 Martin Erich Jobst
+ * Copyright (c) 2023 Martin Erich Jobst,
```
azoitl: Please change to:
` * Copyright (c) 2023, 2025 Martin Erich Jobst, Primetals Technologies Austria GmbH

azoitl (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -69,14 +88,29 @@ public final class FBShapeHelper {
...
-		final double nameWidth = (element.getName().length() * AVARAGE_CHAR_WIDTH) + WIDTH_ADJUST_NAME;
-		final double typeNameWidth = (getTypeNameCharacters(element) * AVARAGE_CHAR_WIDTH) + WIDTH_ADJUST_TYPE_NAME;
-		final double structNameWidth = (getStructNameCharacters(element) * AVARAGE_CHAR_WIDTH)
-				+ WIDTH_ADJUST_STRUCT_NAME;
-		final int inputCharacters = getInterfaceCharacters(element, IInterfaceElement::isIsInput);
-		final int outputCharacters = getInterfaceCharacters(element, Predicate.not(IInterfaceElement::isIsInput));
-		final double interfaceWidth = ((inputCharacters + outputCharacters) * AVARAGE_CHAR_WIDTH)
-				+ WIDTH_ADJUST_INTERFACE;
+		final LibraryElement root = (LibraryElement) EcoreUtil.getRootContainer(element);
+		final IProject project = root.getTypeEntry().getFile().getProject();
```
mx990: Please add an instanceof check in case it is not conatined in a `LibraryElement`. I would suggest use `null` for the project in that case to use the workspace defaults. I would further suggest to extract this into a separate method.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -88,51 +122,136 @@ public static double getWidth(final FBNetworkElement element) {
...
-	}
 
-	private static int getInterfaceLines(final FBNetworkElement element, final Predicate<IInterfaceElement> filter) {
-		return (int) element.getInterface().getAllInterfaceElements().stream().filter(IInterfaceElement::isVisible)
-				.filter(filter).count();
+		final int eventLines = Math.max(eventInputLines, eventOutputLines);
+		final int inOutLines = Math.max(inOutInputLines, inOutOutputLines);
+		final int dataLines = Math.max(dataInputLines, dataOutputLines);
+		final int adapterLines = Math.max(adapterInputLines, adapterOutputLines);
+		final int errorLines = Math.max(errorInputLines, errorOutputLines);
```
mx990: Would it not be easier to iterate over the lists individually instead of first getting all interface elements and then splitting them up? I would also suggest to split each set of lists (e.g., event inputs and outouts) in a separate method and compute the lines per block there.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
mx990: I would suggest to first calculate the full width of the pin, including the symbol, and then get the maximum. That way, you do not need to do any complex sorting.

@azoitl Is it not a problem if the actual width depends on the kind of pin, when considering achieving a consistent and predictable layout?

mx990 (changes requested):

azoitl:
@mx990 width is less critical for predicatable layouts. Also currently using same width symbols for data pins, var in out pins and adapters would not look good.

mx990 (approved):

PR #1490, opened 2025-07-29
Ignore types from variables in typed instances in organize imports

mx990:
The imported namespaces for types from variables in typed instances should be ignored, since they are declared in the type.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/VarDeclarationSearchSupport.java
+++ b/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/VarDeclarationSearchSupport.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import java.util.stream.Stream;
 
 import org.eclipse.emf.ecore.EObject;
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/VarDeclarationSearchSupport.java
```diff
@@ -63,6 +66,11 @@ protected IParseResult prepareInitialValue() {
 		return parseResult;
 	}
 
+	public static boolean isContainedInTypedInstance(final IInterfaceElement element) {
```
azoitl: Do we need to recursively go up the hierarchy, e.g., if we are in an untyped subapp in an typed subapp instance? In oder places we need have that

mx990: Good point. I also changed the organize imports command to skip networks of typed subapps and CFB instances altogether.

azoitl (approved):

PR #1502, opened 2025-08-06
It is not allowed anymore to use reserved attributes like "Visible".

PStmmr:
If so, a Pop- Up Error Message will appear.
Somehow already exported incorrect attribute names get an error marker.


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -410,6 +410,9 @@ private void readNameAttribute(final INamedElement namedElement) throws TypeImpo
 		if (null == name) {
 			throw new TypeImportException(Messages.Import_ERROR_NameNotDefined);
 		}
+		if (null != InternalAttributeDeclarations.getInternalAttributeByName(name)) {
+			System.out.println("error" + " -> " + name);
```
mx990: Please do not commit any debug messages.

plugins/org.eclipse.fordiac.ide.gef/plugin.properties
```diff
@@ -94,3 +94,4 @@ HandlerPreferenceSafeError=Could not save preferences
 context.graphicalEditing.description=4diac IDE Graphical Editing Context
 context.graphicalEditing.name=Graphical Editing
 
+ReservedKeyWordErrorMessage=Invalid Attribute Name: {0} is a reserved keyword and cannot be used here.
```
mx990: Please change the message, since internal attributes are not keywords.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/NamedElementAnnotations.java
```diff
@@ -114,6 +117,19 @@ public static boolean validateName(final INamedElement element, final Diagnostic
...
+						.findFirst();
+				System.out.println(invalidAttr);
+				if (invalidAttr.isPresent() && (diagnostics != null)) {
+					diagnostics.add(new BasicDiagnostic(Diagnostic.ERROR, LibraryElementValidator.DIAGNOSTIC_SOURCE,
+							LibraryElementValidator.INAMED_ELEMENT__VALIDATE_NAME,
+							MessageFormat.format(Messages.Attribute_InternalNameNotAllowed,
+									invalidAttr.get().getName()),
+							FordiacMarkerHelper.getDiagnosticData(element,
+									LibraryElementPackage.Literals.INAMED_ELEMENT__NAME)));
+				}
```
mx990: First of all, please do not add validations for a specific subclass to a general invariant. You can override the invariant for the subclass in the model, create a separate annotation, and put your code there.

Furthermore, it is the attribute and not the interface element that is the problem, so the diagnostic should be added on each attribute instead. I would then also recommend to add the additional validation to the existing override for `Attribute`[^1]. You could also filter for an `IInterfaceElement` container there, but what about the other possible containers?

Additionally, your added `else if` breaks validation of invalid identifiers for `IInterfaceElement`s, since the `else` below is skipped for them now.

Finally, I do not understand why you are adding this validation in the first place. This will also include internal attributes added to the model programatically. We obviously need to allow this, otherwise what is the point of having them in the first place?

[^1]: see https://github.com/eclipse-4diac/4diac-ide/blob/71f0e033d3839fdc3be1abd40c2bb40b3a693865/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/AttributeAnnotations.java#L61

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/AttributeSection.java
```diff
@@ -247,6 +255,30 @@ public void removeEntry(final Object entry, final CompoundCommand cmd) {
...
+	}
+
+	private final IDataValidator attributeNameValidator = new IDataValidator() {
+		@Override
+		public boolean validate(final int columnIndex, final int rowIndex, final Object newValue) {
+			if (!(newValue instanceof final String name)) {
+				return true;
+			}
+
+			if (getType() instanceof VarDeclaration && isInternalAttribute(name)) {
```
mx990: Why do you restrict this to `VarDeclaration`? What about the other possible containers?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/NamedElementAnnotations.java
PStmmr: Thanks for your feedback.
I will add  the validation to the existing attribute override. You're right, that's much better. Regarding "but what about the other possible containers," are there other containers when we're talking about pins? Regarding your question about why validation is being added at all, it's not about restricting functionality that already works, but rather about detecting and eliminating any incorrectly named attributes that might mysteriously exist in the XML. I've noticed that the filter, as it exists now, actually restricts functionality. Could it work if, in addition to the IInterfaceElement check, I also checked whether the typename is not empty?

mx990: > Regarding "but what about the other possible containers," are there other containers when we're talking about pins?

Not when talking about pins, but there could also be malformed internal attributes in other `ConfigurableObject`s, such as FB types or instances. Is there a reason not to include them?

> Regarding your question about why validation is being added at all, it's not about restricting functionality that already works, but rather about detecting and eliminating any incorrectly named attributes that might mysteriously exist in the XML. I've noticed that the filter, as it exists now, actually restricts functionality. Could it work if, in addition to the IInterfaceElement check, I also checked whether the typename is not empty?

Thank you for clarifying. If I understand you correctly, you are looking specifically for attributes with the same name as an internal attribute but with a separate type and thus without the internal attribute declaration? In that case, I would suggest to simply check whether the attribute declaration is `null`. I would further suggest to add a set of internal attribute names in `InternalAttributeDeclarations` for faster lookups.

PStmmr: >Not when talking about pins, but there could also be malformed internal attributes in other `ConfigurableObject`s, such as FB types or instances. Is there a reason not to include them?

All I could gather from the issue given was that the problem must be fixed at pin level, so I didn't look into the matter any further.
﻿
> If I understand you correctly, you are looking specifically for attributes with the same name as an internal attribute but with a separate type and thus without the internal attribute declaration?

Yes and no. It's "possible" that a user-created attribute with such a forbidden name exists in an XML file for a pin. Unlike, for example, if I set a pin to invisible, this attribute then has a type entry in the code.
Is there a way to differentiate between these?


image

In this example, the first is an attribute that is automatically created when a pin's visibility is set to false. Below that is a pin attribute that is created via the property sheet, and whose name should not be allowed if it exists in the XML.

mx990: > All I could gather from the issue given was that the problem must be fixed at pin level, so I didn't look into the matter any further.

In that case, please do not restrict it to pins and just check for all attributes.

> In this example, the first is an attribute that is automatically created when a pin's visibility is set to false. Below that is a pin attribute that is created via the property sheet, and whose name should not be allowed if it exists in the XML.

That is what I meant. Normally, if an attribute has a `Type` XML attribute, it is considered a "free" attribute and not associated with any attribute declaration, even if one exists. However, I just noticed that this is currently not the case when it has the name of an internal attribute. In that case, it is *always* associated with the internal attribute declaration and the `Type` XML attribute is simply ignored:
https://github.com/eclipse-4diac/4diac-ide/blob/00fb19db743aef49274fdfbcb4a4f4b3e6f6afc9/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java#L428-L456
This means there is unfortunately no way to differentiate them in the model. It would thus have to be checked in the importer. You could add a diagnostic during import similar to:
https://github.com/eclipse-4diac/4diac-ide/blob/00fb19db743aef49274fdfbcb4a4f4b3e6f6afc9/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/SystemImporter.java#L290-L291

Nevertheless, it may still be prudent to also have a separate check in the model if such an attribute is somehow added there. That should then also work with checking for a `null` attribute declaration as I suggested before.

PStmmr: I applied your suggestion as I understood it (updated commit), and the check worked, but no error marker appeared. Can you tell me what's wrong with it, or what you meant by it? I think it's easier to communicate this way.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/Messages.java
```diff
@@ -117,6 +117,8 @@ public final class Messages extends NLS {
 
 	public static String HandlerPreferenceSafeError;
 
+	public static String ReservedKeyWordErrorMessage;
```
mx990: Please also rename the message key.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,17 +421,39 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
+		}
+		return false;
+	}
+
 	protected void parseGenericAttributeNode(final ConfigurableObject confObject)
 			throws XMLStreamException, TypeImportException {
+		final boolean hasExplicitType = hasXmlAttribute(LibraryElementTags.TYPE_ATTRIBUTE);
+		final String xmlTypeName = hasExplicitType
+				? getReader().getAttributeValue("", LibraryElementTags.TYPE_ATTRIBUTE)
+				: null;
```
mx990: Why do you have to check if the attribute exists first and not just move `typeName` up here from line 458?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,17 +421,39 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
-
 		final AttributeDeclaration internalAttributeDecl = InternalAttributeDeclarations
 				.getInternalAttributeByName(attribute.getName());
 		if (internalAttributeDecl != null) {
 			// Internal Attributes
 			attribute.setAttributeDeclaration(internalAttributeDecl);
 			attribute.setType(internalAttributeDecl.getType());
+			if (hasExplicitType) {
+				errors.add(new TypeImportDiagnostic("Internal attribute name specified with explicit Type", //$NON-NLS-1$
+						attribute.getName() + " (Type=" + xmlTypeName + ")", getLineNumber())); //$NON-NLS-1$ //$NON-NLS-2$
```
mx990: Please externalize the message and also include the FQN of the `confObject`.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/AttributeAnnotations.java
```diff
@@ -71,4 +71,4 @@ public static boolean validateName(final Attribute attribute, final DiagnosticCh
 	private AttributeAnnotations() {
 		throw new UnsupportedOperationException();
 	}
-}
+}
```
mx990, line 74: Please do not commit pure whitespace changes.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/messages.properties
```diff
@@ -123,3 +123,64 @@ VarDeclarationAnnotations_ValueOverriddenBySubAppInput=Specified value is overri
...
+AttributeTarget_UntypedSubAppEvent_tooltip=Event-Pins of Untyped SubApps
+AttributeTarget_UntypedSubAppData_tooltip=Data-Pins of Untyped SubApps
+AttributeTarget_Systems_tooltip=Systems
+AttributeTarget_Applications_tooltip=Applications
+AttributeTarget_GlobalConstants_tooltip=Global Constants
+AttributeTarget_Connections_tooltip=Connections
+AttributeTarget_Comments_tooltip=Comments
+AttributeTarget_Groups_tooltip=Groups
+AttributeTarget_Links_tooltip=Links
+AttributeTarget_ServiceSequences_tooltip=Service Sequences
```
mx990, line 187: This seems to be unrelated?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/NamedElementAnnotations.java
mx990: The diagnostics did get lost in the resource. I have created a PR https://github.com/eclipse-4diac/4diac-ide/pull/1525 with a fix.

mx990 (changes requested):
I believe this should targeting freeze?

PStmmr:
> I believe this should targeting freeze?

I believed it should target develop since the issue is not marked as a bug.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/messages.properties
PStmmr: There was a merge conflict,so i had to push that to keep the branches up to date

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/NamedElementAnnotations.java
PStmmr: > The diagnostics did get lost in the resource. I have created a PR #1525 with a fix.

Amazing Thanks!!

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,21 +422,44 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
+			return false;
+		}
+		for (int i = 0; i < reader.getAttributeCount(); i++) {
+			if (localName.equals(reader.getAttributeLocalName(i))) {
+				final String v = reader.getAttributeValue(i);
+				return v != null && !v.isBlank();
+			}
+		}
+		return false;
+	}
```
mx990: Please remove this unused method.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,21 +422,44 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
+
 	protected void parseGenericAttributeNode(final ConfigurableObject confObject)
 			throws XMLStreamException, TypeImportException {
+		final String xmlTypeName = getAttributeValue(LibraryElementTags.TYPE_ATTRIBUTE);
```
mx990: Please just name this `typeName`:
```suggestion
		final String typeName = getAttributeValue(LibraryElementTags.TYPE_ATTRIBUTE);
```
and remove line 460.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,21 +422,44 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
+		final boolean hasExplicitType = (xmlTypeName != null) && !xmlTypeName.isBlank();
+
 		final Attribute attribute = LibraryElementFactory.eINSTANCE.createAttribute();
 		readNameCommentAttributes(attribute);
-
 		final AttributeDeclaration internalAttributeDecl = InternalAttributeDeclarations
 				.getInternalAttributeByName(attribute.getName());
 		if (internalAttributeDecl != null) {
-			// Internal Attributes
+			if (hasExplicitType) {
```
mx990: Please just inline the variable here.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/CommonElementImporter.java
```diff
@@ -420,21 +422,44 @@ protected void readCommentAttribute(final INamedElement namedElement) {
...
+				final String fqn = confObject instanceof final LibraryElement libElelemt ? libElelemt.getQualifiedName()
+						: ""; //$NON-NLS-1$
```
mx990: I would suggest using `FordiacMarkerHelper.getLocation(EObject)` instead.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/messages.properties
```diff
@@ -185,4 +186,4 @@ AttributeTarget_Groups_tooltip=Groups
 AttributeTarget_Links_tooltip=Links
 AttributeTarget_ServiceSequences_tooltip=Service Sequences
 
-HidePinCommand_PinCannotBeHidden_ConnectedInside=You cannot hide this pin: it is connected inside the SubApp
+HidePinCommand_PinCannotBeHidden_ConnectedInside=You cannot hide this pin: it is connected inside the SubApp
```
mx990, line 189: It seems there is still a leftover change here, maybe due to whitespace?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/Messages.java
```diff
@@ -117,6 +117,8 @@ public final class Messages extends NLS {
 
 	public static String HandlerPreferenceSafeError;
 
+	public static String AttributeSection_ReservedKeyword;
```
mx990: Please also rename this message key, similar to `plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/Messages.java` below.

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #1508, opened 2025-08-11
DeleteInternalFBChange now based on ModelEdit

pazi146:
This mitigates issues when multiple instances are deleted at once


mx990 (changes requested):
This will not work in case there are name conflicts. I would suggest to use the new model edit infrastructure from https://github.com/eclipse-4diac/4diac-ide/pull/1505 instead. You would need to turn the change into a model edit[^1] and then use `ModelEditChange.fromModelEdits(...)` to pack them into a composite change.

[^1]: similar to https://github.com/eclipse-4diac/4diac-ide/pull/1505/files#diff-f58792416193f69f5a5512c1f55363e1d11ebabf2c522cc6b0304cc07d304da3

pazi146:
Yes, that would work - but how do I deal with the fact that `DeleteInternalFBChange` is a `ConfigurableChange`?

Would I need to split it into two (empty change and configurable ModelEdit)?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/delete/DeleteTypeRefactoringParticipant.java
```diff
@@ -64,7 +72,7 @@ public Change createChange(final IProgressMonitor pm) throws CoreException, Oper
 				return new SafeStructDeletionChange(struct);
 			}
 			if (typeEntry.getType() instanceof final FBType type) {
```
oberlehner: please check instance of FBTypeEntry or DataTypeEntry, since calling typeEntry.getType() is an expensive operation and should be avoided

pazi146: changed

mx990 (approved):

PR #1509, opened 2025-08-11
Avoid editor reload on type save.

azoitl:


plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/editors/AutomationSystemEditor.java
```diff
@@ -248,7 +248,9 @@ protected void execute(final IProgressMonitor monitor)
 				}
 			};
 			try {
+				typeEntryAdapter.setBlockUpdates(true);
 				operation.run(monitor);
+				typeEntryAdapter.setBlockUpdates(false);
```
mx990: Please move this into a `finally` block in case there is an exception during saving.

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/AbstractTypeEditor.java
```diff
@@ -244,7 +244,9 @@ protected void execute(final IProgressMonitor monitor)
 			}
 		};
 		try {
+			typeEntryAdapter.setBlockUpdates(true);
 			operation.run(monitor);
+			typeEntryAdapter.setBlockUpdates(false);
```
mx990: same here

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -89,6 +90,10 @@ protected void reloadEditorType() {
 	public void notifyChanged(final Notification notification) {
 		super.notifyChanged(notification);
 
+		if (blockUpdates) {
```
oberlehner, line 93: Wouldn't it be better to check if the `notification.getNotifier()`  equals the input of `ITypeEntryEditor editor` ? 
In my understanding, this would check if someone notifies itself.

Then we can avoid the boolean flag.

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/editors/AutomationSystemEditor.java
azoitl: done

plugins/org.eclipse.fordiac.ide.typeeditor/src/org/eclipse/fordiac/ide/typeeditor/AbstractTypeEditor.java
azoitl: done

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: The notifier is always a type entry. Therefore this is not working. I also tried other fields of the notification but the only working solution that I could come up with is to block the typeentryadpater.

PR #1524, opened 2025-08-13
VarConfig Error when values

PStmmr:
VarConfig Bug, Error Name Field

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
@@ -24,7 +24,6 @@ import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
-import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.application.Messages;
 import org.eclipse.fordiac.ide.application.editparts.FBNetworkEditPart;
 import org.eclipse.fordiac.ide.application.editparts.SubAppForFBNetworkEditPart;
@@ -34,6 +33,7 @@ import org.eclipse.fordiac.ide.gef.nat.VarDeclarationConfigLabelAccumulator;
 import org.eclipse.fordiac.ide.gef.nat.VarDeclarationDataLayer;
 import org.eclipse.fordiac.ide.gef.nat.VarDeclarationTableColumn;
 import org.eclipse.fordiac.ide.gef.properties.AbstractSection;
+import org.eclipse.fordiac.ide.model.helpers.InterfaceListCopier;
 import org.eclipse.fordiac.ide.model.libraryElement.Application;
 import org.eclipse.fordiac.ide.model.libraryElement.CFBInstance;
 import org.eclipse.fordiac.ide.model.libraryElement.FB;
@@ -44,6 +44,7 @@ import org.eclipse.fordiac.ide.model.libraryElement.SubApp;
 import org.eclipse.fordiac.ide.model.libraryElement.SubAppType;
 import org.eclipse.fordiac.ide.model.libraryElement.TypedSubApp;
 import org.eclipse.fordiac.ide.model.libraryElement.UntypedSubApp;
+import org.eclipse.fordiac.ide.model.libraryElement.VarConfigInstance;
 import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
 import org.eclipse.fordiac.ide.ui.widget.ChangeableListDataProvider;
 import org.eclipse.fordiac.ide.ui.widget.CheckBoxConfigurationNebula;
@@ -276,14 +277,9 @@ public class VarConfigurationSection extends AbstractSection {
 		if (existing != null) {
 			targetVD = existing;
 		} else {
-			targetVD = EcoreUtil.copy(vd);
-			targetVD.setName(relativeName);
-			if (vd.getValue() != null) {
-				targetVD.setValue(EcoreUtil.copy(vd.getValue()));
-			}
-			targetVD.setComment(vd.getComment());
-			targetVD.getAttributes().clear();
-			rootTSA.getVarConfigParams().add(targetVD);
+			final VarConfigInstance vcI = InterfaceListCopier.copyVarConfigInstance(vd, relativeName);
+			rootTSA.getVarConfigParams().add(vcI);
+			targetVD = vcI;
 		}
 		result.put(qualifiedName, targetVD);
 		copiedMap.put(targetVD, Boolean.TRUE);
--- a/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
+++ b/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
@@ -2272,7 +2272,7 @@
     <eStructuralFeatures xsi:type="ecore:EReference" name="subAppNetwork" eType="#//FBNetwork"
         transient="true" derived="true" containment="true"/>
     <eStructuralFeatures xsi:type="ecore:EReference" name="varConfigParams" upperBound="-1"
-        eType="#//VarDeclaration" containment="true"/>
+        eType="#//VarConfigInstance" containment="true"/>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="UntypedSubApp" eSuperTypes="#//SubApp">
     <eOperations name="getType" lowerBound="1" eType="#//SubAppType">
@@ -2518,4 +2518,21 @@
   <eClassifiers xsi:type="ecore:EDataType" name="TypeEntry" instanceClassName="org.eclipse.fordiac.ide.model.typelibrary.TypeEntry"/>
   <eClassifiers xsi:type="ecore:EDataType" name="TypeLibrary" instanceClassName="org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary"/>
   <eClassifiers xsi:type="ecore:EDataType" name="VarDeclList" instanceTypeName="java.util.List&lt;org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration>"/>
+  <eClassifiers xsi:type="ecore:EClass" name="VarConfigInstance" eSuperTypes="#//VarDeclaration">
+    <eOperations name="validateName" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EBoolean">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/Ecore">
+        <details key="invariant" value="true"/>
+      </eAnnotations>
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="return org.eclipse.fordiac.ide.model.libraryElement.impl.VarConfigInstanceAnnotations.validateName(this, diagnostics, context);"/>
+      </eAnnotations>
+      <eParameters name="diagnostics" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EDiagnosticChain"/>
+      <eParameters name="context">
+        <eGenericType eClassifier="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EMap">
+          <eTypeArguments eClassifier="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EJavaObject"/>
+          <eTypeArguments eClassifier="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EJavaObject"/>
+        </eGenericType>
+      </eParameters>
+    </eOperations>
+  </eClassifiers>
 </ecore:EPackage>
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
@@ -35,6 +35,7 @@ public final class IdentifierVerifier {
 
 	private static final String IDENTIFIER_REGEX = "[_A-Za-z][_A-Za-z0-9]*+"; //$NON-NLS-1$
 	private static final Pattern IDENTIFIER_PATTERN = Pattern.compile(IDENTIFIER_REGEX);
+	private static final String QUALIFIED_NAME_DELIMITER = "\\."; //$NON-NLS-1$
 
 	private IdentifierVerifier() {
 		throw new UnsupportedOperationException();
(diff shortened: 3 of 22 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/NamedElementAnnotations.java
```diff
@@ -115,7 +117,9 @@ public static boolean validateName(final INamedElement element, final Diagnostic
 									LibraryElementPackage.Literals.INAMED_ELEMENT__NAME)));
 				}
 			} else {
-				final Optional<String> errorMessage = IdentifierVerifier.verifyIdentifier(element.getName(), element);
+				final Optional<String> errorMessage = (element instanceof VarConfigInstance)
+						? VarConfigVerifier.verifyIdentifier(element.getName(), element)
+						: IdentifierVerifier.verifyIdentifier(element.getName(), element);
```
mx990: Please do not add validations for a specific subclass to a general invariant. You can override the invariant for the subclass in the model, create a separate annotation, and put your code there.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/VarConfigVerifier.java
```diff
@@ -0,0 +1,64 @@
...
+	 * Checks if the identifier is valid, and returns an Optional error message.
+	 *
+	 * @param identifier the identifier
+	 * @param context    The context in which the identifier has to be valid (may be
+	 *                   {@code null})
+	 *
+	 * @return Empty Optional if the identifier is valid, otherwise the error
+	 *         message is contained in the Optional
+	 */
+	public static Optional<String> verifyIdentifier(final String identifier, final Object context) {
```
mx990: Please have a look at `IdentifierVerifier.verifyPackageName(String)`, which does almost exactly what you need. I would also suggest moving this into `IdentifierVerifier`. Furthermore, please add unit tests for your new method, see also `IdentifierVerifierTest`.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/VarConfigVerifier.java
```diff
@@ -0,0 +1,64 @@
...
+		if (identifier == null || (!IDENTIFIER_PATTERN.matcher(identifier).matches() && !identifier.contains("."))) { //$NON-NLS-1$
```
mx990: This means the pattern does not need to match so long as the identifier contains a dot somewhere, so *any* characters would be allowed then. You would instead have to add the dot to the regex.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/VarConfigVerifier.java
```diff
@@ -0,0 +1,64 @@
...
+			return Optional.of(MessageFormat.format(Messages.IdentifierVerifier_NameNotAValidIdentifier, identifier));
+		}
+		if (identifier.contains("__")) { //$NON-NLS-1$
+			return Optional.of(MessageFormat.format(Messages.IdentifierVerifier_NameConsecutiveUnderscore, identifier));
+		}
+		if (identifier.endsWith("_")) { //$NON-NLS-1$
+			return Optional.of(MessageFormat.format(Messages.IdentifierVerifier_NameTrailingUnderscore, identifier));
+		}
+		if (FordiacKeywords.isReservedKeyword(identifier, context)) {
```
mx990: This would never return true for a dot-separated identifier. You would have to split the identifier and check each segement individually, see also `IdentifierVerifier.verifyPackageName(String)`.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -283,7 +285,18 @@ private void addAndCopyElement(final VarDeclaration vd, final String currentPref
...
+			final VarConfigInstance vci = LibraryElementFactory.eINSTANCE.createVarConfigInstance();
+			vci.setName(relativeName);
+			vci.setIsInput(vd.isIsInput());
+			vci.setVisible(vd.isVisible());
+			vci.setType(vd.getType());
+			if (vd.getValue() != null) {
+				vci.setValue(EcoreUtil.copy(vd.getValue()));
+			}
+			vci.setComment(vd.getComment());
+			vci.getAttributes().clear();
```
mx990: Please move this and the similar code in `CommonElementImporter.copyVarDeclFromType(...)` into a separate method in `InterfaceListCopier`. Also, if you freshly created the copy, there should be no need to clear the attributes.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -2518,4 +2518,9 @@
   <eClassifiers xsi:type="ecore:EDataType" name="TypeEntry" instanceClassName="org.eclipse.fordiac.ide.model.typelibrary.TypeEntry"/>
   <eClassifiers xsi:type="ecore:EDataType" name="TypeLibrary" instanceClassName="org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary"/>
   <eClassifiers xsi:type="ecore:EDataType" name="VarDeclList" instanceTypeName="java.util.List&lt;org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration>"/>
+  <eClassifiers xsi:type="ecore:EClass" name="VarConfigInstance" eSuperTypes="#//VarDeclaration">
+    <eAnnotations source="http://www.eclipse.org/emf/2002/Ecore">
+      <details key="constraints" value="ValidQualifiedName"/>
+    </eAnnotations>
```
mx990: In order to override the generic name validation for `VarConfigInstance`, please do the following:
1. create a new class `org.eclipse.fordiac.ide.model.libraryElement.impl.VarConfigInstanceAnnotations`:
    1. with a private constructor, see also https://github.com/eclipse-4diac/4diac-ide/blob/ba029ec4cf474b58079918c680dd9795adc16522/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/AttributeAnnotations.java#L71-L73
    2. with a method `public static boolean validateName(final VarConfigInstance instance, final DiagnosticChain diagnostics, final Map<Object, Object> context)` with your custom implementation,
2. create an overriding invariant by copying https://github.com/eclipse-4diac/4diac-ide/blob/ba029ec4cf474b58079918c680dd9795adc16522/plugins/org.eclipse.fordiac.ide.model/model/lib.ecore#L112-L126 and changing the `body` to point to your annotation class.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/InterfaceListCopier.java
```diff
@@ -185,6 +186,19 @@ public static AdapterDeclaration copyAdapter(final AdapterDeclaration adapter, f
 		return copy;
 	}
 
+	public static VarConfigInstance copyVarConfigInstance(final VarDeclaration src, final String relativeName) {
+		final VarConfigInstance vci = LibraryElementFactory.eINSTANCE.createVarConfigInstance();
+		vci.setName(relativeName);
+		vci.setIsInput(src.isIsInput());
+		vci.setVisible(src.isVisible());
+		vci.setType(src.getType());
```
mx990, line 194: I believe you also need to copy the array size?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -63,6 +65,9 @@ public static Optional<String> verifyIdentifier(final String identifier) {
 	 *         message is contained in the Optional
 	 */
 	public static Optional<String> verifyIdentifier(final String identifier, final Object context) {
+		if (context instanceof VarConfigInstance) {
+			return verifyQualifiedIdentifier(identifier);
+		}
```
mx990: Again, please do *not* add custom handling to generic code.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +92,20 @@ public static Optional<String> verifyPackageName(final String packageName) {
 						.format(Messages.IdentifierVerifier_PackageNameMessage, message, packageName))
 				.findFirst();
 	}
+
+	public static Optional<String> verifyQualifiedIdentifier(final String qualifiedName) {
+		if (qualifiedName == null || qualifiedName.isBlank() || qualifiedName.startsWith(SEPARATION_POINT)
+				|| qualifiedName.endsWith(SEPARATION_POINT) || qualifiedName.contains("..")) { //$NON-NLS-1$
```
mx990: I believe blank identifiers or empty segments are already covered by splitting with limit `-1`?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +92,20 @@ public static Optional<String> verifyPackageName(final String packageName) {
...
+		}
+		return verifyQualifiedByDelimiter(qualifiedName, "\\."); //$NON-NLS-1$
+	}
+
+	private static Optional<String> verifyQualifiedByDelimiter(final String name, final String delimiterRegex) {
+		return Stream.of(name.split(delimiterRegex, -1)).map(seg -> IdentifierVerifier.verifyIdentifier(seg, null))
+				.flatMap(Optional::stream)
+				.map(msg -> MessageFormat.format(Messages.IdentifierVerifier_PackageNameMessage, msg, name))
+				.findFirst();
+	}
```
mx990: Please pass the format string for the message as a parameter, provide an appropriate format string in `verifyQualifiedIdentifier`, and also adapt `verifyPackageName` to use the common implementation.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +92,20 @@ public static Optional<String> verifyPackageName(final String packageName) {
...
+		if (qualifiedName == null || qualifiedName.isBlank() || qualifiedName.startsWith(SEPARATION_POINT)
+				|| qualifiedName.endsWith(SEPARATION_POINT) || qualifiedName.contains("..")) { //$NON-NLS-1$
+			return Optional
+					.of(MessageFormat.format(Messages.IdentifierVerifier_NameNotAValidIdentifier, qualifiedName));
+		}
+		return verifyQualifiedByDelimiter(qualifiedName, "\\."); //$NON-NLS-1$
+	}
+
+	private static Optional<String> verifyQualifiedByDelimiter(final String name, final String delimiterRegex) {
+		return Stream.of(name.split(delimiterRegex, -1)).map(seg -> IdentifierVerifier.verifyIdentifier(seg, null))
```
mx990: I believe this can be simplified to:
```suggestion
		return Stream.of(name.split(delimiterRegex, -1)).map(IdentifierVerifier::verifyIdentifier)
```

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -35,6 +36,7 @@ public final class IdentifierVerifier {
 
 	private static final String IDENTIFIER_REGEX = "[_A-Za-z][_A-Za-z0-9]*+"; //$NON-NLS-1$
 	private static final Pattern IDENTIFIER_PATTERN = Pattern.compile(IDENTIFIER_REGEX);
+	private static final String SEPARATION_POINT = "."; //$NON-NLS-1$
```
mx990: Please rename this to make the meaning more clear, for example, `QUALIFIED_NAME_DELIMITER`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/VarConfigInstanceAnnotations.java
```diff
@@ -0,0 +1,42 @@
...
+		throw new UnsupportedOperationException();
+	}
+
+	public static boolean validateName(final VarConfigInstance instance, final DiagnosticChain diagnostics,
+			final Map<Object, Object> context) {
+		final Optional<String> err = IdentifierVerifier.verifyQualifiedIdentifier(instance.getName());
+		if (err.isPresent()) {
+			if (diagnostics != null) {
+				diagnostics.add(new BasicDiagnostic(Diagnostic.ERROR, LibraryElementValidator.DIAGNOSTIC_SOURCE, 0,
+						err.get(), new Object[] { instance }));
```
mx990: Please provide the error code `LibraryElementValidator.VAR_CONFIG_INSTANCE__VALIDATE_NAME` that was generated with your invariant and use the following to create the diagnostic data:
```java
FordiacMarkerHelper.getDiagnosticData(instance, LibraryElementPackage.Literals.INAMED_ELEMENT__NAME)
```

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +88,23 @@ public static Optional<String> verifyPackageName(final String packageName) {
...
+
+	public static Optional<String> verifyQualifiedIdentifier(final String qualifiedName) {
+		if (qualifiedName == null || qualifiedName.isBlank()) {
+			return Optional
+					.of(MessageFormat.format(Messages.IdentifierVerifier_NameNotAValidIdentifier, qualifiedName));
+		}
+//		return Stream.of(qualifiedName.split(QUALIFIED_NAME_DELIMITER, -1)).map(IdentifierVerifier::verifyIdentifier)
+//				.flatMap(Optional::stream)
+//				.map(msg -> MessageFormat.format(Messages.IdentifierVerifier_PackageNameMessage, msg, qualifiedName))
+//				.findFirst();
```
mx990: Please do not commit commented code.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +88,23 @@ public static Optional<String> verifyPackageName(final String packageName) {
...
+		return verifyByDelimiter(qualifiedName, QUALIFIED_NAME_DELIMITER,
+				Messages.IdentifierVerifier_PackageNameMessage);
```
mx990: Please create a separate message for qualified names.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
```diff
@@ -87,4 +88,23 @@ public static Optional<String> verifyPackageName(final String packageName) {
 						.format(Messages.IdentifierVerifier_PackageNameMessage, message, packageName))
 				.findFirst();
```
mx990: Please also adapt this to use `verifyByDelimiter`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
PStmmr: It seems like I deleted the wrong thing and I need to be more careful with the comments in the future.
I apologize for that.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/VarConfigInstanceAnnotations.java
mx990: Please also use the helper function to create the diagnostic data.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/IdentifierVerifier.java
mx990: Please define a *separate* message and not reuse an existig one.

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #1533, opened 2025-08-14
Fixed FollowConnection to also jump over visible connections

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/FollowConnectionHandler.java
```diff
@@ -220,11 +220,7 @@ public void jumpOverConnections(final IInterfaceElement startPin, final List<IIn
 		}
 		for (final Connection conn : getConnectionList(startPin)) {
 			final IInterfaceElement next = conn.getSource() == startPin ? conn.getDestination() : conn.getSource();
-			if (conn.isVisible()) {
```
azoitl, line 223: I think with this change the condiation in line 209 needs to consider if the pin is a pin in of a typed subapp or CFB. In this cases it could be that the internal network was loaded for showing it in the viewer and then we would jump into the viewer. Or am I missunderstanding something?

azoitl (changes requested):

azoitl (approved):

Issue #1513, opened 2025-08-11
temporary variable type name equal algorithm name or state name


moniwe:
When creating an BFB like this one:
```
<?xml version="1.0" encoding="UTF-8"?>
<FBType Name="dummy" Comment="Basic FB with empty ECC">
	<Identification Standard="61499-2">
	</Identification>
	<VersionInfo Version="1.0" Author="MWR23" Date="2025-08-01">
	</VersionInfo>
	<CompilerInfo>
	</CompilerInfo>
	<InterfaceList>
		<EventInputs>
			<Event Name="REQ" Type="Event">
				<With Var="Parameters"/>
				<With Var="MessageID"/>
				<With Var="Revision"/>
			</Event>
		</EventInputs>
		<EventOutputs>
			<Event Name="CNF" Type="Event">
				<With Var="Status"/>
				<With Var="Data"/>
			</Event>
		</EventOutputs>
		<InputVars>
			<VarDeclaration Name="MessageID" Type="UINT"/>
			<VarDeclaration Name="Revision" Type="UINT"/>
			<VarDeclaration Name="Parameters" Type="ANY" Comment="Expects integrator MID STRUCTs"/>
		</InputVars>
		<OutputVars>
			<VarDeclaration Name="Status" Type="STRING"/>
			<VarDeclaration Name="Data" Type="STRING"/>
		</OutputVars>
	</InterfaceList>
	<BasicFB>
		<ECC>
			<ECState Name="START" Comment="Initial State" x="550" y="425">
			</ECState>
			<ECState Name="MID0012" x="1520" y="1200">
				<ECAction Algorithm="MID0012" Output="CNF"/>
			</ECState>
			<ECTransition Source="MID0012" Destination="START" Condition="1" Comment="" x="1073.33" y="1113.33"/>
			<ECTransition Source="START" Destination="MID0012" Condition="REQ[MessageID = 12]" Comment="" x="1486.67" y="760"/>
		</ECC>
		<Algorithm Name="reset">
			<ST><![CDATA[ALGORITHM reset
Status := '';
Data := '';
END_ALGORITHM]]></ST>
		</Algorithm>
		<Method Name="numberFormatter" Type="STRING">
			<ST><![CDATA[

METHOD numberFormatter : STRING
VAR_INPUT
	toFormat : UINT;
	name : STRING;
	maxNum : UINT;
	maxDigit : UINT;
END_VAR
VAR_TEMP
	result : STRING;
END_VAR
[...]

PR #1541, opened 2025-08-15
Restrict ST variables and type declarations to data types

mx990:
This resolves a long-standing problem in the ST parser that allowed unrelated local declarations to shadow data types for variables, despite not being eligible for them. It restricts ST variables and type declarations to only data types. This not only fixes the shadowing problem but also greatly reduces the search space when resolving data types in the ST parser, thus improving performance.

Fixes https://github.com/eclipse-4diac/4diac-ide/issues/1513


MartinMelikMerkumians:
As a suggestion to naming, in 61331-3 the term program refers to a type of POU, which essentially acts as a singleton FB.
Could naming this ProgramType lead to confusion or conflicts in the future?
I am thinking on the 61131 mode resource prototype we had as a master thesis prototype, which used the resource as kind of 61131 program

mx990:
That is a good point. I am happy to change the name if anyone has a better suggestion?

MartinMelikMerkumians:
That's a though one. Looking at the original problem and some of your code changes, this change marks FB declarations and type declarations with a super type, so you can filter them for context in variable declarations, right?

So they are some kind of `DeclarableTypes`? (No, that's not my suggestion 😄, just to get get the meaning first right)

mx990:
Yes, that is basically it. I could only come up with `Classifier`, which has a similar meaning in EMF but should (hopefully) be unbiased in terms of IEC 61131-3 or 61499.

We could also simply restrict ST variable declarations to just data types in the model, since we currently do not support FB types in variables anyhow. That would remove the need for a common supertype and thus the naming issue :wink:.

azoitl:
I have a slight tendency for only allowing data types. But not sure how future proof this will be.

MartinMelikMerkumians:
I like Classifier but I could also live with only variables (at least for now 😜)

mx990:
I have changed it to allow only data types for now. We can always extend it later, should the need ever arise.

azoitl (approved):

MartinMelikMerkumiansElectrify:
LGTM

MartinMelikMerkumians:
LGTM

MartinMelikMerkumians (approved):

PR #1548, opened 2025-08-18
Fixed issue where jumping over Struct with dead-end causes Stackoverflow

sebHollersbacher:


azoitl (changes requested):
for pins where i got a stack overflow before I'm now not jumping at all. I don't think this intended or?

sebHollersbacher:
@azoitl  Yes, this was not the case, but is the expected behaviour.
It should work now

azoitl (approved):

PR #1554, opened 2025-08-19
Changed FollowConnection to also consider difference between in/output

sebHollersbacher:
e.g when going right, only inputs should be found (except dead-ends)


azoitl:
While for normal situation this works it stops to early at struct muxers. E.g., If I press ctrl+left If I have a struct mux with connection it stops at the structmux input and not as it should at the connected output of the struct muxer. I expect the same will happen for demuxers.

azoitl (approved):

mx990:
I was unable to jump right from `DINT2DINT.OUT` or `STRUCT_MUX.VAR1` in the following example:
image

sebHollersbacher:
@mx990 
Should be fixed now

mx990 (approved):

oberlehner (approved):

azoitl (approved):

PR #1559, opened 2025-08-20
VarConfigs of multiple identical instances are now displayed.

PStmmr:


oberlehner (approved):

azoitl (changes requested):
It can not be that two var configs have the same display name. So having a map with one display name to several var declarations is wrong.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -174,14 +172,16 @@ private void collectVarConfigsRecursive(final INamedElement type, final Map<Stri
 			traverseFBNetwork(utsa.getSubAppNetwork().getNetworkElements(), result, visited, currentPrefix);
 			break;
 		case final SubAppType sat:
+			if (!visited.add(type)) {
```
azoitl: what will happen if I have two instances of the same type in my network?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -287,27 +287,67 @@ private void addAndCopyElement(final VarDeclaration vd, final String currentPref
...
+				continue;
+			}
 		}
 		return null;
 	}
 
 	private static String getRelativeName(final String qualifiedName, final String rootName) {
-		if (qualifiedName.startsWith(rootName + separationPoint)) {
-			return qualifiedName.substring(rootName.length() + 1);
+		final String[] parts = qualifiedName.split(java.util.regex.Pattern.quote(separationPoint));
```
azoitl, line 329: why do you need to build up this name when you allready have it in the `currentPrefix` variable at the place where you call this method?

azoitl (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
PStmmr: Since they have a different name, nothing bad happens, which is good.
image of 3 instances of same subapptype

azoitl: Then why is it here? Can it be removed?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
PStmmr: Since the relative name must never contain the current subapp, which is contained in the currentprefix, the name must be adjusted accordingly. If the fully qualified name for sub1 were sub1.block.pin, then the relative name would be block.pin (for the XML), since the subapp name is already entered.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
PStmmr: You are right.. this whole case is dead logic because the property sheet does not exist for the type.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
azoitl: wouldn't it then not be smarter to have the currentprefix not have that otherwise you do twice the work.

PStmmr: After mutual agreement, this problem has been resolved until further notice.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/VarConfigurationSection.java
```diff
@@ -200,11 +194,12 @@ private void traverseFBNetwork(final Iterable<FBNetworkElement> elements, final
 			final String prefix = currentPrefix + fbne.getName() + separationPoint;
 			final boolean shouldCopy = !isPartOfEditedStructure(getType(), fbne);
 			addPossibleVarConfigs(fbne, result, prefix, shouldCopy);
-			if (fbne.getType() != null) {
-				collectVarConfigsRecursive(fbne.getType(), result, visited, prefix);
+
```
oberlehner, line 197: same here

oberlehner (approved):

azoitl (approved):

PR #1560, opened 2025-08-20
Convert DeleteTypeRefactoringParticipant Changes into ModelEdits

pazi146:


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ModelEditChange.java
```diff
@@ -36,6 +36,8 @@ public final class ModelEditChange extends AbstractCommandChange<LibraryElement>
 	private ModelEditChange(final List<ModelEdit<?>> edits) {
 		super(edits.getFirst().getLibraryElementURI(), LibraryElement.class);
 		this.edits = edits;
+		// disable if none of the ModelEdits are enabled
+		super.setEnabled(edits.stream().anyMatch(ModelEdit::isEnabled));
```
mx990: This should not be necessary. Why did you add this?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ModelEditChange.java
pazi146: it is necessary if all the `ModelEdit`s are disabled
Error cause:
- `ModelEditChange.isEnabled()` is true, but all contained `ModelEdit`s are disabled
- `AbstractCommandChange`: `perform()` -> `doPerform()` -> `performCommand()` -> `createCommand()` creates an _empty_ CompoundCommand (as disabled `ModelEdit`s get ignored)
- thus `command.canExecute()` is false and an Exception is thrown

if `ModelEditChange` is disabled, `perform()` never gets called - and enabling/disabling in preview works correctly
So this happens if you don't change anything in the refactoring preview

mx990: Thank you for clarifying. In that case, please use `setEnabledShallow` instead:
```suggestion
		// disable if none of the ModelEdits are enabled
		setEnabledShallow(edits.stream().anyMatch(ModelEdit::isEnabled));
```

pazi146: done

mx990 (approved):

oberlehner (approved):

azoitl (approved):

PR #1574, opened 2025-08-26
Quickfix for saving AttributeType if it is used in Connections

sebHollersbacher:


azoitl (changes requested):
I think it would be better to catch this exception at the save search dialog so that we get all search problems that may arrise and not just surpress this one here.

oberlehner (approved):

azoitl (approved):

Issue #1573, opened 2025-08-26
XML serialization of .sys uses locale-specific decimal separators; parsing fails for comma locales (-nl)


rudoc:
When the 4diac-IDE is started with a locale via **-nl** that uses a comma as the decimal separator (e.g., de, ru), the generated *.sys XML serializes position attributes x, y, dx, dy using a comma. On reading/parsing the same .sys, the application fails with a parsing exception (Position value (x or y) in wrong format!), so the file is not round-trip safe.

This makes the *.sys XML format locale-dependent and breaks parsing for users with comma locales.

**Steps to Reproduce**
Start the 4diac-IDE with a comma-decimal locale, e.g.:
-nl de
or
-nl ru
Create BOOL2BOOL, LREAL2LREAL, INT2INT, connect events, use "Auto Layouting".

**Expected Behavior**
*.sys XML serialization should be locale-invariant and consistently use a dot (.) as the decimal separator for numeric attributes.
The same file should parse correctly regardless of the UI locale.

**Actual Behavior**
Attributes are serialized with a comma (,) as the decimal separator under comma locales.
Parsing the generated *.sys XML fails with a
<details>
<summary>Position value (x or y) in wrong format!</summary>

### Error loading type testXY.sys: Position value (x or y) in wrong format!

```log
org.eclipse.fordiac.ide.model.dataimport.exceptions.TypeImportException: Position value (x or y) in wrong format!
	at org.eclipse.fordiac.ide.model.dataimport.CommonElementImporter.getXandY(CommonElementImporter.java:401)
	at org.eclipse.fordiac.ide.model.dataimport.FBNetworkImporter.parseFB(FBNetworkImporter.java:207)
	at org.eclipse.fordiac.ide.model.dataimport.FBNetworkImporter.handleFBNetworkChild(FBNetworkImporter.java:116)
	at org.eclipse.fordiac.ide.model.dataimport.SubAppNetworkImporter.handleFBNetworkChild(SubAppNetworkImporter.java:58)
	at org.eclipse.fordiac.ide.model.dataimport.CommonElementImporter.processChildren(CommonElementImporter.java:278)
	at org.eclipse.fordiac.ide.model.dataimport.FBNetworkImporter.parseFBNetwork(FBNetworkImporter.java:99)
	at org.eclipse.fordiac.ide.model.dataimport.SystemImporter.lambda$5(SystemImporter.java:481)
[...]

PR #1576, opened 2025-08-26
Provide locale for FB Position export #1573

azoitl:
The generation of the position text during xml export was not locale independent which lead to wrong XML files.

Fixes: https://github.com/eclipse-4diac/4diac-ide/issues/1573

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
@@ -26,7 +26,6 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -183,7 +182,6 @@ public class CommonElementExporter {
 	public static final String LINE_END = "\n"; //$NON-NLS-1$
 	public static final String TAB = "\t"; //$NON-NLS-1$
 	private static final Pattern CDATA_END_PATTERN = Pattern.compile("\\]\\]>"); //$NON-NLS-1$
-	protected static final DecimalFormat positionFormater = new DecimalFormat("#.##"); //$NON-NLS-1$
 
 	private final XMLStreamWriter writer;
 	private final ByteBufferOutputStream outputStream;
@@ -524,8 +522,8 @@ public class CommonElementExporter {
 	}
 
 	protected void addXYAttributes(final double x, final double y) throws XMLStreamException {
-		writer.writeAttribute(LibraryElementTags.X_ATTRIBUTE, positionFormater.format(x));
-		writer.writeAttribute(LibraryElementTags.Y_ATTRIBUTE, positionFormater.format(y));
+		writer.writeAttribute(LibraryElementTags.X_ATTRIBUTE, formatPosOrSizeVal(x));
+		writer.writeAttribute(LibraryElementTags.Y_ATTRIBUTE, formatPosOrSizeVal(y));
 	}
 
 	protected void writeCDataSection(final String cdataText) throws XMLStreamException {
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
```diff
@@ -183,7 +185,8 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
 	public static final String LINE_END = "\n"; //$NON-NLS-1$
 	public static final String TAB = "\t"; //$NON-NLS-1$
 	private static final Pattern CDATA_END_PATTERN = Pattern.compile("\\]\\]>"); //$NON-NLS-1$
-	protected static final DecimalFormat positionFormater = new DecimalFormat("#.##"); //$NON-NLS-1$
+	protected static final DecimalFormat positionFormater = new DecimalFormat("#.##", //$NON-NLS-1$
```
ernstblechaPT: as I had to learn today: `positionFormater` should be `positionFormatter`

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
```diff
@@ -183,7 +185,8 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
+			DecimalFormatSymbols.getInstance(Locale.US));
```
ernstblechaPT: This does the right thing for the decimal seperator, but tying it to a locale feels wrong; 

Especially for Locale.US: does this work with the grouping / thousands separator?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
azoitl: Thx fixed that.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
azoitl: All what I found is that java always is applying for all formatters the locale. So explicitly choosing seems to be the only option. I don't get the thousands separator or any grouping because my format pattern does not specify one. Also tested it on my machine.

ernstblechaPT: If the grouping has to be enabled in the pattern: LGTM!

ernstblechaPT (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
mx990: `DecimalFormat` is unfortunately quite slow when not using a specific subset of formatting options (including  grouping), because its internal fast path does not apply otherwise[^1]. It also turns out `DecimalFormat` is not thread-safe[^2].

Since we explicitly do not want locale-dependent formatting, how about one of the following options:
- for maximum control:
  ```java
  BigDecimal.valueOf(x).round(new MathContext(2, RoundingMode.HALF_EVEN)).toPlainString()
  ```
- for simplicity and maximum performance:
  ```java
  Double.toString(Math.round(x * 100.0) / 100.0)
  ```

[^1]: see JavaDoc of `java.text.DecimalFormat.checkAndSetFastPathStatus()`
[^2]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/text/DecimalFormat.html#synchronization

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
azoitl: This is a good point. When searching I was to focused on the locale. I like the Double approach. But added a .0 trim afterwards to not produce additional noise for our existing users.

mx990 (approved):

PR #1569, opened 2025-08-25
Added Inherit-Mode for AttributeTypes to copy Attributes from FBTypes to FBInstances

sebHollersbacher:
Contains PR #1568 and #1536

PR #1581, opened 2025-08-27
Inherit attributes for Device/Resource/Segment

sebHollersbacher:
Follow up of #1569

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
@@ -15,17 +15,59 @@ package org.eclipse.fordiac.ide.model;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 import org.eclipse.fordiac.ide.model.datatype.helper.InternalAttributeDeclarations;
 import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
 import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableObject;
+import org.eclipse.fordiac.ide.model.libraryElement.Device;
+import org.eclipse.fordiac.ide.model.libraryElement.DeviceType;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetworkElement;
+import org.eclipse.fordiac.ide.model.libraryElement.FBType;
+import org.eclipse.fordiac.ide.model.libraryElement.IVarElement;
+import org.eclipse.fordiac.ide.model.libraryElement.Resource;
+import org.eclipse.fordiac.ide.model.libraryElement.ResourceType;
+import org.eclipse.fordiac.ide.model.libraryElement.Segment;
+import org.eclipse.fordiac.ide.model.libraryElement.SegmentType;
+import org.eclipse.fordiac.ide.model.libraryElement.TypedConfigureableObject;
+import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
 
 public enum AttributeInheritMode {
 	IGNORE, COPY, INHERIT, COPY_INHERIT;
 
-	public static void copyAttributes(final ConfigurableObject conf, final List<Attribute> attributes) {
+	public static void copyAttributeValuesFromType(final TypedConfigureableObject newElement) {
+		if (newElement.getTypeEntry() == null
+				|| !(newElement.getTypeEntry().getType() instanceof final ConfigurableObject typeElement)) {
+			return;
+		}
+		AttributeInheritMode.copyAttributes(newElement, typeElement.getAttributes());
+
+		if (newElement instanceof final FBNetworkElement fbne && typeElement instanceof final FBType fbType) {
+			fbType.getInterfaceList().getAllInterfaceElements().forEach(typeInterfaceElement -> {
+				final var newInterface = fbne.getInterfaceElement(typeInterfaceElement.getName());
+				AttributeInheritMode.copyAttributes(newInterface, typeInterfaceElement.getAttributes());
+			});
+		} else if (newElement instanceof final Device device && typeElement instanceof final DeviceType deviceType) {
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateFBNElementCommand.java
```diff
@@ -133,15 +133,6 @@ public void execute() {
 		// set Visible attribute after reconnect, to not hide connected In/Outputs
 		// transfer attributes from type first (for new vars), then override them from
 		// old instance
```
mx990: I believe the comment is also no longer valid?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeInheritAttributeCommand.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.fordiac.ide.model.libraryElement.AttributeDeclaration;
+import org.eclipse.gef.commands.Command;
+
+public class ChangeInheritAttributeCommand extends Command implements ScopedCommand {
+	private final AttributeDeclaration attributeDeclaration;
+	private final String newValue;
+	private final String oldValue;
+
+	public ChangeInheritAttributeCommand(final AttributeDeclaration attributeDeclaration,
+			final String attributeInheritValue) {
```
mx990: Would it not be better to use `AttributeInheritMode` here?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
```diff
@@ -0,0 +1,106 @@
...
+	IGNORE(Messages.AttributeInherit_Ignore), COPY(Messages.AttributeInherit_Copy),
+	INHERIT(Messages.AttributeInherit_Inherit);
+
+	private final String name;
+
+	AttributeInheritMode(final String displayName) {
+		this.name = displayName;
+	}
+
+	public String getName() {
```
mx990: I would suggest renaming this to `getDisplayName()` to avoid confusion with `name()` from `Enum`.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
```diff
@@ -0,0 +1,106 @@
...
+								.getName()) instanceof final Attribute declarationsInheritAttibute
+				&& AttributeInheritMode.getInheritMode(declarationsInheritAttibute.getValue()) == mode;
+	}
+
+	private static AttributeInheritMode getInheritMode(final String name) {
+		for (final AttributeInheritMode mode : AttributeInheritMode.values()) {
+			if (mode.getName().equals(name)) {
+				return mode;
+			}
+		}
```
mx990: I believe you should be able to use `valueOf(String)` instead?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
```diff
@@ -0,0 +1,106 @@
...
+	public String getName() {
+		return name;
+	}
+
+	public static int getDefaultIndex() {
+		return IGNORE.ordinal();
+	}
+
+	public static void copyAttributeValuesFromType(final TypedConfigureableObject newElement,
+			final AttributeInheritMode mode) {
```
mx990: It looks like you are also copying attributes with mode *inherit*. I believe those should only be shown to the user in the instance property sheet until the value is changed?

mx990 (changes requested):
Would it not be better to copy attributes centrally in `BlockInstanceFactory` and `InterfaceListCopier`?

sebHollersbacher:
> Would it not be better to copy attributes centrally in `BlockInstanceFactory` and `InterfaceListCopier`?

I think this would not work here, as they are both used for updating and creating of FB's

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
sebHollersbacher: The way I thought it should work is that AttributeTypes with Copy should be copied when creating an Instance of an FB and AttributeTypes with Inherit should be copied when updating the FB

sebHollersbacher: @mx990 
It should work as intended now

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/AttributeInheritMode.java
```diff
@@ -15,17 +15,59 @@
...
 
-	public static void copyAttributes(final ConfigurableObject conf, final List<Attribute> attributes) {
+	public static void copyAttributeValuesFromType(final TypedConfigureableObject newElement) {
+		if (newElement.getTypeEntry() == null
+				|| !(newElement.getTypeEntry().getType() instanceof final ConfigurableObject typeElement)) {
+			return;
+		}
+		AttributeInheritMode.copyAttributes(newElement, typeElement.getAttributes());
+
+		if (newElement instanceof final FBNetworkElement fbne && typeElement instanceof final FBType fbType) {
```
mx990, line 45: What about interface elements of typed subapps?

mx990 (approved):

azoitl (approved):

PR #1583, opened 2025-08-27
Replaced getType with better typeEntry functionality

azoitl:
As getType may lead to a full reparse of the file at several places getType is replaced with less heavy weight functions.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
@@ -194,8 +194,7 @@ public class RenameTypeRefactoringParticipant extends RenameParticipant {
 				if (rootContainer instanceof AttributeDeclaration) {
 					return new RenameUpdateStructDataTypeMemberVariableChange(varDecl);
 				}
-				if (rootContainer instanceof final FBType fbType
-						&& dataTypeEntry.getType() instanceof final StructuredType type) {
+				if (rootContainer instanceof final FBType fbType && dataTypeEntry.getType() instanceof StructuredType) {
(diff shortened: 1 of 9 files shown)


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
```diff
@@ -194,8 +194,7 @@ private Change createSubChange(final VarDeclaration varDecl, final DataTypeEntry
 				if (rootContainer instanceof AttributeDeclaration) {
 					return new RenameUpdateStructDataTypeMemberVariableChange(varDecl);
 				}
-				if (rootContainer instanceof final FBType fbType
-						&& dataTypeEntry.getType() instanceof final StructuredType type) {
+				if (rootContainer instanceof final FBType fbType && dataTypeEntry.getType() instanceof StructuredType) {
```
mx990, line 197: Could we also use `getTypeEClass()` here?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameTypeRefactoringParticipant.java
azoitl: Unfortunately no. `getTypeEntryImpl.getTypeEClass()` always returns `DataPackage.Literals.ANY_DERIVED_TYPE`. One of the reasons I haven't looked to into `getType()` calls of `DataTypeEntries`. The other reason is that for data types we always need in the model the type so I expected less performance improvements here.

mx990: Ah, that makes sense. Thanks for clarifying!

mx990 (approved):

PR #1593, opened 2025-09-02
Fix VAR Config column for composite types

Felix-Schmid:
Fixes a bug where null was returned for a method that returns a primitive boolean, causing a NPE. Can be replicated by trying to view the instance properties of a composite FB.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/VarDeclarationColumnAccessor.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/VarDeclarationColumnAccessor.java
@@ -73,7 +73,7 @@ public class VarDeclarationColumnAccessor extends AbstractColumnAccessor<VarDecl
 		};
 	}
 
-	private static boolean getVarConfString(final VarDeclaration rowObject) {
+	private static Boolean getVarConfString(final VarDeclaration rowObject) {
 		return isCompositeFBType(rowObject) ? null : Boolean.valueOf(rowObject.isVarConfig());
 	}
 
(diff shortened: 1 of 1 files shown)


azoitl (changes requested):
Thx for finding this. However I think the whole check is wrong. CFBInstances may also have var config pins. @oberlehner what do you think?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/VarDeclarationColumnAccessor.java
```diff
@@ -73,7 +73,7 @@ public Object getDataValue(final VarDeclaration rowObject, final VarDeclarationT
 		};
 	}
 
-	private static boolean getVarConfString(final VarDeclaration rowObject) {
+	private static Boolean getVarConfString(final VarDeclaration rowObject) {
 		return isCompositeFBType(rowObject) ? null : Boolean.valueOf(rowObject.isVarConfig());
```
azoitl: After a longer offline discussion with @oberlehner we noticed that the came to the conclusion that the isCompositeFBType is checking the wrong thing. It should check if the pin is from an FB inside of a composite FB type. 

Also the name of this method is missleading.

@Felix-Schmid would you do this or should I do this on top of your commit?

Felix-Schmid: Ah, ok. Yes, I will do it.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/VarDeclarationColumnAccessor.java
```diff
@@ -74,12 +73,19 @@ public Object getDataValue(final VarDeclaration rowObject, final VarDeclarationT
 	}
 
 	private static Boolean getVarConfString(final VarDeclaration rowObject) {
```
azoitl: could you also give that method a sensible name, e.g. getVarConfValue

Felix-Schmid: sure, done.

azoitl (approved):

PR #1602, opened 2025-09-03
Cleaned FBNetwork Editor to reduce requirement on getSystem

azoitl:
getSystem was badly introduced and missused in our diagram and network editors.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBNetworkEditor.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBNetworkEditor.java
@@ -21,11 +21,9 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.editors;
 
-import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.geometry.Dimension;
-import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.application.actions.CopyEditPartsAction;
 import org.eclipse.fordiac.ide.application.actions.CutEditPartsAction;
 import org.eclipse.fordiac.ide.application.actions.DeleteFBNetworkAction;
@@ -43,9 +41,9 @@ import org.eclipse.fordiac.ide.gef.DiagramEditorWithFlyoutPalette;
 import org.eclipse.fordiac.ide.gef.preferences.PaletteFlyoutPreferences;
 import org.eclipse.fordiac.ide.gef.tools.AdvancedPanningSelectionTool;
 import org.eclipse.fordiac.ide.model.CoordinateConverter;
+import org.eclipse.fordiac.ide.model.helpers.ModelHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.AutomationSystem;
 import org.eclipse.fordiac.ide.model.libraryElement.FBNetwork;
-import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
 import org.eclipse.fordiac.ide.model.ui.actions.Open4DIACElementAction;
@@ -122,20 +120,17 @@ public class FBNetworkEditor extends DiagramEditorWithFlyoutPalette {
 	}
 
 	protected TypeLibrary getTypeLibrary() {
-		return TypeLibraryManager.INSTANCE.getTypeLibraryFromContext(getModel());
-	}
-
-	private IProject getProject() {
-		if (EcoreUtil.getRootContainer(getModel()) instanceof final LibraryElement libElem) {
-			return libElem.getTypeEntry().getFile().getProject();
+		final TypeLibrary typeLib = TypeLibraryManager.INSTANCE.getTypeLibraryFromContext(getModel());
+		if (typeLib != null) {
+			return typeLib;
 		}
-		throw new IllegalStateException(
-				"Could not determine root element for finding the project for given element: " + getModel()); //$NON-NLS-1$
+		throw new IllegalStateException("Could not get type lib for: " + getModel()); //$NON-NLS-1$
 	}
 
 	@Override
 	protected TransferDropTargetListener createTransferDropTargetListener() {
-		return new FbTypeTemplateTransferDropTargetListener(getGraphicalViewer(), getProject());
+		return new FbTypeTemplateTransferDropTargetListener(getGraphicalViewer(),
+				ModelHelper.getProjectFromContextChecked(getModel()));
 	}
 
 	@Override
@@ -231,7 +226,8 @@ public class FBNetworkEditor extends DiagramEditorWithFlyoutPalette {
 
 	@Override
 	protected PaletteViewerProvider createPaletteViewerProvider() {
-		return new FBTypePaletteViewerProvider(getProject(), getEditDomain(), getPaletteNavigatorID());
+		return new FBTypePaletteViewerProvider(ModelHelper.getProjectFromContextChecked(getModel()), getEditDomain(),
+				getPaletteNavigatorID());
 	}
 
 	/**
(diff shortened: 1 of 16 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBNetworkEditor.java
```diff
@@ -118,13 +122,20 @@ protected ContextMenuProvider getContextMenuProvider(final ScrollingGraphicalVie
...
 
 	protected TypeLibrary getTypeLibrary() {
-		return getSystem().getTypeLibrary();
+		return TypeLibraryManager.INSTANCE.getTypeLibraryFromContext(getModel());
+	}
+
+	private IProject getProject() {
+		if (EcoreUtil.getRootContainer(getModel()) instanceof final LibraryElement libElem) {
+			return libElem.getTypeEntry().getFile().getProject();
+		}
```
mx990: As this has come up in several places over time, should we add a common helper method for this, similar to `getTypeLibraryFromContext`?

azoitl: Would be a good idea also searched if we have that already but not yet. Didn't know where a good place for that would be. Any suggestions?

mx990: How about just adding it into `TypeLibraryManager`?

azoitl: feels a bit odd. How about a ModelHelper class in the models.helper package and add methods for:
  - getLibraryElementFromContext
  - getFileFromContext
  - getProjectFromContext

mx990: Sounds good to me.

azoitl: Done in a separate commit. I decided to trow an exception if the context is not contained in a library element.

bwiesmayr (approved):
I like the new ModelHelper, this really cleans things up, I hope I didn't miss any errors :D

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
```diff
@@ -0,0 +1,115 @@
...
+	 * achieved by getting the root {@code LibraryElement} and its
+	 * {@code TypeEntry}.
+	 *
+	 * @param context object for which to get the {@link IFile}
+	 * @return {@link IFile} the context object is contained in
+	 * @throws IllegalStateException if the {@link IFile} can not be determined from
+	 *                               context object
+	 */
+	public static IFile getFileFromContextChecked(final EObject context) {
+		return getLibraryElementFromContextChecked(context).getTypeEntry().getFile();
```
mx990: What if the library element has no type entry, such as for elementary data types?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
```diff
@@ -0,0 +1,115 @@
...
+	 * achieved by getting the root {@code LibraryElement} and its
+	 * {@code TypeEntry}.
+	 *
+	 * @param context object for which to get the {@link IProject}
+	 * @return {@link IProject} the context object is contained in
+	 * @throws IllegalStateException if the {@link IProject} can not be determined
+	 *                               from context object
+	 */
+	public static IProject getProjectFromContextChecked(final EObject context) {
+		return getFileFromContextChecked(context).getProject();
```
mx990: What about error types, which do have a type library with a project but not an associated file?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
```diff
@@ -73,7 +71,7 @@ public final class FBShapeHelper {
 	 * @return The width in X coordinates
 	 */
 	public static double getWidth(final FBNetworkElement element) {
-		final IProject project = getProjectFromFBNE(element);
+		final IProject project = ModelHelper.getProjectFromContextChecked(element);
```
mx990: I believe the project can safely be `null` here.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AddFBBookMark.java
```diff
@@ -48,7 +45,7 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final FBNetworkElement element = getSelectedFBElement(event);
 		if (null != element) {
 			final String description = getDescription(element, event);
-			final IResource file = getFile(element);
+			final IResource file = ModelHelper.getFileFromContextChecked(element);
 			if (description != null && file != null) {
```
mx990: I believe either the `Checked` or the null check is redundant.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBNetworkEditor.java
```diff
@@ -118,13 +120,13 @@ protected ContextMenuProvider getContextMenuProvider(final ScrollingGraphicalVie
 	}
 
 	protected TypeLibrary getTypeLibrary() {
-		return getSystem().getTypeLibrary();
+		return TypeLibraryManager.INSTANCE.getTypeLibraryFromContext(getModel());
```
mx990: I believe the type library may not be `null` here.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
azoitl: You are right I should have used the typelib instead of the getFile

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
azoitl: I think using an elementary data type here would be wrong. So throwing an exception is IMHO the right thing.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
azoitl: changed accordingly

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
azoitl: you are right, even though it shouldn't. Changed accordingly.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AddFBBookMark.java
azoitl: yes. removed the null check.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBNetworkEditor.java
azoitl: You are right. Added an exception to get early feedback on such problems.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
mx990: Yes, but maybe not an NPE when `getTypeEntry()` returns `null`?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/FBShapeHelper.java
mx990: I do not see any change here, just in line 129 below?

azoitl: Ah missed that there are twice the project request. Fixed now here as well.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/ModelHelper.java
azoitl: good point. added further null checks to avoid that.

mx990 (approved):

PR #1609, opened 2025-09-07
Fixed switch expression to comply with stricter Java rules

azoitl:
Cheery-picked from develop so that people can use the upcoming 2025-09 Eclipse release also with the freeze branch.


azoitl:
Build error is caused by missing PR #1608. Will still merge it.

PR #1612, opened 2025-09-07
Upgrade Build System to Tycho 5.0

azoitl:
As part of this upgrade all source features where removed from the target file. All versions from features are changed to 0.0.0 to always take the latest from the give p2 repository. The later will simplify upgrading to newer releases.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Set up Maven
       uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
       with:
-        maven-version: 3.9.6
+        maven-version: 3.9.11
     - name: Build with Maven
       uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/maven.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     - name: Set up Maven
       uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
       with:
-        maven-version: 3.9.6
+        maven-version: 3.9.9
```
ernstblechaPT: Is there a reason not to use maven 3.9.11?

MartinMelikMerkumians: Do we need to fix the patch version anyway or could we just say 3.9?

azoitl: To be honest, Eclipse Tycho was complaining that it needs 3.9.9 therefore I changed to that to see if the build is then running through. I'm totally open to make our lives easier here

azoitl: I just checked stCarolas/setup-maven requires to specify the full version including the maintainence version.

azoitl: changed to 3.9.11

mx990 (approved):

MartinMelikMerkumians (approved):

PR #1616, opened 2025-09-09
Create connection into collapsed subapp

MandKastner:
This feature allows Interface elements to me marked by using the context menu entry : Mark as connection source (Shortcut: Left click + SHIFT + CTRL).  After clicking on a second interface element the connection will be established automatically using the marked element as source.


plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/MarkerStore.java
```diff
@@ -0,0 +1,87 @@
...
+		if (EditorUtils.getCurrentActiveEditor() instanceof final MarkerStoreProvider provider) {
+			return Optional.ofNullable(provider.getMarkerStore());
+		}
+		return Optional.empty();
+	}
+
+	/*
+	 * Maps Marker descriptor ID to Editparts
+	 */
+	private final HashMap<String, GraphicalEditPart> store = new HashMap<>();
```
oberlehner: why dont you store the full qualified name of the INamedElement here ?
what if the editor reloads. then the reference to the editPart is lost.
Also split editor is not working because of this.

MandKastner: Before i stored the EditPart reference, i used the EMF URI of the marked model elements. Altough it is possible to get the corresponding EObject using the editor as search context, at some points i need the edit part again. In this case not even the editpart registry can deliver that as it depends on the graphical viewer that might be different. If you have any other suggestions how to solve that issue please let me know as i am also not happy with the current solution.

mx990: How about using the existing marker and graphical annotation infrastructure? That could also be used with qualified names instead of URIs if necessary[^1].

[^1]: see [`DeploymentWatchpoint.getTargets(AutomationSystem)` ](https://github.com/eclipse-4diac/4diac-ide/blob/c4c5485cc8ea9540a9ab9ee1dfa8765ab1c1b5d7/plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/breakpoint/DeploymentWatchpoint.java#L91) for an example

oberlehner: > How about using the existing marker and graphical annotation infrastructure? That could also be used with qualified names instead of URIs if necessary[1](#user-content-fn-1-b1b3d361b160ca0bb0f17db7e7364d03).
> 
> ## Footnotes
> 1. see [`DeploymentWatchpoint.getTargets(AutomationSystem)` ](https://github.com/eclipse-4diac/4diac-ide/blob/c4c5485cc8ea9540a9ab9ee1dfa8765ab1c1b5d7/plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/breakpoint/DeploymentWatchpoint.java#L91) for an example [↩](#user-content-fnref-1-b1b3d361b160ca0bb0f17db7e7364d03)

That sounds promising ! thx for the hint

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/GraphicalAnnotation.java
```diff
@@ -21,6 +21,9 @@ public abstract class GraphicalAnnotation {
 	public static final String TYPE_WARNING = "org.eclipse.fordiac.ide.model.ui.annotation.warning"; //$NON-NLS-1$
 	public static final String TYPE_INFO = "org.eclipse.fordiac.ide.model.ui.annotation.info"; //$NON-NLS-1$
 
+	public static final String TYPE_UTIL_PREDECESSOR = "org.eclipse.fordiac.ide.model.ui.annotation.util.predecessor"; //$NON-NLS-1$
+	public static final String TYPE_UTIL_CONNECTION_SRC = "org.eclipse.fordiac.ide.model.ui.annotation.util.connectionSource"; //$NON-NLS-1$
+
```
mx990: It seems these constants are only used by the `UtilityMarkerAnnotationStyler`, so I would suggest moving them there.

plugins/org.eclipse.fordiac.ide.model/plugin.xml
```diff
@@ -315,4 +315,28 @@
...
+       name="Connection Source Marker"
+       point="org.eclipse.core.resources.markers">
+    <super
+          type="org.eclipse.core.resources.marker">
+    </super>
+    <persistent
+          value="false">
+    </persistent>
+ </extension>
+
```
mx990: As these markers are only used while editing in the UI, I would suggest moving them to `org.eclipse.fordiac.ide.model.ui`.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.model.errormarker.ErrorMarkerBuilder;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
+import org.eclipse.fordiac.ide.model.libraryElement.AutomationSystem;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
+import org.eclipse.fordiac.ide.ui.editors.EditorUtils;
+
+public class UtilityMarkerHelper {
```
mx990: I would suggest to have a look at `FordiacErrorMarker` and `FordiacMarkerHelper` for working with markers. They should cover most of what you are doing here. Importantly, they also properly handle workspace modifications without potential deadlocks in the UI thread.

If you believe they are lacking any useful helper methods, please suggest how to extend them, rather than creating a separate class in a different package.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/MarkerDescriptor.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Contributors:
+ *   Mario Kastner - initial implementation and/or documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.util.marker;
+
+import org.eclipse.jface.resource.JFaceResources;
+import org.eclipse.swt.graphics.Color;
+
+public record MarkerDescriptor(String ID, String name, boolean isUnique, Color color) {
```
mx990: Why do you mix marker types with color preferences here? The whole point of separating annotations and markers is to separate UI and non-UI concerns, respectively.

I would suggest to move all of the color handling to your annotation styler(s) and define plain string constants for your marker types in a new class within `org.eclipse.fordiac.ide.model.ui`.

plugins/org.eclipse.fordiac.ide.ui/plugin.xml
```diff
@@ -192,6 +192,12 @@
              label="%FordiacPreferencePage_LABEL_PredecessorMarkerColor"
              value="99, 179, 31">
        </colorDefinition>
+       <colorDefinition
+             categoryId="org.eclipse.fordiac.ide.themeCategory"
+             id="org.eclipse.fordiac.ide.ui.ConnectionSourceMarkerColor"
+             label="%FordiacPreferencePage_LABEL_ConnectionSourceMarkerColor"
+             value="57, 73, 171">
+       </colorDefinition>
```
mx990: I would suggest to move these colors into the same plugin as the annotation styler.

plugins/org.eclipse.fordiac.ide.model.ui/plugin.xml
```diff
@@ -39,6 +39,18 @@
             name="org.eclipse.fordiac.ide.model.ui.annotation.info"
             super="org.eclipse.ui.workbench.texteditor.info">
       </type>
+      <type
+            markerSeverity="0"
+            markerType="org.eclipse.fordiac.ide.model.predecessor"
+            name="org.eclipse.fordiac.ide.model.ui.annotation.util.predecessor"
+            super="org.eclipse.ui.workbench.texteditor.info">
```
mx990: I believe none of the existing textual annotation types are a good fit here, particularly since these annotations are only graphical. I would suggest to just remove the super type.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
```diff
@@ -433,12 +503,14 @@ public void activate() {
...
-		getModel().eAdapters().remove(getContentAdapter());
-		removeSourcePinAdapter();
+		if (isActive()) {
+			super.deactivate();
+			getModel().eAdapters().remove(getContentAdapter());
+			((AdvancedScrollingGraphicalViewer) getViewer()).getPreferencesCache().getStoreProvider()
+					.removePropertyChangeListener(preferenceListener);
+			getModel().eAdapters().remove(getContentAdapter());
+			removeSourcePinAdapter();
+		}
```
mx990: @azoitl I remember you recently removed the `isActive()` check here[^1]. Should this be re-added?

[^1]: see https://github.com/eclipse-4diac/4diac-ide/pull/1651/files#diff-1f0cef95550cb2f514a311b1c789684ef085a8e6a47be798474736a63389d8ef

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/UtilityMarkerAnnotationStyler.java
```diff
@@ -0,0 +1,113 @@
...
+		final StatusLineContributionItem item = new StatusLineContributionItem("", text.length()); //$NON-NLS-1$
+		item.setId(id);
+		item.setText(text);
+		return item;
+	}
+
+	private static IStatusLineManager getStatusLineManager() {
+		return PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().getActiveEditor().getEditorSite()
+				.getActionBars().getStatusLineManager();
+	}
```
mx990: The status line is supposed to be managed by the editor action contributor[^1]. You could then retrieve the item via the contributor to update its text. I also believe an annotation styler is not the right place for that. I would suggest to update the status line directly in your handler instead.

[^1]: see also `org.eclipse.fordiac.ide.typeeditor.TypeEditorContributor` and `org.eclipse.ui.texteditor.BasicTextEditorActionContributor.contributeToStatusLine(IStatusLineManager)`

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/listeners/UtilityMarkerListener.java
```diff
@@ -0,0 +1,104 @@
...
+	protected void executeOperation(final AbstractOperation op) {
+		// do nothing
+	}
+
+	private static void removeInvalidMarkers(final List<QualNameChange> qualNameChange) {
+		final Set<String> oldQualNames = new HashSet<>();
+		qualNameChange.stream().map(QualNameChange::oldQualName).forEach(oldQualNames::add);
+		UtilityMarkerHelper.getAllUtilityMarkers().filter(m -> oldQualNames.contains(getLocation(m))).forEach(m -> {
+			try {
+				m.delete();
```
mx990: Modifying the workspace from the UI thread may lead to deadlocks. Please use the helpers from `FordiacMarkerHelper` instead.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
azoitl: I agree that this class should be moved and put into a better location. However I have a hard time to move this to the marker helpers. This class is very specific for this utility markers and what is needed for them.

mx990: I have added comments on the individual methods below. I believe everything besides `getMarkedElement(...)` and `addElementMarker(...)` can be either inlined or removed entirely.

I agree that both remaining methods are quite specific to these "utility" markers. We may still be able to justify putting them into `FordiacMarkerHelper`, but it should also be fine to leave them here.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+		try {
+			return Stream.of(resource.findMarkers(markerDescriptor.ID(), true, IResource.DEPTH_ZERO))
+					.map(UtilityMarkerHelper::getTarget).findAny().orElse(null);
+		} catch (final CoreException e) {
+			FordiacLogHelper.logError("Error accessing marked element", e); //$NON-NLS-1$
+		}
+		return null;
+	}
+
+	public static IMarker getMarker(final MarkerDescriptor markerDescriptor) {
```
mx990: I believe this is unused and can thus be removed.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+		return null;
+	}
+
+	public static void deleteElementMarker(final MarkerDescriptor markerDescriptor, final IResource resource) {
+		deleteMarker(markerDescriptor.ID(), resource);
+	}
+
+	public static void deleteElementMarker(final String id) {
```
mx990: I believe this is unused and can thus be removed.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+			final EObject target) {
+		try {
+			ErrorMarkerBuilder.createErrorMarkerBuilder(markerDescriptor.name()).setType(markerDescriptor.ID())
+					.setTarget(target).setSeverity(IMarker.SEVERITY_INFO).createMarker(resource);
+		} catch (final CoreException e) {
+			FordiacLogHelper.logError("Cannot add marker", e); //$NON-NLS-1$
+		}
+	}
+
+	public static boolean hasMarker(final MarkerDescriptor markerDescriptor, final IResource resource) {
```
mx990: This is seems to be used only once in `InterfaceEditPart.createFigure().new MouseListener() {...}.mousePressed(MouseEvent)`, so I would suggest to move it there.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+		try {
+			return Stream.of(resource.findMarkers(markerDescriptor.ID(), true, IResource.DEPTH_ZERO)).findAny()
+					.isPresent();
+		} catch (final CoreException e) {
+			FordiacLogHelper.logError("Cannot fetch marker", e); //$NON-NLS-1$
```
mx990: I would suggest to just silently return false here.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+						Stream.of(sys.getTypeEntry().getFile().findMarkers(MarkerDescriptor.PREDECESSOR.ID(), false,
+								IResource.DEPTH_ZERO)));
+			} catch (final CoreException e) {
+				FordiacLogHelper.logError("Cannot fetch marker", e); //$NON-NLS-1$
+			}
+		}
+		return Stream.empty();
+	}
+
+	public static EObject getTarget(final IMarker marker) {
```
mx990: I believe this is used only in `getMarkedElement(...)` above and can thus either be made private or removed if no longer used.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+						.of(sys.getTypeEntry().getFile().findMarkers(markerDescriptor.ID(), true, IResource.DEPTH_ZERO))
+						.findAny().orElse(null);
+			} catch (final CoreException e) {
+				FordiacLogHelper.logError("Error accessing marked element", e); //$NON-NLS-1$
+			}
+		}
+		return null;
+	}
+
+	public static void deleteElementMarker(final MarkerDescriptor markerDescriptor, final IResource resource) {
```
mx990: I believe this can be replaced inline with:
```java
FordiacMarkerHelper.updateMarkers(resource, markerDescriptor.ID(), List.of(), true)
```

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	public static void addElementMarker(final MarkerDescriptor markerDescriptor, final IResource resource,
+			final EObject target) {
+		try {
+			ErrorMarkerBuilder.createErrorMarkerBuilder(markerDescriptor.name()).setType(markerDescriptor.ID())
+					.setTarget(target).setSeverity(IMarker.SEVERITY_INFO).createMarker(resource);
+		} catch (final CoreException e) {
+			FordiacLogHelper.logError("Cannot add marker", e); //$NON-NLS-1$
+		}
```
mx990: As there may only be once such marker at a given time, I would suggest updating the marker with:
```suggestion
		FordiacMarkerHelper
				.updateMarkers(resource, markerDescriptor.ID(),
						List.of(ErrorMarkerBuilder.createErrorMarkerBuilder(markerDescriptor.name())
								.setType(markerDescriptor.ID()).setTarget(target).setSeverity(IMarker.SEVERITY_INFO)),
						true);
```
This already handles deleting any old markers with the given type.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+		try {
+			return Stream.of(resource.findMarkers(markerDescriptor.ID(), true, IResource.DEPTH_ZERO)).findAny()
+					.isPresent();
+		} catch (final CoreException e) {
+			FordiacLogHelper.logError("Cannot fetch marker", e); //$NON-NLS-1$
+		}
+		return false;
+	}
+
+	public static Stream<IMarker> getAllUtilityMarkers() {
```
mx990: This seems to be used only in `UtilityMarkerListener`, so I would suggest moving it there.

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,123 @@
...
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
+import org.eclipse.fordiac.ide.ui.editors.EditorUtils;
+
+public class UtilityMarkerHelper {
+
+	public static EObject getMarkedElement(final MarkerDescriptor markerDescriptor, final IResource resource) {
+		try {
+			return Stream.of(resource.findMarkers(markerDescriptor.ID(), true, IResource.DEPTH_ZERO))
+					.map(UtilityMarkerHelper::getTarget).findAny().orElse(null);
```
mx990: I believe we should pass the root library element as a parameter and then use `FordiacErrorMarker.getTargetRelative(...)`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.util/src/org/eclipse/fordiac/util/marker/UtilityMarkerHelper.java
azoitl: thx for the feedback that helps a lot.

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
+
+public class UtilityMarkerHelper {
+
+	public static final String PREDECESSOR_MARKER_ID = "org.eclipse.fordiac.ide.model.ui.predecessor"; //$NON-NLS-1$
+	public static final String CONNECTION_SRC_MARKER_ID = "org.eclipse.fordiac.ide.model.ui.connectionSource"; //$NON-NLS-1$
+
+	public static EObject getMarkedElement(final String markerId, final EObject refElement) {
+		final EObject rootContainer = EcoreUtil.getRootContainer(refElement);
+		if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() != null) {
```
mx990:
```suggestion
		if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() == null) {
```

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/UtilityMarkerHelper.java
```diff
@@ -0,0 +1,72 @@
...
+		return null;
+	}
+
+	public static void deleteElementMarker(final String markerId, final IResource resource) {
+		FordiacMarkerHelper.updateMarkers(resource, markerId, List.of(), true);
+	}
+
+	public static void setMarkedElement(final String markerId, final EObject target) {
+		final EObject rootContainer = EcoreUtil.getRootContainer(target);
+		if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() != null) {
```
mx990:
```suggestion
		if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() == null) {
```

plugins/org.eclipse.fordiac.ide.ui/plugin.xml
mx990: I believe these are leftover duplicates now?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
```diff
@@ -201,6 +215,72 @@ private boolean isSubAppPin(final FBNetworkElement connSource) {
 		return false;
 	}
 
+	private final class InterfaceEdiPartMouseListener implements MouseListener {
```
mx990:
```suggestion
	private final class InterfaceEditPartMouseListener implements MouseListener {
```

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
```diff
@@ -201,6 +215,72 @@ private boolean isSubAppPin(final FBNetworkElement connSource) {
...
+
+			// LEFT CLICK
+			if (hasMarker()) {
+				invokeMarkConnectionSourceHandler(null);
+			}
+		}
+
+		private boolean hasMarker() {
+			final EObject rootContainer = EcoreUtil.getRootContainer(InterfaceEditPart.this.getModel());
+			if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() != null) {
```
mx990:
```suggestion
			if (!(rootContainer instanceof final LibraryElement libEl) || libEl.getTypeEntry() == null) {
```

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/InsertFBIntoExecutionChainCommand.java
```diff
@@ -101,7 +99,8 @@ public Set<EObject> getAffectedObjects() {
 
 	private static Command getCreateConnectionCommand(final FBNetwork network, final IInterfaceElement source,
 			final IInterfaceElement target) {
-		if (source.getBlockFBNetworkElement().getFbNetwork() != target.getBlockFBNetworkElement().getFbNetwork()) {
+		if (!EcoreUtil.equals(source.getBlockFBNetworkElement().getFbNetwork(),
+				target.getBlockFBNetworkElement().getFbNetwork())) {
```
mx990: Do we actually need or want a deep equals here?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractMarkerHandler.java
```diff
@@ -0,0 +1,114 @@
...
+		final EObject rootContainer = EcoreUtil.getRootContainer(target);
+		if (rootContainer instanceof final LibraryElement libEl && libEl.getTypeEntry() != null) {
+			return libEl.getTypeEntry().getFile();
+		}
+		return null;
+	}
+
+	private boolean isMarked(final EObject target) {
+		final EObject markedElement = UtilityMarkerHelper.getMarkedElement(getMarkerId(), target);
+		return markedElement != null && EcoreUtil.equals(markedElement, target);
```
mx990: Do we actually need or want a deep equals here?

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/listeners/UtilityMarkerListener.java
mx990: I would suggest to simply use `UtilityMarkerHelper.deleteElementMarker(...)` here now.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/listeners/UtilityMarkerListener.java
```diff
@@ -0,0 +1,124 @@
...
+			} catch (final CoreException e) {
+				FordiacLogHelper.logError(e.getMessage(), e);
+			}
+		});
+
+	}
+
+	private static Stream<IMarker> getAllUtilityMarkers() {
+		if (EditorUtils.getCurrentActiveEditor()
+				.getAdapter(LibraryElement.class) instanceof final AutomationSystem sys) {
```
mx990: Should this be restricted to systems? What about markers in subapps or composite FBs?

plugins/org.eclipse.fordiac.ide.application/plugin.properties
```diff
@@ -46,9 +46,8 @@ FBNetworkElementTooltipFigure_VersionLabel= / Version:
 FBPaletteViewer_SearchForType=Search for Type
 FBPaletteViewer_SelectConnectionEnd=Select Connection End
 FBEditPart_ERROR_UnsupportedFBType=Maybe FBEditPart.createFigureForModel() needs to be extended\!
-FBMarker_ActivePredecessor=Active Predecessor FB: {0}
-FBMarker_MarkAsPredecessor=Mark FB as predecessor
```
mx990, line 50: It seems this text is still used by the menu definition in `plugin.xml` when the handler is disabled.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkPredecessorHandler.java
```diff
@@ -14,147 +14,32 @@
...
-	}
-
-	private static AbstractBlockFBNElementEditPart getValidSelectedFBNElement(final ISelection selection) {
-		if (selection instanceof final IStructuredSelection structSel
-				&& structSel.getFirstElement() instanceof final AbstractBlockFBNElementEditPart ep
+	protected EObject getValidSelectedElement(final Object selectedObject) {
+		if (selectedObject instanceof final AbstractBlockFBNElementEditPart ep
 				&& ep.getRoot() instanceof FBNetworkRootEditPart) {
-			return ep;
+			return ep.getModel();
```
mx990, line 49: Should we enable this only for elements inside (unfolded) subapps, since the predecessor mark seems have no effect otherwise?

plugins/org.eclipse.fordiac.ide.application/plugin.xml
```diff
@@ -1044,6 +1067,11 @@
 	      </and>
         </visibleWhen>
       </command>
```
mx990, line 1053: Should we show this only when the `GefPreferenceConstants.MANAGE_EVENT_CONNECTIONS_AUTOMATICALLY` preference is enabled, since it seems to have no effect otherwise and may be confusing for new users?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/plugin.xml
azoitl: You have a point here but this will be rather hard. For checking the `GefPreferenceConstants.MANAGE_EVENT_CONNECTIONS_AUTOMATICALLY` preference I would need the current project to check if a project or workspace preference is set. In a propertytester I would not have that. What I can do is to add this check in the `checkEnablement` method of the handler so that the menu entry is not enabled when this preference is not set. We can then use the `checkEnabled` to hid it from the context menu or even from the main menu. Would this go in the direction that you had in mind.

plugins/org.eclipse.fordiac.ide.ui/plugin.xml
azoitl: yes removed now

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/annotation/UtilityMarkerAnnotationStyler.java
azoitl: You are right that the annotation styler is not the best place for status line updates. However the handlers also not. I for now removed the status line updates. I think we should reconsider this and do it in a separate PR.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/listeners/UtilityMarkerListener.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/UtilityMarkerHelper.java
azoitl: thx for finding that.

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/UtilityMarkerHelper.java
azoitl: thx for finding that.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
azoitl: fixed

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
azoitl: thx for finding that.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/InsertFBIntoExecutionChainCommand.java
azoitl: I don't think so. replaced with simple ==

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractMarkerHandler.java
azoitl: I don't think so. replaced with simple ==

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/listeners/UtilityMarkerListener.java
azoitl: no. changed accordingly

plugins/org.eclipse.fordiac.ide.application/plugin.properties
azoitl: added dedicated command names and removed the active when from the handler this should provide homogenouse nameing in the ui

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkPredecessorHandler.java
azoitl: yes good point. and based on your comment below I added a check for the preference.

plugins/org.eclipse.fordiac.ide.application/plugin.xml
azoitl: As I first step I've added the anyhow missing step in the enablement check for the handler. Furthermore I changed the pop-up menu entry to use the enablement check for visibility. For the source menu I don't yet know what to do. We could however keep it as is.

mx990: It feels much better now. I also believe we could keep it as is.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkConnectionSourceHandler.java
```diff
@@ -0,0 +1,139 @@
...
+
+		@Override
+		public void execute() {
+			cmd.execute();
+		}
+
+		@Override
+		public void undo() {
+			cmd.undo();
+			FollowConnectionHandler.selectInterfaceElement(jumpBackTarget, editor);
```
mx990: Was this not supposed to restore the marker and *not* change the selection or move the viewport, so that users can quickly recover from a misclick?

azoitl: Hmh. I somehow noted select it again. However I like your proposal more. Will implement it like that.

azoitl: done

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkConnectionSourceHandler.java
```diff
@@ -0,0 +1,134 @@
...
+
+		@Override
+		public void undo() {
+			cmd.undo();
+			UtilityMarkerHelper.setMarkedElement(getMarkerId(), conEndpoint);
+		}
+
+		@Override
+		public void redo() {
+			cmd.redo();
```
mx990, line 121: Do we not need to remove the marker again here?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/MarkConnectionSourceHandler.java
azoitl: good observation. yes we need to changed.

mx990 (approved):

PR #1620, opened 2025-09-11
Changed background in AttributeTargetSection to be consistent

sebHollersbacher:


azoitl (commented):
To get even more consistency you should use the toolkit provided by one of the base classes for creating the widgets.

azoitl (approved):

PR #1641, opened 2025-09-17
Fix import copy when compilerinfo is null

Felix-Schmid:
New compiler infos are now created instead of skipping the import copy when the infos are null.
In addition, adding new imports is done via commands, so that undoing the pasting also removes the newly added imports again.


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/PasteCommand.java
```diff
@@ -72,6 +73,7 @@ public class PasteCommand extends Command implements ScopedCommand {
 
 	private final CompoundCommand connCreateCmds = new CompoundCommand();
 	private final CompoundCommand updateTypeCmds = new CompoundCommand();
+	private final List<AddNewImportCommand> importCmds = new ArrayList<>();
```
oberlehner: Is there a reason why you are not using a CompoundCommand ?

Felix-Schmid: Thanks for the tip, I wasn't aware of that.

oberlehner (approved):

PR #1645, opened 2025-09-21
Update namespaces in forte_ng export

mx990:


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/ForteLibraryElementTemplate.xtend
```diff
@@ -79,8 +80,11 @@ abstract class ForteLibraryElementTemplate<T extends LibraryElement> extends For
 		«type.compilerInfo?.header»
 		
 		using namespace std::literals;
-		using namespace forte::core::literals;
-		using namespace «type.generateTypeNamespace»;
+		using namespace forte;
+		using namespace forte::literals;
+		«IF !PackageNameHelper.getPackageName(type).empty»
+			using namespace «type.generateTypeNamespace»;
```
azoitl: is it really save to use all namespaces from all used types?

mx990: No, this still just adds a single `using` directive for the namespace of the *current* type but *only* if it has a package.

mx990: I have now changed the source templates to use `namespace` blocks, which avoids the extra `using namespace` directives altogether.

azoitl: I think this is the better solution.

azoitl (approved):

PR #1648, opened 2025-09-23
Fix grid dash pattern depending on clip rectangle

Felix-Schmid:
The grid line start and therefore the dash alignment was dependent on the current clip rectangle. This could cause the line dashes to "dance around" and leave behind undesirable patterns, e.g. when dragging a selection from bottom right to top left:
grid-dashes
Fixed by snapping the lines to a global dash pattern.
Also avoided loops when calcualting the line start.


azoitl (approved):
Thx for fixing this nasty issue. For future PRs. Please also try to put some of the description you gave in the par into the commit message.

PR #1649, opened 2025-09-23
(un)check all interface element button for with constraint tables

moniwe:
this adds a checkbox to the tables of the with constraints, that the user can (de)activate all elements. 

image


plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/properties/EventInterfaceElementSection.java
```diff
@@ -58,15 +68,23 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
...
+		deSelectAllWidget.bindToTableViewer(withEventsViewer, this);
 
 		final Table tableWith = withEventsViewer.getTable();
 		configureTableLayout(tableWith);
-		tableWith.addListener(SWT.Selection, event -> {
+		listener = getListener();
+		tableWith.addListener(SWT.Selection, listener);
+	}
+
+	private Listener getListener() {
```
azoitl: this method is creating a listener. So createListener would be the better name.

azoitl (changes requested):
I think it would be better when the DSelectAllWidget is operating not on the table but on a list of interfacelements. Then the widget can create the compound command based on the model state and is independent of the list reperesentation. This would make the widget better usable, maintainable. The latter especially in case of changes in the table.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/properties/EventInterfaceElementSection.java
moniwe: renamed

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/properties/DeSelectAllWidget.java
```diff
@@ -0,0 +1,125 @@
...
+
+	private Listener createDeSelectAllListener(final Button checkbox, final TableViewer withEventsViewer,
+			final AdapterInterfaceElementSection executor) {
+		return ev -> {
+			final CompoundCommand ccmd = new CompoundCommand();
+			final EObject type = executor.getType();
+			if (checkbox.getSelection()) {
+				final Stream<TableItem> unchecktItems = Stream.of(withEventsViewer.getTable().getItems())
+						.filter(item -> !item.getChecked());
+				if (type instanceof Event) {
```
azoitl: Replace this instanceof check and cast with 'instanceof Event event' and use event in the WithcreateCommand

plugins/org.eclipse.fordiac.ide.fbtypeeditor/src/org/eclipse/fordiac/ide/fbtypeeditor/properties/DeSelectAllWidget.java
```diff
@@ -0,0 +1,125 @@
...
+					unchecktItems.forEach(
+							item -> ccmd.add(new WithCreateCommand((Event) type, (VarDeclaration) item.getData())));
+				} else {
+					unchecktItems.forEach(
+							item -> ccmd.add(new WithCreateCommand((Event) item.getData(), (VarDeclaration) type)));
+				}
+			} else if (type instanceof Event) {
```
azoitl: Replace this instanceof check and cast with 'instanceof Event event'

azoitl (changes requested):

azoitl (approved):

PR #1651, opened 2025-09-24
ContentAdapter for EditParts need to Check if EditPart is Active

azoitl:
When a content adapter is refreshing editparts in the display thread it can happen that the editpart was removed in the mean-time. Therefore an isActive check is needed there.


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/editparts/InterfaceEditPart.java
```diff
@@ -428,14 +433,12 @@ public void activate() {
 
 	@Override
 	public void deactivate() {
-		if (isActive()) {
```
mx990, line 431: Should we also remove the check in `UntypedSubAppInterfaceElementEditPart.deactivate()` then?

azoitl: yes we should

azoitl: although this check does not really harm. I removed it only from the files I anyhow worked on.

mx990 (approved):

PR #1664, opened 2025-09-30
Changed jumping over structs to stop if connection is visible

sebHollersbacher:


azoitl (commented):
In my tests the jump still does not stop consistently at the expected pins.

sebHollersbacher:
@azoitl 
Can you explain which cases do not work?

azoitl:
As I was not sure I retested. The problem seems to be when the connection chain is going through an untyped subpp

sebHollersbacher:
Should work now

azoitl (approved):
works now as expected

PR #1676, opened 2025-10-05
First version for splash screen for the 3.0 release

azoitl:
As we are nearing to a 3.0 release I drafted a potential splash screen for the release.


bwiesmayr:
Somehow I find the blue shades of the function blocks and the dark blue background a bit disturbingly disjoint. Especially the tones in the middle are so greyish and don't fit in my opinion. Can we have the main background just in 4diac blue instead?

jwalt:
I agree with bianca. The function blocks in the bottom half send out "cyberspace" vibes, this doesn't fit with 4diac. I also agree the gradient in the middle looks badly done. In addition, I find the font choice for "3.0" dubious, that should be a sans-serif font to better match the overall style.

azoitl:
Thanks for the feeback. I wanted to FBs to have a kind of neon glow. I may overdid it here. Will look how to improve that. The grey middlepart is my attempt to let the FBs fade into a infinite fog. Unfortunately as soon as you make the 4diac blue a bit more transparent it gets very grayish. Will see what I can do here because without it would look much worse. 

@jwalt the font is a sans serif font. I experimented with Tex Gyre Heros or Roboto. If you only looked at the SVG you may have the fonts not installed. As you can see in the png the intent is definitely a different than what you see. The png is also what the users will see. I just added the SVG for beeing able to make adjustments and improvements in the future (e.g., 3.0.1)

azoitl:
I found some time today to play a bit with some settings and colors. I found out why It was so grey in the middle. I had some wrong colors in the fog layer. Also I made the background a bit lighter and the grid color more aligned with Eclipse 4diac colors. I hope this is what you had in mind.

oberlehner (approved):
I like it :)

jwalt (approved):
yes, that's much better. and you are right, the png shows a proper font.

mx990 (approved):
I also really like the new colors, particularly the three orange highlights.

bwiesmayr:
yes, that fixes it. The grey color there was really out of color scheme. And I agree with Martin, that the highlighted FBs integrate the bottom nicely with the surrounding

moniwe (approved):

MartinMelikMerkumians:
I like the general design, but the "3.0" feels a bit out of place. It is somehow detached from the overall design, font and sizewise, but also the location. 

Not sure what I would change, but probably would make it a nuance bigger and play around with a font to be more in line with the one of the 4diac lettering

azoitl:
Thx for the feedback I increased the size of the 3.0 a bit and changed the dot to a square I think that fits now better.

If you agree I would merge it for now. The SVG sources are part of the commit so we can still tune till the final release if you want to.

MartinMelikMerkumians (approved):

PR #1679, opened 2025-10-07
Add implicit imports and resolve imports in importer

mx990:


azoitl (approved):

franz-hoepfinger-4diac:
for me this works in 4diac-ide_3.0.0-win32.win32.x86_64_nightly_2025-10-13_0604 the way i think it was implemented. 
Open a SUB: no more Error. 
Saving a SUB: adding the Namespace to each FB. 

BUT: i cannot "just save" i need at least e.g. add a Space in the Comment for example (and maybe delete it again) and then press save. 
just pressing save does let the IDE think, it is no change.

PR #1687, opened 2025-10-10
Do not update mapped resource FBs

azoitl:
Dependency updates where done on both in the applications and in the resources. Mapped FBs where considered twice. This could lead to strange situations in the model. Now we are only updating resource FBs if they are not mapped.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
+++ b/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
@@ -178,8 +178,14 @@ public class TypeEntryAdapter extends AbstractTypeEntryAdapter {
 					if (fbnEl instanceof final FB fb && fbnEl.eContainer() == editedElement) {
 						return new UpdateInternalFBCommand(fb, typeEntry);
 					}
+
+					if (fbnEl.isMapped() && fbnEl.getMapping().getTo() == fbnEl) {
+						// the resource side will be update by its opposite so we do not need to do it
+						// here
+						return null;
+					}
 					return new UpdateFBTypeCommand(fbnEl, typeEntry);
-				}).forEach(Command::execute);
+				}).filter(Objects::nonNull).filter(Command::canExecute).forEach(Command::execute);
 
 		if (isActiveEditor()) {
 			performLocationRestore();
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -178,8 +179,14 @@ private void handleBlockTypeDependencyUpdate(final LibraryElement editedElement,
...
 					}
-					return new UpdateFBTypeCommand(fbnEl, typeEntry);
-				}).forEach(Command::execute);
+
+					if (fbnEl.isMapped() && !(fbnEl.eContainer().eContainer() instanceof Resource)) {
+						// the resource side will be update by its opposite so we do not need to do it
+						// here
+						return new UpdateFBTypeCommand(fbnEl, typeEntry);
+					}
+					return null;
```
mx990: I believe this is somewhat backwards now? What about unmapped FBs in the application or resource? I would also suggest to use `FBNetworkElement.getResource()` instead of `eContainer().eContainer()`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: Yes you are right I took the wrong turn here. Fixed now.

No we cannot use getResource here. For mapped fbs it will always return a resource.

mx990: > No we cannot use getResource here. For mapped fbs it will always return a resource.

You are right. I missed that it does a bit too much in this case. In that case, how about the following:
```java
if (fbnEl.isMapped() && fbnEl.getMapping().getTo() == fbnEl) {
```

azoitl: Having `getResource` as it is made many places nicer. So even it seems a bit strange it would like to keep it.

Your proposal took me a bit to read but in the end it makes a few things clearer and as you point out more flexible.

mx990 (approved):

PR #1690, opened 2025-10-12
Notify from operation history changes in the display thread

azoitl:
Several CommandStackEventListeners expect that they where called from the display thread.


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/commands/OperationHistoryCommandStack.java
```diff
@@ -319,7 +320,7 @@ public void historyNotification(final OperationHistoryEvent event) {
 
 			final int state = getCommandStackState(event);
 			if (state != 0) {
-				notifyListeners(getCommand(event.getOperation()), state);
+				Display.getDefault().syncExec(() -> notifyListeners(getCommand(event.getOperation()), state));
```
mx990: Should we use `Display.execute(Runnable)` instead? I am worried that we may incur deadlocks if the display thread is already waiting on the current thread.

azoitl: I was not sure if any of the consumers or the producers expect a sync execution. However as we had deadlocks here and there I would go with execute. Changed accordingly.

mx990 (approved):

PR #1691, opened 2025-10-13
Add copy refactoring to system explorer

Felix-Schmid:
* the package name is now updated, when types are copied to new locations via the system explorer
* a new "copy package name" action is available when right clicking types in the system explorer


plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/FordiacCopyProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+		// actual copy/paste is handled by FordiacCopyAction & FordiacPasteAction
+		// all further modifications are done via CopyParticipants
+		return new NullChange();
+	}
+
+	@Override
+	public RefactoringParticipant[] loadParticipants(final RefactoringStatus status,
+			final SharableParticipants sharedParticipants) throws CoreException {
+		return ParticipantManager.loadCopyParticipants(status, this, files, new CopyArguments(destination, log),
+				new String[] { SystemManager.FORDIAC_PROJECT_NATURE_ID }, sharedParticipants);
```
mx990: Please have a look at [^1] and [^2] on how to load participants for multiple resources, so that it works with existing participants (i.e., Xtext).

[^1]: `org.eclipse.ltk.internal.core.refactoring.resource.MoveResourcesProcessor.loadParticipants(RefactoringStatus, SharableParticipants)`
[^2]: `org.eclipse.ltk.core.refactoring.participants.ParticipantManager.loadCopyParticipants(RefactoringStatus, RefactoringProcessor, Object, CopyArguments, String[], SharableParticipants)`

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/FordiacCopyProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Felix Schmid
+ *     - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+package org.eclipse.fordiac.ide.systemmanagement.ui.systemexplorer;
```
mx990: I would suggest moving this to `org.eclipse.fordiac.ide.typemanagement.refactoring.copy`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	@Override
+	public RefactoringStatus isValid(final LibraryElement element, final IProgressMonitor pm)
+			throws CoreException, OperationCanceledException {
+		final RefactoringStatus status = new RefactoringStatus();
+		if (!Objects.equals(PackageNameHelper.getPackageName(element), oldPackageName)) {
+			status.addFatalError(Messages.MoveTypeToPackage_NameChanged);
+		}
+		if (oldPackageName.equals(newPackageName)) {
```
oberlehner: Use Objects.equals(oldPackageName, newPackageName) to avoid NPE.

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/SystemExplorerPasteAction.java
```diff
@@ -0,0 +1,309 @@
...
+			if (resourceData[0].getType() == IResource.PROJECT) {
+				// enablement checks for all projects
+				for (final IResource resource : resourceData) {
+					final CopyProjectOperation operation = new CopyProjectOperation(shell);
+					operation.copyProject((IProject) resource);
+				}
+			} else {
+				// enablement should ensure that we always have access to a container
+				final IContainer container = getContainer(resourceData);
+				final CopyFilesAndFoldersOperation operation = new CopyFilesAndFoldersOperation(shell);
```
oberlehner: @azoitl should we set an UndoManager here ?
i am not sure how we can integrate that in the current undoActionHistory

azoitl: yes we should do that with an operation history so that it gets syncronized with our editors

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
Felix-Schmid: Done. I fixed the same issue in MoveTypeChange, which this class was based on.

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/SystemExplorerPasteAction.java
mx990: The copying is also supposed to be done during the refactoring, see `org.eclipse.jdt.internal.corext.refactoring.changes.CopyResourceChange`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.fordiac.ide.model.helpers.PackageNameHelper;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
+import org.eclipse.fordiac.ide.typemanagement.Messages;
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.RefactoringStatus;
+import org.eclipse.ltk.core.refactoring.participants.CheckConditionsContext;
+import org.eclipse.ltk.core.refactoring.participants.CopyParticipant;
+
+public class CopyTypeParticipant extends CopyParticipant {
```
mx990, line 39: If we already use the `CopyTypeChange` in the processor, why do we need a separate participant?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
oberlehner: i think we should remove it from the processor and use it only here.
so we can add new participants more easily in the future, and dont have to maintain the processor
we should let the processor only execute the participant.
What do you think?

mx990: I agree. The copy processor then only copies the file(s) and the participant handles the type updates. That should also offer a good upgrade path in case ltk adds an upstream copy processor one day.

Felix-Schmid: I am sorry, but I can’t quite follow here. I don’t see where the `CopyTypeChange` is being used in the processor. The processor loads the participants, which in turn create the `CopyTypeChange`. This is already how it is implemented, isn’t it?

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/FordiacCopyProcessor.java
Felix-Schmid: Thanks for the hint, I was wondering how to correctly handle multiple resources.

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/FordiacCopyProcessor.java
Felix-Schmid: Done.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
mx990: I remember seeing the `CopyTypeChange` in the processor, but I may have also been mistaken.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -85,12 +77,9 @@ public RefactoringStatus checkConditions(final IProgressMonitor pm, final CheckC
 
 	@Override
 	public Change createChange(final IProgressMonitor pm) throws CoreException, OperationCanceledException {
-		final CompositeChange compChange = new CompositeChange(Messages.CopyTypeToPackage_CompositeChangeName);
-		for (final URI uri : destinationURI) {
-			if (TypeLibraryManager.INSTANCE.getTypeEntryForURI(uri) != null) {
-				compChange.add(new CopyTypeChange(newPackageName, getName(), uri));
-			}
+		if (TypeLibraryManager.INSTANCE.getTypeEntryForURI(destinationURI) != null) {
```
mx990: I would suggest moving this check to `initialize(...)`, so that we skip the participant entirely if we have no entry for the file.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
```diff
@@ -66,8 +71,16 @@ public Change createChange(final IProgressMonitor pm) throws CoreException, Oper
 	@Override
 	public RefactoringParticipant[] loadParticipants(final RefactoringStatus status,
 			final SharableParticipants sharedParticipants) throws CoreException {
-		return ParticipantManager.loadCopyParticipants(status, this, files, new CopyArguments(destination, log),
-				new String[] { SystemManager.FORDIAC_PROJECT_NATURE_ID }, sharedParticipants);
+		final String[] affectedNatures = new String[] { SystemManager.FORDIAC_PROJECT_NATURE_ID };
```
mx990: Please also include the `XtextProjectHelper.NATURE_ID` or just use `SystemManager.getNatureIDs()`

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
Felix-Schmid: Done.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
Felix-Schmid: True, if we can get a type entry for the original file, we will also be able to get it for the copy.

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/SystemExplorerPasteAction.java
Felix-Schmid: If I move the actual copy operation from the paste action to the processor via a bunch of `CopyResourceChange`s in the processors `createChange` method, then the copies do not exist as files yet at the time the participant changes are created. The `AbstractCommandChange` only tries to get the type entry in its constructor, but it does not exist yet, because the copying has yet to take place.
Should I change the AbstractCommandChange, so that it retries to get the type entry when performing the change, or do you have better suggestions for solving this?

mx990: I believe the `AbstractCommandChange` is not really a good fit here. Even if we were to retry getting the type entry when performing the change, it would still fail earlier during `isValid` when it cannot find the type. Fortunately, we also don't need most of its features in this case. Since copying creates a fresh type, we can be sure that no editor is open for it, yet. We also don't really need to perform any undo, since undoing the copying will just delete the type again.

I would thus suggest to create a simplified implementation that just loads the type via a resource, changes the name, and saves it again. I would also advise against using a type entry to load the type, since that would lead to an intermittent error due to having two types with the same name.

oberlehner:
please remove this commit and make a rebase without a merge commit
[Merge branch 'develop' into copy-refactoring](https://github.com/eclipse-4diac/4diac-ide/pull/1691/commits/4b1ef525fa9a49e21ee3c971d7721e5b5341ae6d)
[4b1ef52](https://github.com/eclipse-4diac/4diac-ide/pull/1691/commits/4b1ef525fa9a49e21ee3c971d7721e5b5341ae6d)

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,112 @@
...
+	@Override
+	protected boolean initialize(final Object element) {
+		if (element instanceof final IFile file && getArguments().getDestination() instanceof final IPath destination) {
+			final Optional<Resource> resource = getResource(file);
+			if (resource.isEmpty()) {
+				return false;
+			}
+			if (getLibraryElement(resource.get()).isEmpty()) {
+				return false;
+			}
```
mx990: I would suggest to check if there is a type entry for `file`, instead of loading a resource here just to check if it contains a library element.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,112 @@
...
+
+			origin = file;
+			destinationURI = URI.createPlatformResourceURI(destination.toString(), true);
+			final Path filePath = new Path(destinationURI.toPlatformString(true));
+			final IFile newFile = ResourcesPlugin.getWorkspace().getRoot().getFile(filePath);
+			newPackageName = PackageNameHelper.getPackageNameFromFile(newFile);
```
mx990:
```suggestion
			newPackageName = PackageNameHelper.getPackageNameFromURI(destinationURI);
```

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,112 @@
...
+	@Override
+	public String getName() {
+		return MessageFormat.format(Messages.MoveTypeToPackage_RenamePackageTo, newPackageName);
+	}
+
+	@Override
+	public RefactoringStatus checkConditions(final IProgressMonitor pm, final CheckConditionsContext context)
+			throws OperationCanceledException {
+		final RefactoringStatus status = new RefactoringStatus();
+		if (!(getArguments().getDestination() instanceof IPath)) {
```
mx990: We should not get here in this case, since `initialize(...)` would have returned false.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
```diff
@@ -0,0 +1,260 @@
...
+	private ExistsResolve queryUserOverwrite(final IResource file) {
+		final StringBuilder msg = new StringBuilder(Messages.Copy_OverwriteDialog_Message);
+		msg.append(System.lineSeparator()).append(System.lineSeparator());
+		msg.append(Messages.Copy_OverwriteDialog_Destination).append(destination.getLocation().append(file.getName()));
+		msg.append(System.lineSeparator()).append(System.lineSeparator());
+		msg.append(Messages.Copy_OverwriteDialog_Source).append(file.getLocation());
+
+		final int result = MessageDialog.open(MessageDialog.WARNING, shell, Messages.Copy_OverwriteDialog_Title,
+				msg.toString(), SWT.NONE, Messages.Copy_OverwriteDialog_Yes, Messages.Copy_OverwriteDialog_No,
+				Messages.Copy_OverwriteDialog_Rename, Messages.Copy_OverwriteDialog_Cancel);
```
mx990: The refactoring may not always be running in a UI thread, so you will have use `Display.syncExec(Runnable)`, see also `ReorgQueries.YesYesToAllNoNoToAllQuery.confirm(String)`.

I would also suggest encapsulating this into a separate class in a way that avoids UI dependencies in the refactoring itself. That would also enable running the refactoring entirely without a UI or user interaction.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
```diff
@@ -0,0 +1,260 @@
...
+
+			if (resolve == ExistsResolve.RENAME) {
+				copies[i] = destination.getFullPath().append(resolve.getNewName());
+			} else if (resolve == ExistsResolve.OVERWRITE) {
+				copies[i] = destination.getFullPath().append(file.getName());
+			} else if (resolve == ExistsResolve.CANCEL_ALL) {
+				canceled = true;
+				break;
+			}
+		}
```
mx990, line 85: I would suggest to perform these checks in `checkFinalConditions` below, similar to the Java copy refactoring, see also `JavaCopyProcessor.checkFinalConditions(...)`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,103 @@
...
+	@Override
+	public String getName() {
+		return name;
+	}
+
+	@Override
+	public Change perform(final IProgressMonitor pm) throws CoreException {
+		final var resource = CopyTypeParticipant.getResource(destination);
+		if (resource.isEmpty()) {
+			return null;
```
mx990: Please throw an appropriate exception if the resource cannot be loaded to abort the refactoring, e.g. https://github.com/eclipse-4diac/4diac-ide/blob/27baaf1560b8af1e587c4398f2ff7da87f6d2dc8/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AbstractCommandChange.java#L159-L163

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,103 @@
...
+		final var optElement = CopyTypeParticipant.getLibraryElement(resource.get());
+		if (optElement.isEmpty()) {
+			return null;
+		}
+		final LibraryElement element = optElement.get();
+
+		final String fileName = destination.trimFileExtension().lastSegment();
+		if (!element.getName().equals(fileName)) {
+			element.setName(fileName);
+		}
```
mx990: Please use `TypeEntry.getTypeNameFromFileName(String)` when determining the type name according to the file name. This already takes care of trimming the file extension.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,103 @@
...
+		if (element.getCompilerInfo() == null) {
+			element.setCompilerInfo(LibraryElementFactory.eINSTANCE.createCompilerInfo());
+		}
+		element.getCompilerInfo().setPackageName(newPackageName);
```
mx990: Please use `PackageNameHelper.setPackageName(LibraryElement, String)` to set the package name. This already tales care of creating the compiler info if necessary.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,103 @@
...
+		try {
+			resource.get().save(Map.of());
+		} catch (final IOException e) {
+			FordiacLogHelper.logError("CopyTypeChange could not save resource", e); //$NON-NLS-1$
```
mx990: Please wrap and re-throw the exception to give a proper error to the user and abort the refactoring.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeChange.java
```diff
@@ -0,0 +1,103 @@
...
+		}
+		return null; // no undo change necessary, the element will be deleted
+	}
+
+	@Override
+	public Object getModifiedElement() {
+		return CopyTypeParticipant.getResource(destination);
+	}
```
mx990, line 114: Please also overwrite `getAffectedObjects()` , for an example see also https://github.com/eclipse-4diac/4diac-ide/blob/27baaf1560b8af1e587c4398f2ff7da87f6d2dc8/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/AbstractCommandChange.java#L289-L296

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	@Override
+	public Change createChange(final IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		return new CopyTypeChange(newPackageName, getName(), origin, destinationURI);
+	}
+
+	public static Optional<Resource> getResource(final IFile typeFile) {
+		return getResource(URI.createPlatformResourceURI(typeFile.getFullPath().toString(), true));
+	}
```
mx990: I believe this is unused now?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+		if (resource == null || !resource.getErrors().isEmpty()) {
+			return Optional.empty();
+		}
+		return Optional.of(resource);
+	}
+
+	public static Optional<LibraryElement> getLibraryElement(final Resource resource) {
+		return resource.getContents().stream().filter(LibraryElement.class::isInstance).map(LibraryElement.class::cast)
+				.findFirst();
+	}
```
mx990: I believe this can be moved to `CopyTypeChange` now?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
```diff
@@ -0,0 +1,191 @@
...
+
+		final String[] affectedNatures = SystemManager.getNatureIDs();
+		final List<CopyParticipant> result = new ArrayList<>();
+
+		for (int i = 0; i < files.length; i++) {
+			if (copies[i] == null) {
+				continue;
+			}
+			final CopyParticipant[] participants = ParticipantManager.loadCopyParticipants(status, this, files[i],
+					new CopyArguments(copies[i], log), affectedNatures, sharedParticipants);
```
mx990: It seems the copy arguments are expected to contain either an `IFolder` or `IProject` to which the current file name is appended[^1] and not a full destination `IPath`. This also means that simultaneously copying and renaming a file is not supported and will lead to errors for both functions and global constants.

[^1]: see also https://github.com/eclipse-xtext/xtext/blob/7a063e7e3aea557d95d31149febdeb25fcb05b72/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/XtextCopyResourceParticipant.java#L74-L76

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/FordiacCopyProcessor.java
Felix-Schmid: I see. A bit unfortunate from a user perspective, but I now removed the rename during copying feature.

mx990 (approved):
I have created a separate PR https://github.com/eclipse-4diac/4diac-ide/pull/1857 that adds a copy participant for handling ST sources. There is still a problem in Xtext that needs to be resolved for that to work, so it is still a draft for now.

Without the other PR, the copy refactoring will simply not update package declarations of functions and global constants in ST sources. As this matches the current behavior, I believe we can merge this PR independently.

azoitl (approved):

azoitl:
@Felix-Schmid could you please rebase your commits and resolve the merge conflict.

azoitl (approved):

azoitl:
I rebased for you once more to hopefully get the test fixes in.

PR #1693, opened 2025-10-13
Changed background of error/warning cells in NatTable when editing

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java
```diff
@@ -460,6 +460,11 @@ public void configureRegistry(final IConfigRegistry configRegistry) {
 				configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.SELECT,
 						ERROR_CELL);
 
+				cellStyle = new Style();
+				cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR, GUIHelper.COLOR_WHITE);
```
mx990, line 464: Can we use the default background color instead of forcing it to white (e.g., for dark mode users)?

sebHollersbacher: I noticed that NatTable Styling does not consider the Theme at all, so the whole styling should be reworked to also work for dark mode

mx990: Makes sense. Thanks for clarifying.

mx990 (approved):

PR #1697, opened 2025-10-13
LibraryManager on UpdateDependencie Actively Refreshes TypeLib

azoitl:
There where situations where new libs added by the LibraryManager where not directly visible to the xtext builder.

As part of this change clean-ups in the TypeLib for better creating new entries and outdated code in the SystemManager where done.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
@@ -197,35 +197,6 @@ public final class TypeLibrary {
 		return fileMap.get(typeFile);
 	}
 
-	/**
-	 * Clear and reload the entire type library
-	 *
-	 * <p>
-	 * This is a dangerous operation, since it will effectively <em>detach</em>
-	 * anyone currently holding a type entry from this type library, including any
-	 * listeners. For most use cases, use {@link #refresh()} instead.
-	 *
-	 * @see #refresh()
-	 */
-	private void reload() {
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -251,16 +222,23 @@ public TypeEntry createTypeEntry(final IFile file) {
 		if (BuildpathUtil.findSourceFolder(buildpath, file).isEmpty()) {
 			return null;
 		}
-		final TypeEntry entry = fileMap.computeIfAbsent(file, TypeEntryFactory.INSTANCE::createTypeEntry);
+
+		TypeEntry entry = fileMap.get(file);
```
mx990: Separating the `get(...)` here and `put(...)` in `addTypeEntry(...)` below may lead to race conditions where two distinct entries are created for the same file.

I would suggest performing the remainder of this method inside the `computeIfAbsent(...)` but instead of calling `addTypeEntry(...)` to extract only the relevant portions, that is:
```java
entry.setTypeLibrary(this);
addTypeEntryNameReference(entry);
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
azoitl: good point. Done.

mx990 (approved):

PR #1700, opened 2025-10-14
Validate nature when project file changes

MandKastner:


plugins/org.eclipse.fordiac.ide.systemmanagement/src/org/eclipse/fordiac/ide/systemmanagement/changelistener/FordiacResourceChangeListener.java
```diff
@@ -140,6 +140,10 @@ private static boolean handleResourceChanged(final IResourceDelta delta) {
 			}
 			break;
 		case IResource.PROJECT:
+			if (testFlags(delta, IResourceDelta.DESCRIPTION)) {
```
azoitl, line 143: why du you tst for description and not for content like the file case above?

MandKastner: If i check the project resource the description flag should be correct, because the description also changes when builders, natures, etc .. are modified.  At this place the content flag also does not work. I could use it in the case above, but i would have to check the filename additionally.

oberlehner (approved):
Tested in UI and it works

plugins/org.eclipse.fordiac.ide.systemmanagement/src/org/eclipse/fordiac/ide/systemmanagement/changelistener/FordiacResourceChangeListener.java
azoitl: thx for the clarification. resource change listeners and projects are always special.

PR #1703, opened 2025-10-14
Fix missed POST_EXECUTE when operation history is full

mx990:
When the operation history is full, it removes the oldest operation before adding the current one. This causes an OPERATION_REMOVED event to arrive in the OperationHistoryListener, which set the prevState and prevOperation to the removed operation. The subsequent DONE event was then ignored.

This adds a condition to avoid updating the prevState and prevOperation for events about added, changed, or removed operations.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/commands/OperationHistoryCommandStack.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/commands/OperationHistoryCommandStack.java
@@ -322,10 +322,31 @@ public class OperationHistoryCommandStack extends CommandStack {
 			if (state != 0) {
 				Display.getDefault().execute(() -> notifyListeners(getCommand(event.getOperation()), state));
 			}
-			prevState = event.getEventType();
-			prevOperation = event.getOperation();
+			// do not update prevState and prevOperation for events about added, changed, or
+			// removed operations, in order to avoid false updates due to intermittent
+			// events, for example, when operations are removed during execute due to the
+			// history being full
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/commands/OperationHistoryCommandStack.java
```diff
@@ -322,10 +322,20 @@ public void historyNotification(final OperationHistoryEvent event) {
...
 			}
-			prevState = event.getEventType();
-			prevOperation = event.getOperation();
+			if (!isOperationEvent(event)) {
+				prevState = event.getEventType();
+				prevOperation = event.getOperation();
+			}
 		}
 
+		private boolean isOperationEvent(final OperationHistoryEvent event) {
```
oberlehner, line 342: this should be documented or renamed.
Otherwise no one will know in the future what this code does.

mx990: I'm open to suggestions on how to rename this.

oberlehner: i dont know a good name in that case, maybe a comment would be better that explains that. similar to the text in the PR

mx990: done

azoitl (approved):

PR #1704, opened 2025-10-14
Remove unqualified lookup during type update

mx990:


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/UpdateFBTypeCommand.java
```diff
@@ -94,11 +92,6 @@ private static TypeEntry reloadTypeEntry(final TypeEntry entry, final FBNetworkE
 			return reloadedTypeEntry;
 		}
 
-		final List<TypeEntry> candidates = typeLibrary.findUnqualified(entry.getTypeName());
```
oberlehner, line 97: in general i agree that we should remove that. but i have the fear that this might break old projects

mx990: How would this be *breaking* old projects? This is only used for type *updates*.

oberlehner: you are right. i missd the context

oberlehner (approved):

PR #1709, opened 2025-10-14
Removed dedicated xbase.lib version

azoitl:
We had different xbase.lib versions in our manifests.

Furthermore added asciidoc option to avoid errors in documenation generation.


plugins/org.eclipse.fordiac.ide.contractspec.tests/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,7 @@ Bundle-ActivationPolicy: lazy
 Require-Bundle: org.eclipse.fordiac.ide.contractspec,
  org.eclipse.xtext.testing,
  org.eclipse.xtext.xbase.testing,
- org.eclipse.xtext.xbase.lib;bundle-version="2.37.0"
```
mx990, line 12: The versions for `org.eclipse.xtext.xbase.lib` are (re-)added automatically when running the MWE2 workflow. I would recommend instead changing them all to the current version `2.40.0`.

mx990 (changes requested):

mx990 (approved):

PR #1711, opened 2025-10-15
Validate project nature on startup

MandKastner:


oberlehner:
for me this solution looks good. can you then remove the old checks you made in your previous commits or are they still necessary ?

oberlehner (approved):

azoitl (approved):

PR #1721, opened 2025-10-16
The Pinname for error marker pins had error messages

azoitl:
When a pin of a container (CFB, Subapp) was missing instead of the pin name an error mssage was stored in the errormaker. This lead to broken XML files.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
@@ -56,9 +56,11 @@ public final class ConnectionHelper {
 		final FBNetworkImporter importer;
 
 		private final String destinationString;
+		private String fullDstPinName;
 		private InterfaceList destInterfaceList;
 
 		private final String sourceString;
+		private String fullSrcPinName;
 		private InterfaceList srcInterfaceList;
 
 		public ConnectionBuilder(final String sourceString, final String destinationString, final T connection,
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -264,39 +266,35 @@ public boolean dataInputHasMultipleConnections() {
...
-			if (connection.getDestination() != null) {
-				return connection.getDestination().getName();
+			if (ie != null) {
+				return ie.getName();
 			}
 
-			final String[] qualNames = destinationString.split("\\."); //$NON-NLS-1$
-			if (qualNames.length < 2) {
-				return MessageFormat.format(Messages.ConnectionHelper_pin_not_found, destinationString);
+			if (!(il.eContainer() instanceof final BlockFBNetworkElement bfbNEl)) {
```
mx990: I would suggest a null check for `il` here.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -264,39 +266,35 @@ public boolean dataInputHasMultipleConnections() {
...
+				// this is a pin to a parent CFB or SubAppInterface
+				return pinIdentifier;
 			}
 
-			return qualNames[1];
+			if (fullPinName == null) {
+				// we have a missing FB pin
+				return pinIdentifier.substring(bfbNEl.getName().length() + 1); // plus 1 for the dot after the FB name
+			}
+			return fullPinName;
```
mx990: Should we prefer the saved pin name if it is set?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
azoitl: This is the saved pin name.

mx990: Yes, why not put a
```java
if (fullPinName != null) {
  return fullPinName;
}
```
at the top?

azoitl: that is a good point. I reworked the whole method to be less as it was before but more on what data we have available at that point.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
azoitl: good point. As written below reworked it as a whole to better reflect the data we may have available at this point.

mx990 (approved):

PR #1725, opened 2025-10-20
Fixed FB Type Query for Generic FBs

azoitl:


plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
```diff
@@ -413,6 +419,28 @@ public Response queryGlobalConstType(final GlobalConstantsEntry entry) throws De
...
+		final FBType type = fbEntry.getType();
+		final Attribute attribute = type.getAttribute(GENERIC_CLASS_NAME_ATTRIBUTE);
+		if (attribute != null && attribute.getValue() != null && !attribute.getValue().isBlank()) {
+			final String packageName = fbEntry.getPackageName();
+			final String genName = StringValueConverter.INSTANCE.toValue(attribute.getValue());
+			if (packageName.isBlank()) {
+				return genName;
+			}
+			return packageName + PackageNameHelper.PACKAGE_NAME_DELIMITER + genName;
+		}
```
mx990, line 442: Should this not better go into `ForteTypeNameCreator.getForteTypeName(TypeEntry)` (only for FB type entries)?

plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
```diff
@@ -413,6 +419,28 @@ public Response queryGlobalConstType(final GlobalConstantsEntry entry) throws De
...
+	private static String getQueryFBTypeNameWithHash(final FBTypeEntry fbEntry) throws LibraryElementHashException {
+		final String hash = fbEntry.getTypeHash();
+		if (hash.isEmpty()) {
+			return getQueryFBTypeName(fbEntry);
+		}
+		return getQueryFBTypeName(fbEntry) + '#' + hash;
+	}
+
+	private static String getQueryFBTypeName(final FBTypeEntry fbEntry) {
+		final FBType type = fbEntry.getType();
```
mx990, line 431: I would suggest a null check for the type.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
azoitl: In the original design of 4diac FORTE's FB creation interface we hid the implementation detail of generic type names. Therefore for other interactions we do not need that. I'm also not happy that we have now 4diac FORTE implementation details in IDE. But therefore I didn't want to move that code to a general place but keep it here private for now.

plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
azoitl: done

mx990:
I believe I have found a simpler and more generic solution for the problem on the FORTE side, see https://github.com/eclipse-4diac/4diac-forte/pull/673.

azoitl:
@mx990  this was the second option I thought about. Seeing no both and my comment above I'm in strong favor of the forte solution.

PR #1735, opened 2025-10-21
Add null check for Segement's communication in system exporter

azoitl:


oberlehner (approved):

bwiesmayr:
I don't object having a null check but want to point out that, in theory, an empty segment should have a default communication item. So if the communication is null, something broke earlier (again). This could possibly cause some issues when selecting property sheets of Segments (although I believe we have null checks almost everywhere by now)

azoitl:
I thought so too. But as I've seen exactly at that point recently more often NPEs and as we have the null checks in many places I thought the null check is maybe the right solution.

PR #1747, opened 2025-10-23
Fix export with no exported types

mx990:


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/cmake/CMakeListsTemplate.java
```diff
@@ -241,7 +241,11 @@ protected static boolean isStandardLibrary(final Required required) {
 	}
 
 	protected List<String> getSubdirectories() {
-		try (Stream<Path> list = Files.list(getCurrentSourceDir())) {
+		final Path currentSourceDir = getCurrentSourceDir();
+		if (!Files.exists(currentSourceDir)) {
```
azoitl, line 245: should we create the directories when they are not existing? AFAIK in some other places we are doing that.

mx990: How would that help? If we were to create an empty directory here, the resulting list would still be empty.

azoitl: The I don't understand the code and the names of the variables.

mx990: This method is used to list *existing* subdirectories for inclusion into the current `CMakeLists.txt` via `add_subdirectory()`. The `currentSourceDir` is the directory the current template *will be* exported into. If the directory does not exist, there is nothing that needs to be included, same for the other methods. It will already be created as needed when the file for the current template is written, as long as the generation is successful. This means there is no need to create it here.

azoitl: ah thx for the clarification.

azoitl (approved):

PR #1749, opened 2025-10-26
Rework version info to follow IEC 61499-2 rules

### LINKED ISSUE #1746: Latest release broke libraries
Just updated new release. Some reason all libraries are now read only and it looks that all fb:s are version1 even dragging version3 on applicattion?

azoitl:
Version infor should be stored in ascending order in xml files. This PR fixes the 3.0 library elements and the data shown in the FB type editor. New version information is added at the beginning of the list.

The pop-up at FB instances allready used the right version information.

This PR also fixes #1746.


data/typelibrary/convert-3.0.0/typelib/GET_AT_INDEX.fbt
```diff
@@ -2,9 +2,9 @@
 
 <FBType Comment="Fetch value from array at given index. Array data type has to match with output data type." Name="GET_AT_INDEX">
   <Identification Description="Copyright (c) 2014 Profactor GmbH&#13;&#10; &#13;&#10;This program and the accompanying materials are made&#13;&#10;available under the terms of the Eclipse Public License 2.0&#13;&#10;which is available at https://www.eclipse.org/legal/epl-2.0/&#13;&#10;&#13;&#10;SPDX-License-Identifier: EPL-2.0"/>
-  <VersionInfo Author="Matthias Plasch" Date="2014-07-11" Organization="Profactor GmbH" Version="1.0"/>
 	<VersionInfo Version="3.0" Author="Patrick Aigner" Date="2025-04-14" Remarks="changed package">
 	</VersionInfo>
+  <VersionInfo Author="Matthias Plasch" Date="2014-07-11" Organization="Profactor GmbH" Version="1.0"/>
```
ernstblechaPT, line 7: Tiny nitpicking: For this and some other FBs the XML is not exactly what would be exported by 4diac-ide (emtpy tags and identation); It's still valid XML so this should not be a problem for any XML-Parser; maybe some time in the future the FBs can be reexported using the Resave-FB-Ant-Task to make sure that the XML-Formatting is in one format.

ernstblechaPT (approved):
Looks good to me!

mx990 (approved):

data/typelibrary/convert-3.0.0/typelib/GET_AT_INDEX.fbt
azoitl: Yes I noticed that. But given the amount of work I wanted to spend only the most minimal effort here. I also want to check seperatly if the versionInfo can have even children. If not I would change the xml generation to single line tags. But as said that is something for a different day.

oberlehner (approved):

PR #1756, opened 2025-10-27
Optimize annotation model updates

mx990:
Optimize annotation model updates:
- Add refresh on save to remove all state and transient annotations and
add new annotations. This avoids a full reload and separate removal of
transient annotations on save.
- Optimize information stored in annotation model events for fast access
to affected targets and updated annotations per target when dispatching
annotations in the UI thread.
- Run batch operations on the annotation model, such as refresh and
reload, in a background thread to improve responsiveness.
- Dispatch annotation model events in a UI job using an event-loop monitor
that periodically runs the event loop to keep the UI responsive.


azoitl (approved):
For a large number of warnings and errors the update feels a bit strange but performance and responsiveness is much better.

PR #1762, opened 2025-10-28
Fix insert FB command

MandKastner:
The insert FB command targeted the wrong editpart, therfore the insertion into unfolded subapps did not work correctly.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/InsertFB.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/InsertFB.java
@@ -15,6 +15,8 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.handlers;
 
+import java.util.Optional;
+
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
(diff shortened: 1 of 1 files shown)


oberlehner (approved):

azoitl (changes requested):
Unfortunately this does not work. The FBNetwork of expanded subapps is not selectable. Therefore the selection is always the root fbnetwork. 

I guess we need to find the editpart under the mouse and check for that.

oberlehner:
> Unfortunately this does not work. The FBNetwork of expanded subapps is not selectable. Therefore the selection is always the root fbnetwork.
> 
> I guess we need to find the editpart under the mouse and check for that.

For me it works.

I also tested again on 3.0.x -> here i was able to reproduce the bug.

With this PR it was working fine.
i also tested it  with multiple nested expanded subapps and it was always inserted in the correct hieracchy level

azoitl:
I did the same. However it looks like this is a Linux problem. On Linux on right click the pop-up menu is raised but no selection is performed in GEF. This is a general problem of either SWT, GEF Classic, or some code deeper in our GEF Classic adaptation. So we can merge this and I have to check where and how to fix the root problem in Linux.

azoitl (approved):

mx990:
I wanted to check if I have to be aware of the problem, so I just tried it on my Linux machine. However, it also seemed to work for me.

I tried the following steps:
1. select an FB instance in the application,
2. right click inside an expanded subapp and choose *Insert FB*,
3. the edit part returned by `InsertFB.getTargetEditPart(IWorkbenchPart, GraphicalViewer)` is an `UnfoldedSubappContentEditPart` and the new FB is inserted correctly.

I also tested with the 3.0.x branch prior to this PR for comparison and there I got an `FBNetworkEditPart` and the new FB is inserted in the application instead.

azoitl:
Given your feedback @mx990 I tested again. I can confirm that on my machine it now also mostly works. Not sure why I had issues in the first place.

PR #1779, opened 2025-11-03
Add git issue reporting feature

Felix-Schmid:
Documentation can be found here: https://github.com/eclipse-4diac/4diac-documentation/pull/76

### DIFF
--- a/plugins/org.eclipse.fordiac.ide/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide/plugin.properties
@@ -49,6 +49,7 @@ IssueReporting_AutoReport=Automatic reporting
 IssueReporting_DestinationGroupLabel=Issue Reporting Destination
 IssueReporting_GitLab=GitLab
 IssueReporting_GitHub=GitHub
+IssueReporting_GitHub_Manual=GitHub (Manual Submit)
 IssueReporting_URLInput=URL:
 IssueReporting_TokenInput=Access Token:
 IssueReporting_ProjectPathInput=Project Path:
--- a/plugins/org.eclipse.fordiac.ide/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide/plugin.properties
@@ -38,7 +38,7 @@ product.name = 4diac IDE
 
 FordiacLogListener_ErrorDialogRestartSave=We recommend to save your changes and restart 4diac IDE.
 FordiacLogListener_ErrorDialogReportPrompt=Please report this issue by submitting the details below\!
-FordiacLogListener_ErrorDialogAutoReportInfo=This issue will be reported by submitting the details below.
+FordiacLogListener_ErrorDialogAutoReportInfo=This issue has been reported by submitting the details below.
 FordiacLogListener_ErrorDialogTitle=Major Error Occurred\!
 FordiacLogListener_ErrorDialogReportIssue=Report Issue
 
--- a/plugins/org.eclipse.fordiac.ide/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide/plugin.properties
@@ -36,6 +36,20 @@ introQuicklink.New4diacIDEExampleProject.description = Create a new 4diac IDE IE
 perspective.name = System
 product.name = 4diac IDE
 
-FordiacLogListener_ErrorDialogDescription=We recommend to save your changes and restart 4diac IDE.\n\nPlease report this issue by submitting the details below\!
+FordiacLogListener_ErrorDialogRestartSave=We recommend to save your changes and restart 4diac IDE.
+FordiacLogListener_ErrorDialogReportPrompt=Please report this issue by submitting the details below\!
+FordiacLogListener_ErrorDialogAutoReportInfo=This issue will be reported by submitting the details below.
 FordiacLogListener_ErrorDialogTitle=Major Error Occurred\!
-FordiacLogListener_ReportIssue=Report Issue
+FordiacLogListener_ErrorDialogReportIssue=Report Issue
+
+IssueReporting_ModeGroupLabel=Issue Reporting Mode
+IssueReporting_NoReport=No issue reporting
+IssueReporting_PromptReport=Prompt reporting
+IssueReporting_AutoReport=Automatic reporting
+IssueReporting_DestinationGroupLabel=Issue Reporting Destination
+IssueReporting_GitLab=GitLab
+IssueReporting_GitHub=GitHub
+IssueReporting_URLInput=URL:
+IssueReporting_TokenInput=Access Token:
+IssueReporting_ProjectPathInput=Project Path:
+IssueReporting_ProjectPathToolTip=Usually in the format <owner>/<project name>.
--- a/plugins/org.eclipse.fordiac.ide/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide/plugin.properties
@@ -38,7 +38,10 @@ product.name = 4diac IDE
 
 FordiacLogListener_ErrorDialogRestartSave=We recommend to save your changes and restart 4diac IDE.
 FordiacLogListener_ErrorDialogReportPrompt=Please report this issue by submitting the details below\!
-FordiacLogListener_ErrorDialogAutoReportInfo=This issue has been reported by submitting the details below.
+FordiacLogListener_ReportInProgress=Reporting...
+FordiacLogListener_BrowserOpened=A browser window to report the issue has been opened.
+FordiacLogListener_ReportingError=An error occurred while trying to report this issue.
+FordiacLogListener_ReportingResult=This issue has been reported <a href=\"%s\">here</a>.
 FordiacLogListener_ErrorDialogTitle=Major Error Occurred\!
 FordiacLogListener_ErrorDialogReportIssue=Report Issue
 
--- a/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
+++ b/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
@@ -14,12 +14,6 @@ package org.eclipse.fordiac.ide;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
-import java.text.MessageFormat;
 import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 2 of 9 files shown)


plugins/org.eclipse.fordiac.ide/plugin.properties
```diff
@@ -36,6 +36,8 @@ introQuicklink.New4diacIDEExampleProject.description = Create a new 4diac IDE IE
 perspective.name = System
 product.name = 4diac IDE
 
-FordiacLogListener_ErrorDialogDescription=We recommend to save your changes and restart 4diac IDE.\n\nPlease report this issue by submitting the details below\!
+FordiacLogListener_ErrorDialogRestartSave=We recommend to save your changes and restart 4diac IDE.
+FordiacLogListener_ErrorDialogReportPrompt=Please report this issue by submitting the details below\!
+FordiacLogListener_ErrorDialogAutoReportInfo=This issue will be reported by submitting the details below.
```
oberlehner: This issue has been reported

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitIssueReportingPage.java
```diff
@@ -0,0 +1,243 @@
...
+		gitLabURLText = createFieldInput(group, isGitLab, PreferenceConstants.P_BUG_REPORT_GITLAB_URL);
+
+		createFieldLabel(group, Messages.IssueReporting_ProjectPathInput);
+		gitLabProjectPathText = createFieldInput(group, isGitLab, PreferenceConstants.P_BUG_REPORT_GITLAB_PROJECT_PATH);
+		gitLabProjectPathText.setToolTipText(Messages.IssueReporting_ProjectPathToolTip);
+
+		createFieldLabel(group, Messages.IssueReporting_TokenInput);
+		gitLabTokenText = createFieldInput(group, isGitLab, PreferenceConstants.P_BUG_REPORT_GITLAB_TOKEN);
+		gitLabTokenText.setEchoChar('*');
+
```
oberlehner: please also add the github URL here, since it is possible to install a self hosted github

plugins/org.eclipse.fordiac.ide/plugin.properties
oberlehner: Please also add the link to the created issue if possible.

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitIssueReportingPage.java
```diff
@@ -0,0 +1,243 @@
...
+		final boolean isGitHub = reportDestination == ReportDestination.GITHUB;
```
oberlehner: Please show github before gitlab in the UI

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitIssueReportingPage.java
Felix-Schmid: Done.

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitIssueReportingPage.java
Felix-Schmid: Done.

plugins/org.eclipse.fordiac.ide/plugin.properties
Felix-Schmid: > Please also add the link to the created issue if possible.

Would be possible, but requires to parse the JSON response and we currently do not have a parser available within the codebase.

oberlehner: In my oppinion we should add a parser.

@azoitl what do you think ? i know you dont like the dependency, but i think at some point we need one.

azoitl: Showing the link to the created issue would be important to gain the trust of our users. However given our experience with integrating json libraries in the past, and the required new dependency I'm a bit reluctant. Is there an easy way to just extract the link? Because a full fledged json parser seems to be a bit of an overkill here. I know I keep that saying for quite some time.

oberlehner:
should we move the implementation to `/org.eclipse.fordiac.ide.gitlab` ?

I think this would fit better here, so we have the URLs for the api calls  collected in one plugin

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -0,0 +1,197 @@
...
+		exception.printStackTrace(new PrintWriter(writer));
+		return writer.toString();
+	}
+
+	private static int createGitHubIssueManual(final IssueInfo info) {
+		try {
+			final IWorkbench wb = PlatformUI.getWorkbench();
+			final IWebBrowser browser = wb.getBrowserSupport().createBrowser(Activator.PLUGIN_ID);
+			final String reportingURI = MessageFormat.format(
+					"https://github.com/eclipse-4diac/4diac-ide/issues/new?title={0}&body={1}", //$NON-NLS-1$
```
oberlehner: please extract as constant

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -0,0 +1,197 @@
...
+		if (info.body().length() > 1048575) { // ~max GitLab description length
+			return -1;
+		}
+
+		final String baseURI = removeLeadingTrailingSlashes(PreferenceConstants.getReportGitLabURL());
+		final String projectPath = removeLeadingTrailingSlashes(PreferenceConstants.getReportGitLabProjectPath());
+		final String accessToken = PreferenceConstants.getReportGitLabToken();
+		final String labels = String.join(",", info.labels()); //$NON-NLS-1$
+
+		final String uri = baseURI + "/api/v4/projects/%s/issues?title=%s&description=%s&labels=%s"; //$NON-NLS-1$
```
oberlehner: i think we already have constants for that in org.eclipse.fordiac.ide.gitlab.management.GitLabDownloader

plugins/org.eclipse.fordiac.ide/plugin.properties
oberlehner: I think in the future, this feature will be extented or we will get similar features.
So in my oppionion i would add a parser. As alternative we can also copy the source code of one, so we have more controll over it.
Maybe `com.google.gson` would be better regarding conflicts compared to the `jackson `parser.

azoitl:
> should we move the implementation to `/org.eclipse.fordiac.ide.gitlab` ?
> 
> I think this would fit better here, so we have the URLs for the api calls collected in one plugin

I was not so sure about that. First of all we get a dependency from the ide plugin to the gitlab plugin, which I'm not so happy about. Also now we have github and gitlab reporting. Therefore I think we should for now keep it in the IDE plugin.

plugins/org.eclipse.fordiac.ide/plugin.properties
azoitl: I think you are right about this. I would prefere if we could take something that is already part of the eclipse bundles. 

I had a quick look at https://download.eclipse.org/releases/2025-09 and there are several jars with json in its name:

  - https://www.eclipse.org/downloads/download.php?file=/releases/2025-09/202509101001/plugins/json_20230227.0.0.jar
  - https://www.eclipse.org/downloads/download.php?file=/releases/2025-09/202509101001/plugins/org.eclipse.json_1.2.0.v202308160419.jar

Not sure if any of these is suite.
 
Also Gson is there: https://www.eclipse.org/downloads/download.php?file=/releases/2025-09/202509101001/plugins/com.google.gson_2.13.1.jar

This should greatly reduce our dependency problem. Because anything there we can use without additional effort. Just by extending our target platform's dependencies.

Felix-Schmid: The newest commit adds a link to the created issue. I chose Gson as parser.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	@SuppressWarnings("nls")
+	private final static String[] LABELS = new String[] { "bug", "autoreport" };
+	private final static String SESSION_ID = UUID.randomUUID().toString();
+
+	public static Optional<String> createIssue(final IStatus status) {
+		final IssueInfo info = new IssueInfo(status.getMessage(), buildBody(status), LABELS);
+
+		final PreferenceConstants.ReportDestination repDest = PreferenceConstants.getReportDestination();
```
azoitl: I personally find a switch case for this kind of code better to read:

```
return switch (PreferenceConstants.getReportDestination()) {
		case PreferenceConstants.ReportDestination.GITLAB -> createGitLabIssue(info);
		case PreferenceConstants.ReportDestination.GITHUB -> createGitHubIssue(info);
		case PreferenceConstants.ReportDestination.GITHUB_MANUAL -> createGitHubIssueManual(info);
		default -> Optional.empty();
		};
```

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -0,0 +1,198 @@
...
+	private static Optional<String> makeRequest(final HttpRequest request) {
+		try {
+			final HttpResponse<String> response = HttpClient.newHttpClient().send(request,
+					HttpResponse.BodyHandlers.ofString());
+
+			if (response.statusCode() != 201) { // 201 - Created
+				return Optional.empty();
+			}
+			final Gson gson = new Gson();
+			final PreferenceConstants.ReportDestination repDest = PreferenceConstants.getReportDestination();
```
azoitl: also here I think switch case is nicer

azoitl (changes requested):

azoitl:
Please also rebase on top of latest 3.0.x branch

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
ernstblechaPT: This feels almost like the distinction between Gitlab and Github should be done in the method calling makeRequest and the (reference to the) full response (maybe just response.body()) should be returned as an Optional

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -25,35 +26,72 @@
...
 		}
 
-		private static String getMessage() {
-			final StringBuilder sb = new StringBuilder();
-			sb.append(Messages.FordiacLogListener_ErrorDialogRestartSave);
-			sb.append(System.lineSeparator());
-			sb.append(System.lineSeparator());
+		@Override
+		protected Control createDialogArea(final Composite parent) {
+			final Control control = super.createDialogArea(parent);
```
ernstblechaPT, line 56: Not sure if I fully understand this: This is a dialog that can be closed by the user before the reporting has finished?

Does this still work if the user closes the dialog and a lot later your code tries to display a message (e.g. very slow response from GitHub or GitLab?)

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -84,24 +108,21 @@ private static String getStackTrace(final Throwable exception) {
...
 	}
 
-	private static int createGitHubIssueManual(final IssueInfo info) {
+	private static Optional<String> createGitHubIssueManual(final IssueInfo info) {
 		try {
-			final IWorkbench wb = PlatformUI.getWorkbench();
-			final IWebBrowser browser = wb.getBrowserSupport().createBrowser(Activator.PLUGIN_ID);
-			final String reportingURI = MessageFormat.format(
-					"https://github.com/eclipse-4diac/4diac-ide/issues/new?title={0}&body={1}", //$NON-NLS-1$
+			final String reportingURI = MessageFormat.format(FORDIAC_IDE_ISSUE_URL,
```
ernstblechaPT: What is the reason for using MessageFormat.format here, String.formatted for GitLab and concatination with `+` for GitHub for constructing urls?

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -84,24 +108,21 @@ private static String getStackTrace(final Throwable exception) {
...
 					URLEncoder.encode(info.title(), StandardCharsets.UTF_8), // title
 					URLEncoder.encode(info.body(), StandardCharsets.UTF_8)); // body
-			browser.openURL(new URI(reportingURI).toURL());
-			return 200;
+			openLinkInBrowser(reportingURI);
+			return Optional.empty(); // no issue created yet...
```
ernstblechaPT, line 108: I would have expected the comment to be something along "Issue will be created through direct user interaction in the webbrowser"

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
Felix-Schmid: I agree - done.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
Felix-Schmid: Good catch – I need to check if the label is already disposed. Otherwise yes, the report will work even if the dialog is closed prematurely.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
Felix-Schmid: I harmonized the styles now.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
Felix-Schmid: The return here is only used for the created issue URL (which does not exist yet for manual submitting), but the dialog will read "A browser window to report the issue has been opened.".

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
Felix-Schmid: Good suggestion – done.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -121,18 +120,21 @@ private static Optional<String> createGitLabIssue(final IssueInfo info) {
...
-
 		try {
 			final HttpRequest request = HttpRequest.newBuilder().uri(URI.create(reportingURI))
 					.header("PRIVATE-TOKEN", accessToken) //$NON-NLS-1$
 					.POST(HttpRequest.BodyPublishers.noBody()).build();
-			return makeRequest(request);
+			final Optional<String> body = makeRequest(request);
+			if (body.isEmpty()) {
+				return body;
+			}
```
ernstblechaPT: If you catch the `NoSuchElementException` you can drop this (same for GitHub) and directly call `.get()` on the `makeRequest(...)`-Call

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
```diff
@@ -121,18 +120,21 @@ private static Optional<String> createGitLabIssue(final IssueInfo info) {
...
+			final Gson gson = new Gson();
```
ernstblechaPT: This (`final Gson gson = new Gson();`) should be moved outside the `try`-block as you might hide exceptions from Gson otherwise

ernstblechaPT (approved):

azoitl (approved):

PR #1781, opened 2025-11-03
Support full build in export builder

MandKastner:


plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
mx990, line 53: I believe we are missing the `fct` file ending here. Please also use the constants from `TypeLibraryTags`.

mx990: Why do you create the output folder here? First of all, the exporter creates parent folders as necessary, so this should not be necessary at all. We should also neither modify the workspace when initializing the builder nor create an output folder if the export builder is disabled.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -135,8 +156,7 @@ private void addPreferenceListener() {
 		getProjectPreferenceNode().addPreferenceChangeListener(evt -> {
 			switch (evt.getKey()) {
 			case PreferenceConstants.ENABLE_TYPE_EXPORT:
-				this.enableAutoExport = getProjectPreferenceNode().getBoolean(PreferenceConstants.ENABLE_TYPE_EXPORT,
-						false);
+				initializePreferences();
```
mx990: Why do we need to cache the preferences in the builder at all? It should have no noticeable impact on performance but may be a recurring source of problems.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
mx990: Why are you basing this around `AbstractExporter`? I believe it would be much simpler just using `IExportFilter` directly and exporting each type as you are walking through the deltas.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -211,7 +243,7 @@ protected Exporter(final IConfigurationElement filterConfig, final String output
 
 		@Override
 		protected void showErrorWarningSummary(final IExportFilter filter) {
-			filter.getErrors().stream().forEach(FordiacLogHelper::logError);
+			filter.getErrors().stream().filter(Objects::nonNull).forEach(FordiacLogHelper::logError);
```
mx990: This will show an error dialog asking to restart 4diac IDE whenever a block with an error is saved while the export builder is enabled. Instead of logging an error, we should create appropriate error markers for the user (and remove stale markers when a file is updated or on clean).

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
mx990: same here

mx990: This will always create a folder with the default name, even when exporting to a different folder. However, as mentioned above, this should not even be necessary at all.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -164,15 +184,27 @@ private void addResourceDeltas(final List<IFile> exportees, final IResourceDelta
...
+				.filter(ExportBuilder::getExportAttributeValue).toList();
+	}
+
 	private boolean isExportable(final IFile file) {
-		if (file.getFileExtension() != null && fileTypes.contains(file.getFileExtension())) {
+		if (isExportableFileType(file)) {
+			// if we have performance issues we might want to cache the buildpath and detect
+			// buildpath file changes via resource deltas to reload only on changes, as
+			// loading the buildpath involves file operations
 			final Buildpath buildpath = BuildpathUtil.loadBuildpath(getProject());
```
mx990: I believe it would be better to use the buildpath from the type library via `TypeLibrary.getBuildpath()` or just load it once at the beginning of the build.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -164,15 +184,27 @@ private void addResourceDeltas(final List<IFile> exportees, final IResourceDelta
...
 			final Optional<SourceFolder> sourceFolder = BuildpathUtil.findSourceFolder(buildpath, file);
 			return sourceFolder.isPresent() && getExportAttributeValue(sourceFolder.get());
 		}
 		return false;
 	}
 
+	private static boolean isExportableFileType(final IFile file) {
+		return file.getFileExtension() != null && fileTypes.contains(file.getFileExtension());
```
mx990: Please use a case-insensitive comparison when looking for matching file extensions.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -43,218 +47,197 @@
...
-
-	@Override
-	protected void startupOnInitialize() {
-		createOutputFolder();
-		initializePreferences();
-		addPreferenceListener();
-		super.startupOnInitialize();
-	}
+	private IExportFilter filter;
+	private MultiStatus status;
```
mx990: I would suggest to encapsulate this into a *build context* record and pass it as a parameter, instead of using member variables. That makes it clear that this information is only needed during build and also allows the resources to be reclaimed by the garbage collector.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -43,218 +47,197 @@
...
 			return;
 		}
 
-		this.outputDirectory = getProjectPreferenceNode().get(PreferenceConstants.OUTPUT_FOLDER,
-				PreferenceConstants.DEFAULT_OUTPUT_FOLDER_NAME);
-		final String exportFilterID = getProjectPreferenceNode().get(PreferenceConstants.EXPORT_FILTER_ID, ""); //$NON-NLS-1$
-		this.exportFilterConfiguration = ExportFilterUtil.getExportFilter(exportFilterID);
-		updateExporter();
-	}
+		performSetup();
```
mx990: Please only get the output folder here, since the export filter and status is not needed during a clean.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -43,218 +47,197 @@
...
-		this.enableAutoExport = getProjectPreferenceNode().getBoolean(PreferenceConstants.ENABLE_TYPE_EXPORT, false);
+	private void exportElement(final SubMonitor monitor, final IFile file) throws CoreException {
+		try {
+			if (!hasRelevantErrorMarker(file)) {
+				monitor.subTask(MessageFormat.format(Messages.FordiacExporter_ExportingType, file.getName()));
+				filter.export(file, getProject().getLocation().append(new Path(outputDirectory)).toString(), true);
+				monitor.split(1);
+			}
+		} catch (final ExportException e) {
+			status.add(new Status(IStatus.ERROR, getClass(), e.getMessage()));
```
mx990: Please add a message indicating the file that caused the problem and pass the exception as the cause instead. That will make it much easier to identify the problematic file. I would also suggest to catch all `Exception`s.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -43,218 +47,197 @@
...
+			monitor.subTask(MessageFormat.format(Messages.FordiacExporter_ExportingType, marker.getName()));
 			try {
-				folder.create(true, true, new NullProgressMonitor());
-			} catch (final CoreException e) {
-				FordiacLogHelper.logError(MessageFormat.format(Messages.ExportBuilder_CannotCreateDirectory,
-						PreferenceConstants.DEFAULT_OUTPUT_FOLDER_NAME));
+				filter.export(null, location.toString(), true, marker);
+				monitor.worked(1);
+			} catch (final ExportException e) {
+				status.add(new Status(IStatus.ERROR, getClass(), e.getMessage()));
```
mx990: Please add a message indicating that the `CMakeLists.txt` export caused the problem and pass the exception as the cause instead. That will make it much easier to identify the problem. I would also suggest to catch all `Exception`s.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -43,218 +47,197 @@
...
-		protected void processError(final String errorMessage) {
-			FordiacLogHelper.logError(errorMessage);
-		}
-
-		@Override
-		protected boolean exportIsCanceled(final IProgressMonitor monitor) {
-			return super.exportIsCanceled(monitor) || isInterrupted();
+	private static boolean hasRelevantErrorMarker(final IFile file) throws CoreException {
+		for (final IMarker marker : file.findMarkers(IMarker.PROBLEM, true, IResource.DEPTH_INFINITE)) {
+			if (FordiacErrorMarker.isModelMarkerType(marker.getType())) {
```
mx990: This will both include non-error markers and exclude problems in ST code. Please just look for `FordiacErrorMarker.PROBLEM_MARKER` and then filter for `IMarker.SEVERITY_ERROR`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
azoitl: done

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
azoitl: done

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
azoitl: done

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
azoitl: done

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
azoitl: done

mx990 (approved):

PR #1791, opened 2025-11-05
signalprocessing-1.0.0  Adds an empty "TypeHash" attribute

franz-hoepfinger-4diac:
Adds an empty "TypeHash" attribute to several function definitions
within the signalprocessing type library.

This change ensures consistency and prepares for future type
hashing implementation.


azoitl:
We are not maintaining 1.0 versions anymore. Also this attribtute should not exist in the 1.0.0lib.

PR #1792, opened 2025-11-05
signalprocessing-3.0.0 Empty Type Hash.

franz-hoepfinger-4diac:
signalprocessing-3.0.0 Empty Type Hash.


azoitl:
There is already a pullrequest handling that: #[1789](https://github.com/eclipse-4diac/4diac-ide/pull/1789). 

Please next time check first to save us both time.

PR #1794, opened 2025-11-05
utils-1.0.0 - add empty Type hash

franz-hoepfinger-4diac:
utils-1.0.0


azoitl:
we are not maintaining 1.0. lib anymore.

PR #1797, opened 2025-11-05
Avoid follow-up errors due to missing types

mx990:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -337,7 +337,7 @@ public static boolean validateVarInOutConnectionsFormsNoLoop(final Connection co
 	}
 
 	private static boolean isDuplicateConnection(final Connection connection) {
-		return connection.getSource().getOutputConnections().stream().anyMatch(
+		return connection.getSource() != null && connection.getSource().getOutputConnections().stream().anyMatch(
```
azoitl, line 343: how can a connection source be null, it should be at least to an errormarker pin

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -400,4 +403,14 @@ private static boolean isPlainBoolVariable(final IInterfaceElement element) {
 		return element instanceof final VarDeclaration varDeclaration && !varDeclaration.isInOutVar()
 				&& !varDeclaration.isArray() && ElementaryTypes.BOOL.equals(varDeclaration.getType());
 	}
+
+	private static boolean isIncomplete(final Connection connection) {
+		return connection.getSource() == null || connection.getSource() instanceof ErrorMarkerInterface
```
azoitl, line 408: same here

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
```diff
@@ -400,4 +403,14 @@ private static boolean isPlainBoolVariable(final IInterfaceElement element) {
...
+				|| connection.getDestination() == null || connection.getDestination() instanceof ErrorMarkerInterface;
```
azoitl, line 409: also for destination

azoitl (approved):

azoitl (changes requested):
accidently approved

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ConnectionAnnotations.java
mx990: I am not sure, but it did occur during testing, so I added the checks. I suspect there may be a problem in the `ConnectionBuilder`.

azoitl: Yes the `ConnectionBuilder` would need more love. Then I guess it is good to have that here. I hope this also helps us to better "see" when things go wrong so that we can later fix the `ConnectionBuilder`.

azoitl (approved):

PR #1799, opened 2025-11-06
Add MRU tracking infrastructure for function blocks

Amadeus1791:
This PR introduces the `MostRecentlyUsedTracker` to track recently used FB types per project. It addresses inefficiencies in the double-click search dialog for element creation where users must repeatedly search large lists. 

Key features that the class provides:
- Per-project MRU tracking using Eclipse ProjectScope preferences
- Maintains ordered list of up to 5 most recent FB types
- Thread-safe implementation with synchronization
- Automatic persistence to preference store

A debug output is included (marked with TODO) to assist with verification and review. 
These statements will be removed once UI integration makes MRU behavior visible to users.

To keep this PR small, there are no UI changes here. This is only the infrastructure that will be integrated in subsequent PRs.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/utilities/MostRecentlyUsedTracker.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/utilities/MostRecentlyUsedTracker.java
@@ -1,0 +1,233 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Wolfgang Schedl.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Wolfgang Schedl - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.fordiac.ide.gef.utilities;
+
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ProjectScope;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/utilities/MostRecentlyUsedTracker.java
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	/**
+	 * Record that a FB type was used (created/inserted). Moves the type to the
+	 * front of the MRU list.
+	 *
+	 * @param fbTypeName the full type name (e.g., "E_CYCLE", "E_SWITCH")
+	 */
+	public synchronized void recordUsage(final String fbTypeName) {
```
azoitl: I think it would be nicer to provide here the TypeEntry. This gives more flexibility what the class is tracking and how it is doing that. Inside you can for now use `getFullTypeName` to get the right name.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/utilities/MostRecentlyUsedTracker.java
```diff
@@ -0,0 +1,233 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Wolfgang Schedl - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.fordiac.ide.gef.utilities;
```
azoitl: I would prefer if this calls is moved to `org.eclipse.fordiac.ide.model.typelibrary`. And that an instance of a `TypeLibraryObject`will maintain an instance of this class. This would give you easy access to the project and also the lifetime of the MRU tracker can be better controlled.

azoitl (changes requested):

azoitl (approved):

PR #1805, opened 2025-11-07
Fix for: Generating Test Infrastructure within IEC 61499

bwiesmayr:
The plug-in generates a test infrastructure directly within IEC 61499 for a test specification. Due to changes with the standard libraries, this feature did not work anymore at all - the adapter / FB types were not found, causing NPEs. This fix updates the names to ensure that they can be identified correctly.


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/testappgen/TestGenBlockNames.java
```diff
@@ -1,10 +1,25 @@
...
+ *     - initial API and implementation and/or initial documentation
+ *   Bianca Wiesmayr - updated identifiers
+ *******************************************************************************/
+
 package org.eclipse.fordiac.ide.fb.interpreter.testappgen;
 
 public class TestGenBlockNames {
 
-	public static final String TIMEOUT_ADAPTER_NAME = "ATimeOutS"; //$NON-NLS-1$
+	public static final String TIMEOUT_ADAPTER_NAME = "iec61499::events::atimeout"; //$NON-NLS-1$
```
azoitl: Although IEC 61499 is case insensitive it may be nicer to use the same case style as in the type: `ATimeOut`.

azoitl (approved):

Issue #1808, opened 2025-11-09
Testing IDE in Linux


franz-hoepfinger-4diac:
<img width="699" height="794" alt="Image" src="https://github.com/user-attachments/assets/38d83314-2f86-4c4e-a867-3230b01061c2" />

[.log](https://github.com/user-attachments/files/23442024/default.log)

what i did: 

Open the IDE (imported a Project) Closed. 

Open the IDE - closed. 
(happens every time i close) 

Ubuntu
Fresh install. 

```
franz@franz-Nitro-AN517-55:~$ lsb_release -a 
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 25.10
Release:	25.10
Codename:	questing
franz@franz-Nitro-AN517-55:~$
```

franz-hoepfinger-4diac:
<img width="831" height="951" alt="Image" src="https://github.com/user-attachments/assets/c48e5a25-ef98-4101-bb4d-fb4c4d5066a8" />

sure no RAM Problem. 
maybe a CPU Power Problem, but i am far away from 100% ...

franz-hoepfinger-4diac:
not happening when idle. so maybe some Task Mutex Problem, which only occurs on Medium and more CPU Load. ? 

[.log](https://github.com/user-attachments/files/23442063/default.log)
latest log.

azoitl:
I can only reproduce it if I have a dirty type editor.

franz-ms-muc:
> I can only reproduce it if I have a dirty type editor.

this means what ? 
dirty means "edit not saved" ?

azoitl:
> > I can only reproduce it if I have a dirty type editor.
> 
> this means what ? dirty means "edit not saved" ?

@franz-ms-muc yes a dirty editor means that there are changes in the editor that are not saved. In 4diac IDE this is indicated with a `*` in its tab.

azoitl:
In addition to my fix in PR#1808 I created an upstream PR to the Eclipse platform: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3530

franz-hoepfinger-4diac:
when can we see this integrated into 4d ? for me this is not very clear, as it is many "layers" now ... (Platform, IDE etc..)

PR #1811, opened 2025-11-10
Harden against selection and update notifications during shutdown #1808

azoitl:
On Linux the description editor's web browsers sends selection and other update notifications which can not be correctly handled. This commit adds checks to protect against so that a save and correct shutdown is happening.

Fixes: https://github.com/eclipse-4diac/4diac-ide/issues/1808


azoitl:
I'll try to upstream the MultipageEditPart fix.

azoitl:
Upstream is in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3530

mx990 (approved):

PR #1812, opened 2025-11-11
Fix watchpoint annotation on FB instances

mx990:
There were recent changes regarding how annotations on FBN elements are shown[^1], which caused watchpoint annotations to appear in the wrong place. This fixes the problem by checking for the correct figure.

[^1]: see also https://github.com/eclipse-4diac/4diac-ide/pull/1797/commits/a4a850f4bf165ec8c9ddc5c38222383ad2660a95


azoitl (changes requested):
The watchpoint annotation is now shown at the right place but If I remove the watchpoint stays

mx990:
I had missed also adapting the `removeStyles(...)` method. Thanks for noticing!

azoitl (approved):

PR #1813, opened 2025-11-11
LibraryBuilder not reporting on removed links

azoitl:
The LibraryBuilder didn't stop and report on library links that could not be removed. Furthermore all strings are now exported.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
@@ -146,7 +146,7 @@ public enum LibraryManager {
 	 * accordingly
 	 */
 	private void checkLibChanges(final SubMonitor progress) {
-		progress.beginTask("Check for library changes", 10); //$NON-NLS-1$
+		progress.beginTask(Messages.LibraryManager_ChekForLibraryChanges, 10);
 		if (watchService == null) {
 			return;
 		}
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -847,17 +849,14 @@ private void importDependencyNodes(final IProject project, final Map<String, Dep
...
-			try {
-				f.delete(true, null);
-				progress.worked(1);
-			} catch (final CoreException e) {
-				// empty
-			}
-		});
+		for (final IFolder folder : linked.values()) {
+			folder.delete(true, progress);
+			progress.worked(1);
```
mx990:
```suggestion
			folder.delete(true, progress.split(1));
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
azoitl: delete is doing its own subprogress monitor so I'm not thinking it is a good idea to split here.

mx990: That is how a `SubMonitor` is intended to be used[^1]. If we do not use `split(...)` when calling `delete(...)`, it will consume the entire ticks of the current progress monitor for each link via its sub-monitor, instead of one tick for each link.

[^1]: see also the JavaDoc for `SubMonitor`, particularly https://github.com/eclipse-equinox/equinox/blob/bd06ec661a36dcf2c6d8cdafa4e554fb22e6f656/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java#L52-L59

azoitl: I guess I just give up on understanding the progressmonitors. changed accordingly.

mx990 (approved):

Issue #1815, opened 2025-11-11
Editing WITH in type does not automatically update the instances


bwiesmayr:
**Typical Scenario:**

- editing FB type that is instantiated in network
- usually: upon saving, popup appears that lists all instances of this type
- all instances are updated automatically

**Problem:**
When the only change is an added/removed WITH construct, the instances are not updated automatically. A pop-up does not appear and instances need to be manually updated in the editor. 

**Possible fix:**
It is possible that the WithCommand has to be extended to trigger the instance search and subsequent update.

PR #1816, opened 2025-11-11
With changes where not considered interface changes #1815

azoitl:
As such changes to the interface didn't result in in automatic type updates.

Fixes: https://github.com/eclipse-4diac/4diac-ide/issues/1815


bwiesmayr (approved):
tested, works fine and makes editing feel a lot more consistent

PR #1820, opened 2025-11-12
Add referenced projects in type library

mx990:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -609,4 +624,26 @@ public int getFeatureID(final Class<?> expectedClass) {
 			return featureID;
 		}
 	}
+
+	protected class TypeLibraryAdapter extends SingletonAdapterImpl {
```
azoitl: shouldn't we hook this adapter to the referenced project's typelibrary?

mx990: Yes, we definitely should. Thanks for noticing!

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -609,4 +627,26 @@ public int getFeatureID(final Class<?> expectedClass) {
...
```
azoitl: not that it matters but why is this class protected?

mx990: No reason at all, changed accordingly.

azoitl (approved):

PR #1828, opened 2025-11-17
Merge Release into 3.0.x

azoitl:


azoitl:
merges from release are currently not building correctly. Still merging.

PR #1834, opened 2025-11-18
Extend network interpreter to support realistic use cases

bwiesmayr:
This pull request introduces several improvements to the operational semantics interpreter, particularly focusing on supporting missing features (adapters, composite type network editors) 

- Extended the network interpreter to correctly process triggers when output events occur. This makes it possible to trigger an output event of an FB instance and then process all connected FB instances
- Added support for adapters in the GUI (dialog lets user select a specific event pin)
- Added support for adapters in the FB type interpreter including recording service sequences
- Added support for adapters in the network (for now only event connections, because data storage is an open question for adapter-contained variables)
- Corrected the saving of execution traces for networks inside composite FB types, the project was previously not correctly located
- Resolved an issue where econtainer was lost during value interpretation by ensuring correct copying of value containers.


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/Messages.java
```diff
@@ -20,6 +20,10 @@ public class Messages extends NLS {
 	public static String RecordExecutionTraceHandler_Incorrect_Selection;
 	public static String RecordExecutionTraceHandler_Select_FB_input_event;
 	public static String CreateRuntimeTestFunctionBlockHandler_Select_Service_Model;
+	public static String SelectAdapterEventDialog_0;
```
azoitl: could you please give these variables sensible names that can help later in translation

azoitl (changes requested):

azoitl (approved):

PR #1841, opened 2025-11-19
tree viewer for library import

moniwe:
This PR replaces the grid for the library importer by a tree. The tree can be sorted according to library names or versions. When sorted by library name users can only select one version of a library the other one should be deselected automatically. When sorted by version library names are matched at selection that one library cannot be selected in two versions (since System is not the same as Device even if Device is contained within System, this is not detected). The tree recognizes the selections when switching the sorting. It also considers the selection current/old/none when a new project is created.

Please test if this fits the user needs.

image
image

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/wizards/LibrarySelectionPage.java
+++ b/plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/wizards/LibrarySelectionPage.java
@@ -60,8 +60,8 @@ public class LibrarySelectionPage extends WizardPage {
 	private final List<LibraryRecord> libraries = new ArrayList<>();
 	private CheckboxTreeViewer treeViewer;
 	private Button libraryNameSort;
-	private boolean showStandard;
-	private boolean showWorkspace;
+	private final boolean showStandard;
+	private final boolean showWorkspace;
 	private VersionRange range;
 	private SelectionAdapter listener;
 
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/wizards/LibrarySelectionPage.java
```diff
@@ -65,310 +76,379 @@ public LibrarySelectionPage(final String pageName, final boolean alwaysComplete,
...
+	private void fillViewer(final boolean isLibrary) {
+		final Object[] tempSelection = treeViewer.getCheckedElements();
+		createColumns(isLibrary);
+		if (null != listener) {
+			treeViewer.getTree().removeSelectionListener(listener);
 		}
-		lib.setSelected(selected);
+		if (isLibrary) {
+			treeViewer.setInput(libraries.stream().map(LibraryRecord::name).distinct().sorted().toList());
+			treeViewer.setCheckedElements(tempSelection);
```
sebHollersbacher, line 251: When changing between Version/Name while having a group selected (e.g. 3.0.0 is selected and changing to Grouping by Name), a StackOverflowError is thrown here

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/wizards/LibrarySelectionPage.java
```diff
@@ -65,310 +76,379 @@ public LibrarySelectionPage(final String pageName, final boolean alwaysComplete,
...
+			treeViewer.setCheckStateProvider(getLibraryNameCheckStateProvider());
+			treeViewer.getTree().addSelectionListener(getLibraryNameListener());
+		} else {
+			treeViewer.setInput(libraries.stream().map(LibraryRecord::version).distinct()
+					.sorted(Comparator.reverseOrder()).map(Version::toString).toList());
+			treeViewer.setCheckedElements(tempSelection);
```
sebHollersbacher: Same here

sebHollersbacher (changes requested):
- When I select a Group (e.g. 3.0.0) the elements inside it are not selected if the Group has not been expanded
- Checking a Group while Grouping by Name does not do anything, maybe in this case we could select the newest version?

moniwe:
Stack overflow should be fixed now. When clicked on the library name the newest version should be selected. But there are still remaining selection problems.

sebHollersbacher (approved):

azoitl (approved):

PR #1858, opened 2025-11-22
Log failed write requests

azoitl:


plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
```diff
@@ -194,6 +195,8 @@ public void writeFBParameter(final Resource resource, final String value, final
 		try {
 			sendREQ(resource.getName(), request);
 		} catch (final IOException e) {
+			FordiacLogHelper.logError(
+					"Could not send request.\nDestionation: " + resource.getName() + "\nRequest:\n" + request); //$NON-NLS-1$ //$NON-NLS-2$
```
mx990: This will likely open two error dialogs, one for the logged error and one for the deployment exception. Should we log it as a warning instead and also include the exception itself?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java
azoitl: The exception is already logged. However that provides to little information to know more. Therefore I wanted to give us more information. But you are right logging that as error is wrong. Warning or info is much better.

mx990 (approved):

PR #1861, opened 2025-11-23
Member access property sheet

azoitl:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/memberaccess/MemberAccessLabelProvider.java
```diff
@@ -0,0 +1,32 @@
+package org.eclipse.fordiac.ide.application.properties.memberaccess;
```
sebHollersbacher, line 13: The Copyright header is missing here

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ShowMemberAccessPinCommand.java
sebHollersbacher, line 1: Is there a reason why this is not combined with `HideMemberAccessPinCommand`?

sebHollersbacher (changes requested):
When updating the FB, pins are removed if they are Structs themselves
image
Here, after the Update, VAR1 would be unchecked and the pin would be removed

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ShowMemberAccessPinCommand.java
azoitl: Yes HideMemberAccessPinCommand utilizes the checks for connections when pins may be hidden and when not. As this is the more complicated code I didn't want to duplicate that code.

azoitl:
> When updating the FB, pins are removed if they are Structs themselves image Here, after the Update, VAR1 would be unchecked and the pin would be removed

Thx for pointing that out. During my tests I noticed this problem as well and it should be fixed in PR #1868.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/memberaccess/MemberAccessLabelProvider.java
azoitl: thx for seeing that should be fixed.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/HidePinCommand.java
mx990: Should we not also skip those checks when `visible` is true (i.e., showing the pin)?

PR #1862, opened 2025-11-23
View mem access pins

azoitl:


plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/InterfaceListImpl.java
```diff
@@ -358,7 +358,7 @@ public InterfaceList copy() {
 	 */
 	@Override
 	public List<VarDeclaration> getVisibleInputVars() {
```
mx990: It seems this is only used in `AbstractFBNetworkEditPart.getFBValues()`. Maybe it would be better to just move it there?

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/InterfaceListImpl.java
```diff
@@ -367,7 +367,7 @@ public List<VarDeclaration> getVisibleInputVars() {
 	 */
 	@Override
 	public List<VarDeclaration> getVisibleOutputVars() {
```
mx990: It seems this is entirely unused. Maybe we could just remove it?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/InterfaceListImpl.java
azoitl: good point. removed it.

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/InterfaceListImpl.java
azoitl: I was not fully sure but you have a point here. removed it.

mx990 (approved):

PR #1864, opened 2025-11-24
Fix JUnit test failures with Eclipse 2025-12 in Maven builds

mx990:
Pin JUnit version to JUnit 5 and upgrade to Tycho 5.0.1-SNAPSHOT to fix test failures with Eclipse 2025-12 target platform in Maven builds[^1][^2]. Also remove JUnit 4 dependencies and fix compilation problems.

[^1]: https://eclipse.dev/eclipse/markdown/?f=news/4.38/pde.md#junit-5-and-junit-6-conflicts-in-eclipse-438
[^2]: https://github.com/eclipse-tycho/tycho/issues/5611


azoitl (changes requested):
I think we need this on 3.0.x branch.

When updating the tycho version in the root pom you also need to update `.mvn/extension.xml`

azoitl (approved):

mx990:
> When updating the tycho version in the root pom you also need to update `.mvn/extension.xml`

In order to use the snapshot version in `.mvn/extension.xml`, we would need to add the snapshot repository globally[^1]. However, it seems to use the snapshot version and build fine regardless. Do you mind if we leave it as is for now?

[^1]: https://github.com/eclipse-tycho/tycho/wiki#configure-snapshots-repository-globally

azoitl:
> > When updating the tycho version in the root pom you also need to update `.mvn/extension.xml`
> 
> In order to use the snapshot version in `.mvn/extension.xml`, we would need to add the snapshot repository globally[1](#user-content-fn-1-6ac5e44c355a365f294e597c34dea133). However, it seems to use the snapshot version and build fine regardless. Do you mind if we leave it as is for now?
> ## Footnotes
> 
>     1. https://github.com/eclipse-tycho/tycho/wiki#configure-snapshots-repository-globally [↩](#user-content-fnref-1-6ac5e44c355a365f294e597c34dea133)


I mostly asked for it as we had build issues when both where not the same. As now the opposite is the case I'm fine with it.

azoitl (approved):

PR #1868, opened 2025-11-24
Parse pin visible

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/CommonElementExporter.java
mx990: Could we just use `addParam()` above now?

azoitl: I can only do that if addParam gets an additional parameter for the name to use for the xml file. This method is used also for devices, segments, resources parameters, and var config, and for these the `pin.getName()` is the right thing. While for the others I need `getRelativeName()`. Not sure what is cleaner. What do you think?

mx990: How about the following:
```java
addNameAttribute(ie.getBlockFBNetworkElement() != null ? ie.getRelativeName(ie.getBlockFBNetworkElement()) : ie.getName());
```

azoitl: that could be an option. Have to check if it breaks var config in typed subapps

sebHollersbacher: It does not change the behavior of VarConfigs in TypedSubApps

azoitl: Just tested and saw the same. Changed accordingly. I think now its much nicer.

mx990 (approved):

sebHollersbacher (approved):

PR #1869, opened 2025-11-24
Check if connection to member access pin can be done

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/validation/LinkConstraints.java
```diff
@@ -227,6 +232,42 @@ public static boolean hasAlreadyInputConnectionsCheck(final IInterfaceElement sr
...
+			ErrorMessenger.popUpErrorMessage(MessageFormat
+					.format(Messages.LinkConstraints_STATUSMessage_ParentHasInputConnection, target.getName()));
+			return false;
+		}
+		return true;
+	}
+
+	private static boolean isAnyChildConnected(final ContainerVarDeclaration contVarDecl) {
+		for (final VarDeclaration mem : contVarDecl.getCachedMembers()) {
+			if (!mem.getInputConnections().isEmpty()) {
```
mx990, line 251: What about output connections (on outputs)?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/validation/LinkConstraints.java
```diff
@@ -227,6 +232,42 @@ public static boolean hasAlreadyInputConnectionsCheck(final IInterfaceElement sr
...
+			}
+			if (mem instanceof final ContainerVarDeclaration memContVarDecl && isAnyChildConnected(memContVarDecl)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	private static boolean isAnyParentConnected(final IInterfaceElement ie) {
+		if (!ie.getInputConnections().isEmpty()) {
```
mx990: same here

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/validation/LinkConstraints.java
azoitl: output connections are like a fan out, therefore we can connect to all members. It means reading a subset.

mx990: Ah, makes sense. Should we then maybe rename this, for example, to `isAnyChildInputConnected` to make the distinction more clear?

azoitl: Good point. Changed accordingly.

mx990 (approved):

PR #1895, opened 2025-11-27
Member access property sheet

azoitl:


azoitl:
I accidently deleted my old PR. So I recreated this one on top of curent develop.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/HideMemberAccessPinCommand.java
```diff
@@ -0,0 +1,81 @@
...
+
+	@Override
+	public void redo() {
+		super.redo();
+		if (parent != null) {
+			parent.getCachedMembers().remove(index);
+		}
+	}
+
+	private void checkRemovePin() {
```
mx990: Do we really need to go to the trouble of removing stale pins?

azoitl: In my first version I had it without it and I found this a bit sloppy so I added it. If you think it is not worth the effort I can easily remove it.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/ConfigurableMoveFBSection.java
```diff
@@ -80,6 +118,29 @@ private void createDataTypeSelector(final Composite composite) {
...
+					// reset checkmark as command was not executed
+					((TreeViewer) event.getSource()).update(node, null);
+				}
+			}
+
+			private Command createMemberAccessPinChangeCommand(final boolean checked, final MemberAccessTreeNode node) {
+				if (!checked) {
+					return new HideMemberAccessPinCommand(getType(), node.getNamePath());
+				}
+				return new ShowMemberAccessPinCommand(getType(), node.getNamePath());
```
mx990: In that case, could we simplify this to:
```suggestion
				return new HidePinCommand(getType().getInterface().getInterfaceElement(node.getNamePath(), checked), checked);
```
and avoid the separate commands?

azoitl: Only if we remove the code that is purging hidden elements and if it is ok that on hiding we may search for pins with create mode on.

mx990: If we can otherwise avoid the separate commands, I believe it is not worth it to purge them. However, why do we need to demand create pins when hiding? The `HideMemberAccessPinCommand` currently does not demand create pins either?

azoitl: The `HideMemberAccessPinCommand` assumes that the pin to hide is there. For the `ShowMemberAccessPinCommand`, or if we combine both, we cannot and should not assume that. Independent if we purge or not. Because we are lazy loading member access pins when we need them and I would like to keep it like that. This would be important for large structs in structs where only a few members should be shown, or only the first level.

That was the main reason for doing two commands. The reason I asked for removing the purging is that purging is different when hiding or showing. The next reason I did two commands.

azoitl: I just noticed that I can use the visible flag to define if a new member needs to be created or not.

mx990: Yes, as I suggested above.

azoitl: ah sorry I missed that.  Do you expect that we will have different places where we will need to change member access pin visibility? If yes I would prefer to have a `ChangeMemberAccessPinVisibilityCommand`. This would allow to encapsulate future changes and simplifies the visibility change for other users.

mx990: In general, I would like to avoid adding separate commands that essentially perform the same action for a subset of elements. We had that for the `ChangeNameCommand` and that made it easy to use the wrong one by mistake.

How about just adding an additional constructor to `HidePinCommand`?

azoitl: You have a point here. I like the idea of a new constructor. However I also changed the name to `ChangePinVisibilityComand` to better reflect the task of the command.

mx990 (approved):

PR #1898, opened 2025-12-01
Fixed creation of duplicate ConfigurableFB specific Attributes

sebHollersbacher:
If an ErrorMarker with a StructuredType attribute got Updated into a Struct-Mux/Demux, the StructuredType attribute got duplicated


azoitl (commented):
Is this not breaking undo? Or would undo in this case not make sense?

sebHollersbacher:
I don't think undoing an Update from an ErrorMarkerFB to a ConfigurableFB does make much sense.

azoitl (approved):
Thx for the feedback and the assurance.

PR #1899, opened 2025-12-01
Restore composite type network traces

Felix-Schmid:
* restore functionality of network traces with composite types
* fix a bug when the network trace starts in an inner network

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
+++ b/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
@@ -448,18 +448,19 @@ public class DefaultRunFBType implements IRunFBTypeVisitor {
 		final FBNetworkRuntime innerRT = fbTypeRuntime.getNetworkRuntime();
 		final FBNetworkRuntime outerRT = innerRT.getOuterNetworkRuntime();
 		final String eventName = eventOccurrence.getEvent().getName();
-		final EList<Connection> outputs;
+		final Event output;
 		final FBNetworkRuntime networkRT;
 
 		if (InterfacePinUtils.isInput(eventOccurrence.getEvent())) {
-			outputs = ConnectionUtils.getOutputConnections(compType.getInterfaceList().getEvent(eventName));
+			output = compType.getInterfaceList().getEvent(eventName);
 			networkRT = innerRT;
 			copyTransferDataToInnerRuntime(outerRT, innerRT, fbTypeRuntime);
 		} else {
+			output = fbTypeRuntime.getFbElement().getInterface().getEvent(eventName);
 			networkRT = outerRT;
 			copyTransferDataToOuter(outerRT, innerRT, fbTypeRuntime);
 		}
-		return switchNetwork(fbTypeRuntime.getFbElement().getInterface().getEvent(eventName), networkRT);
+		return switchNetwork(output, networkRT);
 	}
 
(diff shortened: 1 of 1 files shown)


bwiesmayr:
Hm his undoes quite a lot of changes I lately did to fix some problems I encountered - including the support for adapters. I doubt that it is all needed - also, could you clarify which problems this fixes? Maybe we even need some test cases again to reach all cases...

Felix-Schmid:
Ok, here is a breakdown of the changes in this PR and why I think they are needed:
* In `runCompositeType`, we cannot always use the interface event of the FB element, because the input events of the FB element do not have any output connections. We need to use the interface event of the composite type to correctly find the inner network connections.
* In `createNewTransaction`, the check if the destination interface element is an input is a problem with composite types, as connections of the inner network can actually lead to output pins (the ones of the composite type).
* For a similar reason, the `isInput` check in `runFBNetwork` needs to happen after the check for composite types, as those can also have outputs that need special treatment.
* I moved the untyped SubApp handling to its own method for readability and added the null check for the outer network in case the execution trace was started in an inner network. The outer network needs to be first created in this case.

I don't think this PR broke any of your recent changes, but please correct me if I am wrong on this. I also agree, adding some basic test for composite types would be helpful in the future.

bwiesmayr:
For instance, the change in Line 455 undoes the connection handling for adapters.

Felix-Schmid:
> For instance, the change in Line 455 undoes the connection handling for adapters.

I am sorry, but I don't see how. The outputs variable assigned on line 455 was unused before my change and the `switchNetwork` method itself makes sure to use the `ConnectionUtils.getOutputConnections` for dealing with adapters.

bwiesmayr:
I see. Maybe I have to look at it from a computer to get a better overview

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -448,18 +448,19 @@ public EList<EventOccurrence> runCompositeFBType(final CompositeFBTypeRuntime fb
 		final FBNetworkRuntime innerRT = fbTypeRuntime.getNetworkRuntime();
 		final FBNetworkRuntime outerRT = innerRT.getOuterNetworkRuntime();
 		final String eventName = eventOccurrence.getEvent().getName();
-		final EList<Connection> outputs;
+		final Event output;
 		final FBNetworkRuntime networkRT;
 
 		if (InterfacePinUtils.isInput(eventOccurrence.getEvent())) {
-			outputs = ConnectionUtils.getOutputConnections(compType.getInterfaceList().getEvent(eventName));
+			output = compType.getInterfaceList().getEvent(eventName);
```
bwiesmayr, line 455: I see now what you meant 👍 this was written with the best intentions but not executed well 😆

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -578,6 +568,27 @@ public EList<EventOccurrence> runFBNetwork(final FBNetworkRuntime fBNetworkRunti
 		return typeOutputEos;
 	}
 
+	private EList<EventOccurrence> runUntypedSubApp(final FBNetworkRuntime fBNetworkRuntime,
```
bwiesmayr, line 571: I think the clean solution (mid-term) will be to create a dedicated UntypedSubappRuntime when it is an untyped subapp

bwiesmayr (approved):

PR #1921, opened 2025-12-05
Gracefully handle empty type hash attributes

mx990:
This defaults to an empty string when the value of the type hash attribute is empty, instead of throwing an exception.


moniwe (approved):
error is gone when deployed with empty hash value

PR #1929, opened 2025-12-07
Improve Default Block Size Paramters

azoitl:
The default values of the block size preferences used to much space. This change tries to find a compromise betwen space and look and feel.


mx990 (approved):

ernstblechaPT (approved):

bwiesmayr (commented):
these are really useful settings

PR #1948, opened 2025-12-13
Introduce dedicated instance and full copy methods for interface list

azoitl:
The previous copy method did only copy a subset required for instance creation. This was easily misused leading to strange behavior.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBNetworkElementEvaluator.java
+++ b/plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBNetworkElementEvaluator.java
@@ -62,7 +62,7 @@ public abstract class DeploymentFBNetworkElementEvaluator<T extends FBType, I ex
 		super(type, context, variables, parent);
 		instance.setName(getName() + "_" + UUID.randomUUID().toString()); //$NON-NLS-1$
 		instance.setTypeEntry(getType().getTypeEntry());
-		instance.setInterface(getType().getInterfaceList().copy());
+		instance.setInterface(getType().getInterfaceList().instanceCopy());
 		this.instance = instance;
 		this.eventCounters = instance.getInterface().getEventOutputs().stream()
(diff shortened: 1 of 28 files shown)


plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBNetworkElementEvaluator.java
```diff
@@ -62,7 +62,7 @@ protected DeploymentFBNetworkElementEvaluator(final T type, final I instance, fi
 		super(type, context, variables, parent);
 		instance.setName(getName() + "_" + UUID.randomUUID().toString()); //$NON-NLS-1$
 		instance.setTypeEntry(getType().getTypeEntry());
-		instance.setInterface(getType().getInterfaceList().copy());
+		instance.setInterface(getType().getInterfaceList().instanceCopy());
```
azoitl, line 65: @mx990 should this maybe be a full copy?

mx990: I do not believe so, since I am only creating an instance of the type here.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.deployment.eval/src/org/eclipse/fordiac/ide/deployment/eval/fb/DeploymentFBNetworkElementEvaluator.java
azoitl: I thought that maybe you want to have type values for debugging here. Even if it is an instance.

mx990: The variables of the evaluator are initialized based on the type in the `FBEvaluator` super class.

azoitl: thx for the explanation. That helps to better find my way in the code.

Issue #1916, opened 2025-12-04
adding a plug/socket to CFB interface adds FB representation with interfaces of two adapters


moniwe:
On develop branch, create a new CFB and add a plug/socket to its interface. When this adapter is created at first it is of type ATimeOut. Change this type to another one. Then you get a representation inside the CFB that contains your adapter interface but also those of the ATimeOut e.g.

<img width="166" height="224" alt="Image" src="https://github.com/user-attachments/assets/e1d085d7-dcbb-458e-99fe-43af96de1b2a" />

PR #1950, opened 2025-12-14
Changing Adpater Types for Basics Didn't Correctly Update the Model

azoitl:
As the model now relies on a AdapterFB instance for all FB types the change adapter type command needs to update the AdapterFB for all FB Types.

This is a precondition for fixing: #1916

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
+++ b/plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
@@ -28,6 +28,7 @@ import org.eclipse.fordiac.ide.model.commands.change.UpdateFBTypeCommand;
 import org.eclipse.fordiac.ide.model.commands.change.UpdateInternalFBCommand;
 import org.eclipse.fordiac.ide.model.data.AnyDerivedType;
 import org.eclipse.fordiac.ide.model.data.StructuredType;
+import org.eclipse.fordiac.ide.model.libraryElement.AdapterDeclaration;
 import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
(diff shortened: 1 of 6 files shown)


mx990:
Is it not overkill to use the `AbstractUpdateBlockFBNElementCommand` when the (adapter) FB is not inside a network? Can we not just use the `UpdateInternalFBCommand` for adapters in basic FBs?

azoitl:
It is not only an overkill it was wrong to `AbstractUpdateBlockFBNElementCommand` for everything. However because of some specialties in the adapter model the `UpdateInternalFBCommand` could also not be used. Furthermore updating adapter types was also broken. I think we are now in a better state.

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
```diff
@@ -138,19 +141,41 @@ private void handleDependencyUpdate(final TypeEntry typeEntry) {
...
+					// do nothing
+				}
+				}
+
 				if ((typeEntry instanceof FBTypeEntry || typeEntry instanceof SubAppTypeEntry)) {
 					handleBlockTypeDependencyUpdate(editedElement, typeEntry);
 				}
 				if (typeEntry instanceof final DataTypeEntry dtEntry) {
 					handleDataTypeEntryUpdate(editedElement, dtEntry);
 				}
```
mx990: I believe this is duplicate now?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/TypeEntryAdapter.java
azoitl: Ah damn, yes. Removed.

mx990 (approved):

PR #1951, opened 2025-12-14
Changing Adapter Type in CFBs Added wrong Error Event Pins #1916

azoitl:
The FB update code didn't correctly check when pins need to be preserved. This lead to error pins when there shouldn't be ones. This was most visible for changing adapter types in CFBs but happened also for all other FB instances.

Fixes https://github.com/eclipse-4diac/4diac-ide/issues/1916

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
@@ -395,12 +395,12 @@ public abstract class AbstractUpdateBlockFBNElementCommand extends Command
 
 	private void processEvents(final InterfaceList interfaceList) {
 		for (final Event input : interfaceList.getEventInputs()) {
-			if (input.getInputConnections().isEmpty() || (!input.getAttributes().isEmpty())) {
+			if (input.getInputConnections().isEmpty() && !input.getAttributes().isEmpty()) {
 				updateSelectedInterface(input, newElement);
 			}
 		}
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
```diff
@@ -395,12 +395,12 @@ protected void handleParameters() {
 
 	private void processEvents(final InterfaceList interfaceList) {
 		for (final Event input : interfaceList.getEventInputs()) {
-			if (input.getInputConnections().isEmpty() || (!input.getAttributes().isEmpty())) {
+			if (!input.getInputConnections().isEmpty() || !input.getAttributes().isEmpty()) {
```
mx990: As I understand it, this is supposed to specifically handle event pins *without* connections but *with* attributes. I believe it should be changed to the following instead[^1]:
```suggestion
			if (input.getInputConnections().isEmpty() && !input.getAttributes().isEmpty()) {
```

[^1]: see also the similar condition in `processVars(..)` below

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
azoitl: Good Point. I missed that we have a dedicated section checking for connections. When we finde some time this class would really need some love for beeing cleaned and duplicated loops to be harmonized. but that is not for a bug fix on 3.0.x.

mx990 (approved):

PR #1952, opened 2025-12-14
Renaming System File was refering to an old rename participant

azoitl:


plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
```diff
@@ -474,7 +474,13 @@
             id="org.eclipse.fordiac.ide.typemanagement.renameType"
             name="Rename Type handler">
          <enablement>
-         	<reference definitionId="org.eclipse.fordiac.ide.typemanagement.typeFile" />
+           <or>
+             <reference definitionId="org.eclipse.fordiac.ide.typemanagement.typeFile" />
+             <and>
+               <instanceof  value="org.eclipse.core.resources.IFile" />
+               <test forcePluginActivation="true" property="org.eclipse.core.resources.extension" value="sys"/>
+             </and>
```
mx990: I would suggest to use the following instead:
```suggestion
            <adapt type="org.eclipse.core.resources.IFile">
              <test forcePluginActivation="true" property="org.eclipse.core.resources.extension" value="sys"/>
            </adapt>
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
azoitl: done

mx990 (approved):

PR #1958, opened 2025-12-16
Refactor exporter

MandKastner:
Abstract exporter is not needed anymore

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/FordiacExportWizard.java
+++ b/plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/FordiacExportWizard.java
@@ -23,14 +23,11 @@ import java.util.List;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.runtime.IConfigurationElement;
-import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.fordiac.ide.export.AbstractExporter;
+import org.eclipse.fordiac.ide.export.Exporter;
 import org.eclipse.fordiac.ide.export.IExportFilter;
 import org.eclipse.fordiac.ide.export.ui.Messages;
 import org.eclipse.fordiac.ide.ui.FordiacLogHelper;
 import org.eclipse.jface.dialogs.IDialogSettings;
-import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.wizard.Wizard;
@@ -69,7 +66,6 @@ public class FordiacExportWizard extends Wizard implements IExportWizard {
 	@Override
 	public void addPages() {
 		super.addPages();
-
 		page = new SelectFBTypesWizardPage(Messages.FordiacExportWizard_WizardPage, selection);
 		page.setDescription(Messages.FordiacExportWizard_DESCRIPTION_WizardPage);
 		page.setTitle(Messages.FordiacExportWizard_TITLE_WizardPage);
@@ -77,6 +73,7 @@ public class FordiacExportWizard extends Wizard implements IExportWizard {
 	}
 
 	@Override
+	@SuppressWarnings("squid:S2142")
 	public boolean performFinish() {
 		page.saveWidgetValues();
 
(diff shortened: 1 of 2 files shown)


azoitl (changes requested):
The problem of this change is that after the runable is started imediatly the erros are checked. But you don't know if the exporter has finished. Therefore this has to be done inside of the runable.

mx990:
From the JavaDoc of `org.eclipse.jface.wizard.WizardDialog.run(boolean, boolean, IRunnableWithProgress)`:
> This implementation of `IRunnableContext.run(boolean, boolean, IRunnableWithProgress)` blocks until the runnable has been run, regardless of the value of `fork`.

It should therefore be safe to assume the exporter has finished after returning from `run(...)`.

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/FordiacExportWizard.java
```diff
@@ -86,12 +83,13 @@ public boolean performFinish() {
...
-		final ExporterRunnable exporter = new ExporterRunnable(page.getSelectedExportFilter(), exportees,
-				page.getDirectory(), page.overwriteWithoutWarning(), page.enableCMakeLists());
+		final Exporter exporter = new Exporter(page.getSelectedExportFilter(), page.getDirectory(),
+				page.overwriteWithoutWarning(), page.enableCMakeLists());
 		try {
 			setNeedsProgressMonitor(true);
-			getContainer().run(true, true, exporter);
-		} catch (final Exception e) {
+			getContainer().run(true, true, monitor -> exporter.exportElements(monitor, exportees));
+			Display.getDefault().asyncExec(() -> showErrorWarningSummary(exporter.getExportFilter()));
```
mx990: Would it not be better to show this synchronously in the current UI thread?

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/Exporter.java
```diff
@@ -16,31 +16,45 @@
...
+	}
+
 	public void exportElements(final IProgressMonitor monitor, final List<IFile> exportees) {
 		monitor.beginTask(MessageFormat.format(Messages.FordiacExporter_ExportingSelectedTypesUsingExporter,
 				filterConfig.getAttribute("name")), exportees.size() + 1); //$NON-NLS-1$
 
 		if (null != filter) {
 			for (final IFile file : exportees) {
-				if (!exportIsCanceled(monitor)) {
+				if (!monitor.isCanceled()) {
```
mx990: I would suggest also throwing an `OperationCanceledException` here, same as below, instead of continuing with the loop.

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/Exporter.java
mx990: same here

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/Exporter.java
```diff
@@ -107,20 +110,13 @@ protected void exportElement(final IProgressMonitor monitor, final IExportFilter
...
-			exportFilter = (IExportFilter) filterConfig.createExecutableExtension("class"); //$NON-NLS-1$
-		} catch (final CoreException e) {
-			processError(Messages.FordiacExporter_ERROR + e.getMessage());
-		}
-		return exportFilter;
+	protected static void processError(final String errorMessage) {
+		FordiacLogHelper.logError(errorMessage);
+		final MessageBox msg = new MessageBox(Display.getDefault().getActiveShell());
+		msg.setMessage(errorMessage);
+		msg.open();
```
mx990: I believe this will result in an illegal thread access, since we are not running the exporter in the UI thread. The `FordiacLogHelper.logError(...)` should also already result in a message dialog, so this should not be necessary.

mx990 (changes requested):

azoitl:
> From the JavaDoc of `org.eclipse.jface.wizard.WizardDialog.run(boolean, boolean, IRunnableWithProgress)`:
> 
> > This implementation of `IRunnableContext.run(boolean, boolean, IRunnableWithProgress)` blocks until the runnable has been run, regardless of the value of `fork`.
> 
> It should therefore be safe to assume the exporter has finished after returning from `run(...)`.

Interesting. I was just reading the documentation of the `fork` attribute. Should have read everything.

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/FordiacExportWizard.java
```diff
@@ -88,21 +86,15 @@ public boolean performFinish() {
 		try {
 			setNeedsProgressMonitor(true);
 			getContainer().run(true, true, monitor -> exporter.exportElements(monitor, exportees));
-			Display.getDefault().asyncExec(() -> showErrorWarningSummary(exporter.getExportFilter()));
-		} catch (final InterruptedException | InvocationTargetException e) {
-			showExceptionErrorDialog(e);
+			Display.getDefault().syncExec(() -> showErrorWarningSummary(exporter.getExportFilter()));
```
mx990: The `performFinish()` already runs in the UI thread, so there is no need for the `syncExec(...)`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/FordiacExportWizard.java
MandKastner: Makes sense, thank you.

mx990 (approved):

azoitl (approved):

PR #1960, opened 2025-12-16
extend interpreter so that it can interpret a subapp type from interface pin

bwiesmayr:
handling of SIFBs
add all types to the resourceset for saving
fix creation of composite fb type runtime
handle the text for event occurrences at the interface of subapp types

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
+++ b/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
@@ -84,6 +84,7 @@ import org.eclipse.fordiac.ide.model.libraryElement.FunctionFBType;
 import org.eclipse.fordiac.ide.model.libraryElement.IInterfaceElement;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementFactory;
 import org.eclipse.fordiac.ide.model.libraryElement.STAlgorithm;
+import org.eclipse.fordiac.ide.model.libraryElement.ServiceInterfaceFBType;
 import org.eclipse.fordiac.ide.model.libraryElement.SimpleECAction;
 import org.eclipse.fordiac.ide.model.libraryElement.SimpleFBType;
 import org.eclipse.fordiac.ide.model.libraryElement.UntypedSubApp;
@@ -342,6 +343,7 @@ public class DefaultRunFBType implements IRunFBTypeVisitor {
 		case final BasicFBTypeRuntime b -> b.setBasicfbtype((BasicFBType) copyFBType);
 		case final SimpleFBTypeRuntime s -> s.setSimpleFBType((SimpleFBType) copyFBType);
 		case final FunctionFBTypeRuntime f -> f.setFunctionFBType((FunctionFBType) copyFBType);
+		case final ServiceInterfaceFBTypeRuntime si -> si.setServiceFBType((ServiceInterfaceFBType) copyFBType);
 		default -> throw new UnsupportedOperationException();
 		}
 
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
Felix-Schmid:
```suggestion
if (compTypeRT.getNetworkRuntime().getOuterNetworkRuntime() == null) {
  compTypeRT.getNetworkRuntime().setOuterNetworkRuntime(fBNetworkRuntime);
  // put the composite runtime into the inner network, so we will find our way
  // back to the outer network
  compTypeRT.getNetworkRuntime().getTypeRuntimes().put(eventOccurrence.getParentFB(), compTypeRT);
}
```
Since the network runtime is now set when creating a `CompositeFBTypeRuntime`, we need to change this check, as the outer runtime and the type runtimes still need to be initialized.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -527,6 +527,11 @@ public EList<EventOccurrence> runFBNetwork(final FBNetworkRuntime fBNetworkRunti
 			return runUntypedSubApp(fBNetworkRuntime, uSubApp);
 		}
 
+		// handle initial triggers at outputs of FBs (e.g., for SIFBs)
+		if (!InterfacePinUtils.isInput(eventOccurrence.getEvent())) {
+			return switchNetwork(eventOccurrence.getEvent(), EcoreUtil.copy(fBNetworkRuntime));
+		}
```
Felix-Schmid: Performing this check before the check for composite runtimes at line 544 stops the execution trace from leaving composites. I moved this check to lines 556-559 for this reason, but I guess I broke something else by doing so? (I checked if triggering outputs still work...)

Felix-Schmid (changes requested):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/DefaultRunFBType.java
```diff
@@ -762,6 +775,10 @@ protected static Value getOutputValue(final VarDeclaration variable, final FBNet
 
 	@Override
 	public EList<EventOccurrence> runServiceInterfaceFBType(final ServiceInterfaceFBTypeRuntime fbTypeRuntime) {
+		if (this.eventOccurrence.getEvent().getName().equals("INIT")) { //$NON-NLS-1$
```
cochicde, line 776: this seems to be a workaround to make sure the chain of events at least start. But this is very specific and non-complete for all cases. 

The actual behaviour of the SIFB is unknown, but what about if the `ServiceInterfaceFBTypeRuntime` contains some type of member that executes the actual functionality? This probably shouldn't be part of the EMF model as  to not pollute it and the serialized information, but in some kind of wrapper like

```

public interface IServiceInterfaceFBSimulator {

    /**
     * Triggers the simulation logic for the given FB.
     *
     * @param fb the function block to simulate
     * @param event the even to trigger
     * @return list of generated event occurrences
     */
    EList<EventOccurrence> trigger(FB fb, Event event);
}

public class ServiceInterfaceFBTypeRuntimeSimulator {

    private final ServiceInterfaceFBTypeRuntime runtime;
    private final IServiceInterfaceFBSimulator simulator;

    public ServiceInterfaceFBTypeRuntimeSimulator(
            ServiceInterfaceFBTypeRuntime runtime,
            IServiceInterfaceFBSimulator simulator) {
        this.runtime = runtime;
        this.simulator = simulator;
    }

    public ServiceInterfaceFBTypeRuntime getRuntime() {
        return runtime;
    }

    public IServiceInterfaceFBSimulator getSimulator() {
        return simulator;
    }
}
```

Then the user can define the exact behaviour of each SIFB, which could default to a dummy version which just passes around events wihtout any issues. 

This would allow us to simulate for example blocking SIFB and other behaviours

bwiesmayr: yes this was some initial work around

I like your proposal with the simulators, the default behavior that I have in mind is to have a look at the service sequence model and see if there is some suggestions there
The simulators could then jump in whenever it's getting more complicated and the service sequences don't help

Felix-Schmid (approved):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/handler/RecordExecutionTraceHandler.java
```diff
@@ -90,39 +94,72 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
 		final ResourceSet reset = new ResourceSetImpl();
 
 		final EObject rootContainer = EcoreUtil.getRootContainer(network);
-		final IProject project = switch (rootContainer) {
-		case final Application app -> app.getAutomationSystem().getTypeLibrary().getProject();
-		case final AutomationSystem sys -> sys.getTypeLibrary().getProject();
-		case final FBType type -> type.getTypeLibrary().getProject();
+		final TypeLibrary typelib = switch (rootContainer) {
+		case final Application app -> app.getAutomationSystem().getTypeLibrary();
```
azoitl: with all the changes we did the rootContainer should never be an application and always be a LibraryElement, which is the base class for AutomationSystem and FBType. I think this would make the code simpler.

azoitl (approved):

PR #1971, opened 2025-12-19
Updated Splash to 3.0.1

azoitl:


mx990:
Should we really update the splash for every micro version?

azoitl:
It is very little work and users on startup see what they are using. I think it would be better.

mx990 (approved):

bwiesmayr:
fine with me, but I wonder if it would be less confusing if they get the .1 as a somewhat smaller font next to the 3.0
like THREE POINT ZERO update 1 style 

because I suppose that 3.0 and 3.0.1 would still work well together

azoitl:
Something like this?
image

bwiesmayr:
yes I thought about having there "Update 1" literally but I see that there is too little space for it. So I like your proposal, it keeps more focus on the 3.0

azoitl:
I was inspired by older Eclipse splash screens when they had only one release per year and some update releases:
image

azoitl:
Uploaded the version with reduced size of .1

bwiesmayr (approved):
I'm fine with having such a splash screen :) if other's don't mind that we are updating
I think it's a little different from the Eclipse IDE, because neon would probably equivalent to 3 and the number is then the first subdigit

But I think that it's generally good to see easily which version is in use

PR #1994, opened 2025-12-28
Added MemberAccess pin to InitialValueHelper

azoitl:


plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
```diff
@@ -58,27 +63,75 @@ public static String getInitialOrDefaultValue(final Object element) {
...
+	private static String getMemberAccessDTInitialValue(final VarDeclaration v) {
+		final var namePath = getNamePath(v);
+		final VarDeclaration root = getRootPin(v);
+		return getMemberAccessDTInitialValue(root, namePath);
+	}
+
+	private static String getMemberAccessDTInitialValue(final VarDeclaration parent, final List<String> namePath) {
+		if (parent != null && parent.getType() instanceof final StructuredType struct) {
+			if (!getValue(parent).isBlank()) {
+				// TODO extract initial value from struct value
```
azoitl: @mx990 do we have an infrastructure for getting the initial value for a member from a struct literal?

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
```diff
@@ -58,27 +63,75 @@ public static String getInitialOrDefaultValue(final Object element) {
...
 			}
+		} else if (v.eContainer() instanceof VarDeclaration) {
+			return getMemberAccessDTInitialValue(v);
 		}
 
 		return ""; //$NON-NLS-1$
 	}
 
+	private static String getMemberAccessDTInitialValue(final VarDeclaration v) {
+		final var namePath = getNamePath(v);
```
mx990: I would suggest to use `v.getBlockRelativePath()` instead.

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
mx990: The following should give you the value of a member, regardless of the type of variable:
```java
	private static String getMemberDefaultValue(final VarDeclaration parent, final List<String> namePath) {
		try {
			Variable<?> variable = VariableOperations.newVariable(parent);
			for (final String name : namePath) {
				variable = variable.getChildren().filter(child -> Objects.equals(child.getName(), name)).findFirst()
						.orElseThrow();
			}
			return variable.toString();
		} catch (final Exception e) {
			return ""; //$NON-NLS-1$
		}
	}
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
azoitl: thx for this info. applied accordingly

plugins/org.eclipse.fordiac.ide.model.edit/src/org/eclipse/fordiac/ide/model/edit/helper/InitialValueHelper.java
azoitl: You are right it is cleaner to reuse this method even if I don't need the first element, which was the reason I did my own version. Changed accordingly.

mx990 (approved):

PR #1999, opened 2026-01-05
Fixed Update-FB-Type quickfix to include SubAppTypes

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -47,7 +47,7 @@ public ChangeFBMarkerResolution(final IMarker marker) {
 	protected boolean prepare(final IMarker[] markers, final IProgressMonitor monitor) throws CoreException {
 		final DataTypeTreeSelectionDialog dialog = new DataTypeTreeSelectionDialog(
 				PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
-				FBTypeSelectionTreeContentProvider.INSTANCE, FBTreeNodeLabelProvider.INSTANCE);
+				FBSelectionTreeContentProvider.INSTANCE, FBTreeNodeLabelProvider.INSTANCE);
```
mx990: This will also allow to select SubApps and Function FBs for internal FBs. I believe we have to restrict the selection as follows:
- allow only FB types for Base FB types,
- allow only FB and FC types for Composite FB types,
- allow all FB, FC, and SubApp types for SubApp types and systems.

I would suggest to use `TypeEntry.getTypeEClass()` to determine the class of the type for each marker and show an error when an incompatible type is selected.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
sebHollersbacher: Thank you for pointing that out.
However, I'm not sure how to get the marker class using `TypeEntry.getTypeEClass`.
I would have gotten the markers TypeEntry over the Resource and TypeLibrary

mx990: Yes, I would have gotten it that way, as well.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/marker/resolution/ChangeFBMarkerResolution.java
```diff
@@ -81,4 +89,13 @@ public String getDescription() {
 	public Image getImage() {
 		return null;
 	}
+
+	private boolean isCompatible(final LibraryElement newElement, final IMarker[] markers) {
+		return Arrays.stream(markers).map(IMarker::getResource).filter(IFile.class::isInstance).map(IFile.class::cast)
+				.map(file -> getTypeLibrary().getTypeEntry(file).getType())
```
mx990: Please add a null check/filter for the result of `getTypeEntry(file)` in case the file has been deleted in the meantime. Please also use `TypeEntry.getTypeEClass()` instead of `getType()` to avoid potentially loading the type unnecessaily. I would also suggest to cache the list of `EClass`es in `prepare` to avoid re-retrieving them on each selection change.

mx990 (changes requested):

mx990 (approved):

PR #2010, opened 2026-01-10
Limit Size of github url error reporting

azoitl:
For very long stack traces github reports an error. To avoid that the reported text is capped at 4k chars.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
+++ b/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/issuereport/GitIssueCreator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Felix Schmid
+ * Copyright (c) 2025, 2026 Johannes Kepler University Linz
  *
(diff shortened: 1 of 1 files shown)


Felix-Schmid (changes requested):
Using `substring` can result in invalid URLs. This can happen, for example, if a space encoded as “%20” is truncated directly after “%”.

ernstblechaPT:
> Using `substring` can result in invalid URLs. This can happen, for example, if a space encoded as “%20” is truncated directly after “%”.

I second that - limiting the length should probably be done in the `buildBody` method when adding the full stacktrace to also ensure that the code-block is closed

azoitl:
> Using `substring` can result in invalid URLs. This can happen, for example, if a space encoded as “%20” is truncated directly after “%”.

Ah I missed that the encoding is not done in the URI/URL line below. Thx for pointing that out.

@ernstblechaPT 
> I second that - limiting the length should probably be done in the buildBody method when adding the full stacktrace to also ensure that the code-block is closed

The `buildBody` is not a good place because I only want to do that in the manual github url case. But this brought me to the right place to fix it: `createGitHubIssueManual`

Felix-Schmid (approved):

PR #2012, opened 2026-01-11
Prepare replay debugging to be used with the intepreter in the GUI

cochicde:
As it's now, the replay debugging mechanism uses forte as a simulator to generate all the needed information.

This PR prepares it to be used with the local interpreter.

The major things done here are separated into commits:
1. Read traces directly in the GUI: before, forte read the traces. Now the tracecompass pluging allows us to have the traces loaded directly in the GUI and work from there
2. Separate the forte specfic part and the replay abstraction: The forte simulator needs connecting to forte and reading the watches. The local interpreter doesn't. This tries to create a common interface
3. Small refactoring to move the core of replay debugging (no simulator involved) into its own package
4. Add an interface to the local interpreter: This is mainly taking the functions already existing in the EventManagerUtils and encapsulating it into a class, with extra functions such as injectEvent and similar to what's already there in the FakeEcet in forte.
5. Add GUI stuff to let the user select the simulator.

Disclaimer: replay with the interpreter doesn't work yet. There are fixes still needed there, but I thought it would be better to let this be reviewed/merged and work on that in the meantime.  Also keeps the changes separated.


bwiesmayr (approved):
this is indeed very non-breaking and mainly extends replay debugging

PR #2014, opened 2026-01-11
Change Type Hash to use IEC 61499 XMl as base for Hash Generation

azoitl:
The used XMI representation had referneces to files as part of the text and therefore led to unstable hashes. In order to circumvent that the type hash is now based on the IEC 61499 XML representation.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024, 2025 Primetals Technologies Austria GmbH
+ * Copyright (c) 2024, 2026 Primetals Technologies Austria GmbH
  *                          Martin Erich Jobst
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
```diff
@@ -100,11 +88,10 @@ public static String hash(final EObject eObject, final String version, final Str
 				DigestOutputStream dos = new DigestOutputStream(nullOut, digest)
 
 		) {
-			final HashMap<String, Object> options = new HashMap<>();
-			options.put(XMLResource.OPTION_PROCESS_DANGLING_HREF, XMLResource.OPTION_PROCESS_DANGLING_HREF_DISCARD);
-			options.put(XMLResource.OPTION_SKIP_ESCAPE_URI, Boolean.FALSE);
-
-			xmiResource.save(dos, options);
+			final var exporter = FordiacTypeResource.getTypeExporter(copyForHashing(eObject));
```
mx990: Why not just use the resource to serialize the type, instead of making `FordiacTypeResource.getTypeExporter(...)` public?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/util/LibraryElementHasher.java
azoitl: Hmh, I was not sure in which way we will evolve our resource. But you have a good point here. Using the FordiacTypeResource is much cleaner. Changed accordingly.

mx990 (approved):

PR #2015, opened 2026-01-11
Cleaned SystemLayoutHandler to not use getTypeEditable

azoitl:


plugins/org.eclipse.fordiac.ide.elk/src/org/eclipse/fordiac/ide/elk/handlers/SystemLayoutHandler.java
```diff
@@ -263,23 +258,29 @@ private static void layoutSubappInTypedSubapp(final SubApp subapp, final SubAppT
...
 
-	private static void layoutType(final IFile file, final Set<Object> visited, final Set<IEditorPart> editors) {
+	private static void layoutType(final IFile file, final Set<IEditorPart> editors) {
 		if (openEditor(file) instanceof final FormEditor multiPageEditor) {
-			final var typeEditable = TypeLibraryManager.INSTANCE.getTypeEntryForFile(file).getTypeEditable();
-			if (typeEditable instanceof final SubAppType subappType) {
-				handleSubappType(multiPageEditor, typeEditable, subappType);
-			} else if (typeEditable instanceof CompositeFBType) {
-				handleCompositeType(multiPageEditor, typeEditable);
+			switch (multiPageEditor.getAdapter(LibraryElement.class)) {
```
mx990, line 267: I would suggest adding a `case null` to prevent an NPE in case `getAdapter` returns `null`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.elk/src/org/eclipse/fordiac/ide/elk/handlers/SystemLayoutHandler.java
azoitl: I thought this is handled by default. But I just learned I was wrong. Fixed!

mx990 (approved):

PR #2016, opened 2026-01-12
Detect broken links in library builder

MandKastner:
The library builder now detects broken links of imported libraries and aborts the build process.


azoitl (changes requested):
When you extract code the new code file needs to have as starting point also the original copyright from the origin file.

azoitl (changes requested):
Could you please perform the requested changes on the original commits and not in seperate ones. If you do it like this I have to squash the commit and I would like to avoid this for the set of clean-ups you performed.

azoitl (approved):

PR #2024, opened 2026-01-15
Update Eclipse Milo to version 1.1.2

m-meingast:


azoitl (commented):
LGTM

@m-meingast it is still a draft. Any reason for that?

ruspl-afed:
@m-meingast this is reasonable workaround for issues in Eclipse Milo packaging, so I would merge this to unblock other items and then we may try work more with Eclipse Milo team to make things better.

azoitl:
@ruspl-afed thx for your feedback. I, however, find it strange that only we should have the problem with this setup. I have the feeling that we are wrongly including the Milo Maven dependencies as OSGI boundles in 4diac IDE. But I must say I never fully understood how to do it. So if you say this is for now fine it is fine for me as well.

ruspl-afed:
@azoitl my primary goal is to resolve a kind of blocking we have here for other releng work in progress. After this one is merged I plan to continue improvements in this area,

mx990 (approved):

azoitl (approved):
@ruspl-afed thx for the feedback. I'm fine with merging this now as it works and we finally are on the 1.x release of Eclipse Milo.

The main reason I was a bit reluctant is that we have in Austria a saying: "Nothing stays longer then a workaround".

ruspl-afed:
> The main reason I was a bit reluctant is that we have in Austria a saying: "Nothing stays longer then a workaround".

@azoitl this is totally fine! 
BTW, we have similar saying in Russian, that may be translated like "there is nothing more permanent than temporary". And needless to say, that this saying works like a charm. 
But.
We need to move forward with limited options in everything, so there is no way to avoid temporary solutions.

azoitl:
yes definitely I want to get forward.

PR #2028, opened 2026-01-19
Fixed setting of VarConfig in FBTypes

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeVarConfigurationCommand.java
```diff
@@ -36,6 +36,10 @@ public ChangeVarConfigurationCommand(final VarDeclaration varDeclaration, final
 
 	@Override
 	public boolean canExecute() {
+		if (varDeclaration.getFBType() != null) {
```
azoitl, line 39: I think it would be nicer to check the root below in line 43 if it is instanceof FBType this makes the code cleaner

azoitl (changes requested):
pressed send to early

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeVarConfigurationCommand.java
sebHollersbacher: Unfortunately, we cannot check the root, as VarConfigs are not allowed in FB's inside CFB's
So I have to check if it's the direct parent of the pin.

azoitl: thx for adding that comment. This helps to better understand.

azoitl (approved):

PR #2036, opened 2026-01-20
Disallow adding recursive members to structs

Felix-Schmid:


mx990:
I do not believe this is a good way to handle the problem. This will let the user happily create recursions until it is time to save. It will then not even provide any information where the recursions are and how to fix them. Aside from that, as a general rule, we should not keep users from saving their work.

I would instead suggest to check for recursions in the `ChangeDataTypeCommand` and prevent creating them in the first place. We can also leverage the `TypeEntry.getDependencies()` to simplify scanning for recursive dependencies.

azoitl:
> I do not believe this is a good way to handle the problem. This will let the user happily create recursions until it is time to save. It will then not even provide any information where the recursions are and how to fix them. Aside from that, as a general rule, we should not keep users from saving their work.
> 
> I would instead suggest to check for recursions in the `ChangeDataTypeCommand` and prevent creating them in the first place. We can also leverage the `TypeEntry.getDependencies()` to simplify scanning for recursive dependencies.

When I saw the title of the PR I thought exactly the same. We need to prevent that the user is loosing data.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/ConcurrentNotifierImpl.java
```diff
@@ -40,7 +41,12 @@ public void eSetDeliver(final boolean deliver) {
 	@Override
 	public void eNotify(final Notification notification) {
 		if (deliver) {
+			if (notifying) {
+				return;
+			}
+			notifying = true;
```
mx990: I am concerned that this may lead to race conditions, since notifications are intentionally dispatched outside of synchronized blocks. Is this really necessary if we prevent adding recursive definitions?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeDataTypeCommand.java
```diff
@@ -108,6 +113,35 @@ public static ChangeDataTypeCommand forTypeDeclaration(final IInterfaceElement i
 		return forTypeName(interfaceElement, typeDeclaration);
 	}
 
+	@Override
+	public boolean canExecute() {
+		if (getInterfaceElement() instanceof VarDeclaration
+				&& getInterfaceElement().eContainer() instanceof final StructuredType parentStruct
+				&& dataType instanceof final StructuredType structType) {
+			final Set<String> structs = new HashSet<>();
+			structs.add(parentStruct.getQualifiedName());
```
mx990: Please use `PackageNameHelper.getFullTypeName(LibraryElement)`, since `getQualifiedName()` does not include the package.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/ConcurrentNotifierImpl.java
Felix-Schmid: One can still create recursive structs by making two structs reference each other (without saving them yet) and then saving them one after the other. This would create a stackoverflow when notifying the adapters. Should i change the "notifying" into an atomic boolean?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangeDataTypeCommand.java
Felix-Schmid: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/ConcurrentNotifierImpl.java
mx990: I believe I have found a simpler way to prevent an infinite recursion in the type entry and also reduce unnecessary notifications in general, see https://github.com/eclipse-4diac/4diac-ide/pull/2051.

Felix-Schmid: Looks good to me :+1:

mx990 (approved):

azoitl (approved):

PR #2042, opened 2026-01-21
Fix Editor and Viewer Background

azoitl:
For the editor background now always the theme color is used this ensures that also in darkmode the editor background is always correct.

For the viewer the editor background and the editor color are swapped so that for all themes a clear differentiation between viewer and editor is possible.

Is this something for a maintenance release as it fixes also some viewer issues on windows?


diplfranzhoepfinger:
> Is this something for a maintenance release as it fixes also some viewer issues on windows?

YES !

azoitl:
> > Is this something for a maintenance release as it fixes also some viewer issues on windows?
> 
> YES !

Done

mx990 (approved):

PR #2048, opened 2026-01-25
Removed any access to systems from the System Manager

azoitl:
As any library element handling is now performed by the type library this functionality is not needed any more and leads only to issues.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/CheckSystem.java
+++ b/plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/CheckSystem.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Primetals Technologies Austria GmbH
+ * Copyright (c) 2021, 2026 Primetals Technologies Austria GmbH
  *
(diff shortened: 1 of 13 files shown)


plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/CheckSystem.java
```diff
@@ -46,8 +46,6 @@ public void execute() throws BuildException {
 					systemPathString));
 		}
 
-		// load the system to get the error markers in place
-		SystemManager.INSTANCE.getSystem(systemFile);
```
azoitl, line 50: As the builder is now parsing and checking all files I don't think a dedicated parsing of the system file is needed. Am I right?

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.ant/src_ant/org/eclipse/fordiac/ide/ant/ant/CheckSystem.java
mx990: Yes, that is unnecessary now.

mx990 (approved):

PR #2056, opened 2026-01-27
Fix SystemExplorerPasteAction project paste to restore "Copy Project"

oberlehner:
Copy paste projects in 4diac workspace failed.
This fix copies all files and sets the new project name to the project descriptor

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.systemmanagement.ui/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide.systemmanagement.ui/plugin.properties
@@ -80,3 +80,10 @@ SystemEditor_SystemInformation=System Information:
 
 WrongBuilderOrderMarkerResolution_Description=Orders Project Builders as recommended
 WrongBuilderOrderMarkerResolution_Label=Reorder Project Builders
+
+
+SystemExplorerPasteAction_CopyProject=Copy Project
+SystemExplorerPasteAction_newProjectName=New project name:
+SystemExplorerPasteAction_nonEmptyProject=Project name must not be empty.
+SystemExplorerPasteAction_projectExists=A project with this name already exists.
+SystemExplorerPasteAction_copy=-copy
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.systemmanagement.ui/plugin.properties
```diff
@@ -80,3 +80,11 @@ SystemEditor_SystemInformation=System Information:
...
 WrongBuilderOrderMarkerResolution_Description=Orders Project Builders as recommended
 WrongBuilderOrderMarkerResolution_Label=Reorder Project Builders
+
+
+SystemExplorerPasteAction_0=Copy Project
+SystemExplorerPasteAction_1=New project name:
+SystemExplorerPasteAction_2=Project name must not be empty.
+SystemExplorerPasteAction_3=A project with this name already exists.
+SystemExplorerPasteAction_4=-copy
+SystemExplorerPasteAction_5=-copy
```
Felix-Schmid: I would suggest using more descriptive names for translation purposes and avoiding the duplicate "-copy" entry.

Felix-Schmid (commented):
Maybe we can find a more general solution in the future to restore this function for type files as well, but I think this fix is fine for now.

azoitl (changes requested):
Haven't full looked at all changes but all changes regarding the externlization of strings and the marking of strings that should be externalized as non externalizable makes me nervous. Please fix that.

oberlehner:
> Maybe we can find a more general solution in the future to restore this function for type files as well, but I think this fix is fine for now.

You're right, this isn't an ideal solution, but currently it's not working at all, so at least users can use the functionality again.

oberlehner:
> Haven't full looked at all changes but all changes regarding the externlization of strings and the marking of strings that should be externalized as non externalizable makes me nervous. Please fix that.



sry the WorkbenchNavigatorMessages just got deleted accidentally by the "externalise strings wizard". i will change that.

Felix-Schmid (approved):

azoitl (approved):

azoitl:
> > Haven't full looked at all changes but all changes regarding the externlization of strings and the marking of strings that should be externalized as non externalizable makes me nervous. Please fix that.
> 
> sry the WorkbenchNavigatorMessages just got deleted accidentally by the "externalise strings wizard". i will change that.

Unfortunately I also had the case where the "externalize string wizard" messed with the rest of the strings

PR #2057, opened 2026-01-27
Show ignored files as infos after export

mx990:


plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/ExportStatusMessageDialog.java
```diff
@@ -89,7 +93,11 @@ public void printMessages() {
 		}
 
 		if (!warnings.isEmpty()) {
-			printMessageList(Messages.ExportStatusMessageDialog_WarningsNotEmpty, warnings, count);
+			count += printMessageList(Messages.ExportStatusMessageDialog_WarningsNotEmpty, warnings, count);
+		}
```
ernstblechaPT, line 95: I think the `count += ` has been incorrect for some time and should in both cases be `count = `; after adding printing infos this now tries to add styling past the end of the text when using printMessageList for the third time

ernstblechaPT (changes requested):

plugins/org.eclipse.fordiac.ide.export.ui/src/org/eclipse/fordiac/ide/export/ui/wizard/ExportStatusMessageDialog.java
mx990: Thanks for noticing! I have now removed the `count` entirely.

ernstblechaPT (approved):

azoitl (approved):

PR #2073, opened 2026-01-29
Refactor interpreter

cochicde:
This PR does to main things:
1. Separation of the execution logic into their own classes: all the methods in the `DefaultFBRunType` class were moved to simpler classes. Eveything was just copied and pasted, nothing new here. Methods only used by one type are private in it and shared methods are protected in a common parent one (FBWithEvaluator or FBWithNetwork) or in some cases in a Utils class (shared among classes without a common parent). The execution itself of the type is in a method call `run` for each class.
2. Recursive creation of runtimes: I'm working in separating the creating logic of the runtimes from their execution logic.  This makes sure that all runtimes of a network already exist whent the parent one is created. Still, I didn't remove the creation out of the execution logic yet, as they're still needed as there are times where runtimes are moved out of their containers.

I know this PR has conflicts with #1960 but any change in the `DefaultFBRunType`  there should be easily transferible as the methods name didn't change. The changes in the factory in this PR are the same as there, but with the extra recursive behaviour


bwiesmayr:
Is there a reason why they are called FunctionFB, BasicFB etc and not Runtime anymore? I find this very easy to confuse with actual FB instances (this is a naming-related question only).
As I already previously mentioned, I fully agree that refactoring is needed in this place.

Isn't the recursive creation of runtimes a performance nightmare? If your idea is to only create each runtime once, how will you know what the state of the network was three event executions ago?

You mentioned that we could have a call, maybe that would help?

cochicde:
> Is there a reason why they are called FunctionFB, BasicFB etc and not Runtime anymore? I find this very easy to confuse with actual FB instances (this is a naming-related question only). As I already previously mentioned, I fully agree that refactoring is needed in this place.

No reason, just a lack of imagination and self-imposed rule to avoid repetion in the name if the package already has information about it. 
The `BasicFB` is actually the "default interpreter of a BasicFBTypeRuntime". I thought first in renaming the `org.eclipse.fordiac.ide.fb.interpreter` package to `org.eclipse.fordiac.ide.fb.interpreter.default` to indicate that these are the default interpreters for all the XXXXTypeRuntime, but that would require more changes and it could become a messy PR. 
`BasicFBRuntime` could work, but isn't it too similar to `BasicFBTypeRuntime`? What about `BasicFBTypeDefaultInterpreter`? or `BasicFBDefaultInterpreter`? or just `BasicFBInterpreter` and we assume all under this pacakge are the default ones? As said, just a lack of imagination mainly. 


> Isn't the recursive creation of runtimes a performance nightmare? If your idea is to only create each runtime once, how will you know what the state of the network was three event executions ago?

Now that I think about it again, this recursive creation might actually be too soon to merge. This is part of my pending branch where I'm cleaning up the extra assignments of runtimes that create some issues maintaining all the runtimes in place. I thought it won't affect anything else, but it might cause issues on the tests checking pre/post runtimes, but I'm not sure.
Regarding performance, for my use case, this is not relevant, as I store the whole state of the network only once at the beginning, and then only the state changes.  

> You mentioned that we could have a call, maybe that would help?
Yes, we could do that.

bwiesmayr:
> > Is there a reason why they are called FunctionFB, BasicFB etc and not Runtime anymore? I find this very easy to confuse with actual FB instances (this is a naming-related question only). As I already previously mentioned, I fully agree that refactoring is needed in this place.
> 
> No reason, just a lack of imagination and self-imposed rule to avoid repetion in the name if the package already has information about it. The `BasicFB` is actually the "default interpreter of a BasicFBTypeRuntime". I thought first in renaming the `org.eclipse.fordiac.ide.fb.interpreter` package to `org.eclipse.fordiac.ide.fb.interpreter.default` to indicate that these are the default interpreters for all the XXXXTypeRuntime, but that would require more changes and it could become a messy PR. `BasicFBRuntime` could work, but isn't it too similar to `BasicFBTypeRuntime`? What about `BasicFBTypeDefaultInterpreter`? or `BasicFBDefaultInterpreter`? or just `BasicFBInterpreter` and we assume all under this pacakge are the default ones? As said, just a lack of imagination mainly.
> 
I get it. I have cursed a bit too many times about the different existences of "Point" leading to fully qualified names in many places when multiple implementations are needed in one place. If we get multiple versions sooner or later, I'd prefer BasicFBTypeDefaultInterpreter but it violates my self-imposed rule of not having too long names :D naming is hard. 

> > Isn't the recursive creation of runtimes a performance nightmare? If your idea is to only create each runtime once, how will you know what the state of the network was three event executions ago?
> 
> Now that I think about it again, this recursive creation might actually be too soon to merge. This is part of my pending branch where I'm cleaning up the extra assignments of runtimes that create some issues maintaining all the runtimes in place. I thought it won't affect anything else, but it might cause issues on the tests checking pre/post runtimes, but I'm not sure. Regarding performance, for my use case, this is not relevant, as I store the whole state of the network only once at the beginning, and then only the state changes.

It for sure breaks the visualizations because they depend on historic information at the moment. I always kept every state so that I can also go a step / event occurrence back technically (after I once read about omniescient debugging). 
> 
> > You mentioned that we could have a call, maybe that would help?
> > Yes, we could do that.

cochicde:
After talking about this, it was decided to leave the recursive creation as a separate utility to avoid performance issues where only parts of a network are necessary to be interpreted.

Also, this should only be merge after #1960, therefore I'm marking this PR as draft in the meantime

cochicde:
I rebase my branch to the latest develop and kept the DefaultRunFBType as it was on my branch and moved the new changes to their respectives classe in a separate commit at the end for clarity. 
I had to move `createOutputEventOccurrence` to Utils as is now shared betwenn classes withouta  common parent (serviceInterface and  the ones with a an interpreter).

bwiesmayr (approved):
I have only reviewed the code, but would suggest that we try to integrate your changes and then can test the interpreter in detail again

azoitl (approved):

PR #2089, opened 2026-02-01
Add event selection function blocks for multiple channels

franz-hoepfinger-4diac:


azoitl (changes requested):
The same as for your E_SWITCH Pr #2090 applies for this one.

azoitl:
After a longer thinking I don't think these blocks should be in the lib.

PR #2090, opened 2026-02-01
Adds event switching function blocks (E_SWITCH_1 to E_SWITCH_4)

franz-hoepfinger-4diac:


azoitl (changes requested):
For me these are very strange FBs. When I read E_SWITCH_3 or E_SWITCH_4 I would have epected that I have an USINT input and 3 or 4 Event outputs to which I switch. Waht you are showing here can be done with n E_SWITCH blocks much cleaner and without filling our lib.

franz-hoepfinger-4diac:
> For me these are very strange FBs. When I read E_SWITCH_3 or E_SWITCH_4 I would have epected that I have an USINT input and 3 or 4 Event outputs to which I switch.

this would be redundant, as we already have them as E_DEMUX_2 and E_DEMUX_4

Requesting a Review from @oellith

oellith:
I currently use multiple E_SWITCH blocks to handle this, so for me it would be ok to keep it like this. Or we need a new name when we want a generic function block for this?
Screenshot 2026-03-30 113109

azoitl:
@franz-hoepfinger-4diac pointed me to the Matrix channel where you started the discussion. As Matrix is not the best tool for keeping track of ideas we have put in our [updated contribution guides](https://eclipse.dev/4diac/doc/development/contribute.html) the suggestion to discuss such things in a Github discussion. This allows to be better track design decisions and refer them from PRs.

I forgot that we also have the `E_DEMUX` blocks.  I think we really need to come up with cleaner names and discuss them better. I know nameing is hard.

azoitl:
As nobody came up with a better name I have the impression that these blocks are not as important. Therefore I'm closing this pr.

franz-hoepfinger-4diac:
i was not aware we search for a better name. 
@oellith what do you suggest ? 
it was your original idea ? 

E_MULTISWITCH ?

azoitl:
After thinking more about this I strongly believe that a name change will not help and these blocks should not be added to Eclipse 4diac's library.

The reason for that is that these blocks feature x independent functions with x independent event inputs, x independent boolean inputs and 2*x independent event outputs. Internally their functionality is totally separated. This is a classical example of the bad code smell feature envy. I don't think we should have examples for code smells in our library.

franz-hoepfinger-4diac:
i will not propose this further. 
all my customers use a Adapter-Version of E_SWITCH, as this one is to difficult to understand. 

is there a way to agree on Common Adapters ? 
i know, it is difficult.

azoitl:
> is there a way to agree on Common Adapters ? i know, it is difficult.

@franz-hoepfinger-4diac I don't think everything should immediately go into Eclipse 4diac's "standard" library. I think we need other ways to develop and maintain libs. One option could be some kind of incubation project. Or also a project as part of the examples. With the now rather stable library concept (white paper coming up) there are also options to host libraries outside of Eclipse 4diac. Something that should not be underestimated.

f-hoepfinger-hr-agrartechnik:
i am definitely planning to do so. 
https://components.espressif.com/components/espressif/elf_loader
should be enable that even Customers make their own libs, and then load the Bin of the lib onto the PLC without need to recompile everything. 

keep you posted.

PR #2091, opened 2026-02-01
add one missing FB FB_CTUD_UDINT to fullfill Standard 61131-3

franz-hoepfinger-4diac:
update all other in this Folder.


azoitl (approved):

franz-hoepfinger-4diac:
here we also need the FORTE FB.

PR #2092, opened 2026-02-01
Set priority and scheduling rule for OCL Job

azoitl:


plugins/org.eclipse.fordiac.ide.validation/src/org/eclipse/fordiac/ide/validation/handlers/ValidationHelper.java
```diff
@@ -224,9 +227,19 @@ private static String getApplicationHierarchy(EObject runner) {
 
 	public static void validate(final INamedElement namedElement) {
 		final Job job = new OCLJob("OCL validation", namedElement);
+		job.setUser(true);
+		job.setPriority(Job.BUILD);
+		job.setRule(getResource(namedElement));
 		job.schedule();
 	}
 
+	private static ISchedulingRule getResource(final INamedElement namedElement) {
```
mx990: It seems there is already a `getFile(INamedElement)` above. I would suggest combining both if possible.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.validation/src/org/eclipse/fordiac/ide/validation/handlers/ValidationHelper.java
azoitl: Stupid me I didn't look left or right. Used this one now and adjusted a bit the location of elements to make clearer what to use at which place.

mx990 (approved):

PR #2094, opened 2026-02-01
Add interpreter deployment

cochicde:
This is  a big change, but many things are just being moved around. I tried separtaing into commits to give an idea of what's been done. The main points are:
~- Create an interpreter deployment so the user can deploy to an "interpreter" device.~
~- The interpreter device reuse many of the things from the replay debugging plugin.~
- A full watcher of the interpreter was implemented, helped by a class which bridges the used inteface elements by the interpreter to the values requested when watching an application.

- Create an interpreter deployment which interacts with an already existing ReplaDebuggingDevice. This means that the reaply debugging must be launched first.
- The replay debugging feature was extended to be able to inject events after it was created. This allows:
  - launching a replay debugging session without any previous traces, in which case the network starts at the initial state and events can manually be triggered. Basically, replaying a device is possible with or without previously stored traces.
  -  for sessions with traces  the same applies: new events can be triggered (currently only at the end of the execution timeline)
  -  The timeline widget adjust itself when new events are triggered. 
  -  For all cases where new events are triggered, the same logic occurs where the interpreter is executed and the new state is stored for later reproduction.  

The idea is as follow:
~1. The device is set with the Interpreter profile~
~2. Deploying creates an InterpreterDevice and stores in a singleton. When resources are created, resource executors are created which handle the interpreter interface and the helper watches classes.~
~3. As there's no underlying resource, SIFB are not triggered (START, CYCLE, and so on). However, manual triggering of events allow execution of the app until no events is left to executed. Watches are updated accordingly.~
1. Create a replay debugging configuration for a device with or without traces (empty path for no traces).
2. The Replay Debugging device pass the "interpreter" profile to the creation of the DeviceManamgentExecutor and calls connect.
3. When connecting, the DeviceManamgentExecutor looks for existing replay debugging devices and gets the launched one.
4. Events can be triggered via the DeviceManamgentExecutor to the ReplayDebugging device which handles as usual. 

No more singletons and specific interpreter devices. 

This doesn't break anything as it's about adapting the replay debugging to be used differently. However it needs the recursive creation of networks from #2073 and a small fix in the subapplication, which a merge all in my own fork https://github.com/cochicde/4diac-ide/tree/working-deployment


cochicde:
Work in progress to discuss first the approach and possible ides before diving into this

cochicde:
Now it should be ready. I updated the initial comment. Mainly changes in the replay debugging plugin to be reused without traces and allowing triggering events

cochicde:
nothing changed, I just brought the branch to the current development branch state to make sure nothing breaks. This PR works independently from #2073

cochicde:
any updates on this PR and on #2073 ? I have some other big changes on the pipeline and would like to start merging  step by step

bwiesmayr (approved):
looks good to me

azoitl (approved):
@cochicde sorry for the delay. Can you please provide in a separate PR an update of all plugins you touched to 3.2 version.

PR #2096, opened 2026-02-01
Nattable themes

azoitl:


plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+		configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.SELECT,
+				NatTableWidgetFactory.ERROR_CELL);
+
+		cellStyle = new Style();
+		cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR, defaultBgColor);
+		configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.EDIT,
+				NatTableWidgetFactory.ERROR_CELL);
+	}
+
+	private void configureWarningCell(final IConfigRegistry configRegistry) {
```
mx990:
```suggestion
	private void configureWarningCells(final IConfigRegistry configRegistry) {
```

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+				DisplayMode.NORMAL, NatTableWidgetFactory.LEFT_TRUNCATING);
+
+		configRegistry.unregisterConfigAttribute(CellConfigAttributes.CELL_STYLE, DisplayMode.SELECT,
+				SelectionStyleLabels.SELECTION_ANCHOR_STYLE);
+	}
+
+	private void configureErrorCells(final IConfigRegistry configRegistry) {
+		Style cellStyle = new Style();
+
+		cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR, GUIHelper.COLOR_RED);
```
mx990, line 97: Should we use the configurable error color here?
https://github.com/eclipse-4diac/4diac-ide/blob/32f3f5884341258c75a23f1508aade68fe2bbf15/plugins/org.eclipse.fordiac.ide.gef/plugin.xml#L38-L43

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+
+		cellStyle = new Style();
+		cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR, defaultBgColor);
+		configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.EDIT,
+				NatTableWidgetFactory.ERROR_CELL);
+	}
+
+	private void configureWarningCell(final IConfigRegistry configRegistry) {
+		Style cellStyle = new Style();
+		cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR, GUIHelper.COLOR_YELLOW);
```
mx990, line 114: Should we use the configurable warning color here?
https://github.com/eclipse-4diac/4diac-ide/blob/32f3f5884341258c75a23f1508aade68fe2bbf15/plugins/org.eclipse.fordiac.ide.gef/plugin.xml#L44-L49

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.themes.ITheme;
+
+final class FordiacNatTableConfiguration extends ModernNatTableThemeConfiguration {
+
+	public FordiacNatTableConfiguration() {
+		defaultBgColor = Display.getDefault().getSystemColor(SWT.COLOR_LIST_BACKGROUND);
+		defaultFgColor = Display.getDefault().getSystemColor(SWT.COLOR_LIST_FOREGROUND);
```
mx990: I believe setting both to `null` to use the inherited colors of the container would look cleaner, at least on dark mode:

Image

Image

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+		configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.EDIT,
+				NatTableWidgetFactory.WARNING_CELL);
+	}
+
+	private static void configureDisabledCells(final IConfigRegistry configRegistry) {
+		final Style cellStyle = new Style();
+		cellStyle.setAttributeValue(CellStyleAttributes.BACKGROUND_COLOR,
+				Display.getDefault().getSystemColor(SWT.COLOR_TEXT_DISABLED_BACKGROUND));
+		cellStyle.setAttributeValue(CellStyleAttributes.FOREGROUND_COLOR,
+				Display.getDefault().getSystemColor(SWT.COLOR_TITLE_INACTIVE_FOREGROUND));
```
mx990: I believe using both disabled background and inactive foreground colors makes the text a bit difficult to read on dark mode:

Image

The following leaves the foreground color unchanged:

Image

The following instead leaves the background colors unchanged:

Image

If we use a dark background color, we should definitely also change the color of the tree icon to make it more visible[^1].

[^1]: see `DefaultNatTableThemeConfiguration.treeStructurePainter` and `DefaultNatTableThemeConfiguration.treeStructureSelectionPainter`

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: I tried that, but somehow it didn't look good and we do not have a good solution for the selection case.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: I tried that, but somehow it didn't look good and we do not have a good solution for the selection case.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: hmh I was not aware that the I can set colors to null. This is super. I don't want to have that for the editable cells but for the none editable cells. The new Eclipse theme work with putting the focus on something that is editable. E.g., the text editor has in dark mode a dark background in light mode white. If I now flip that for the non editable cells I think I get what I wanted. Will give it a try.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java
```diff
@@ -400,140 +378,8 @@ private static void addEditDisabledLabel(final DataLayer dataLayer, final IEdita
...
-		table.addConfiguration(selectionStyle);
-		table.addConfiguration(columnHeaderStyle);
-		table.addConfiguration(rowHeaderStyle);
-		table.addConfiguration(new AbstractRegistryConfiguration() {
-			@Override
-			public void configureRegistry(final IConfigRegistry configRegistry) {
-				Style cellStyle = new Style();
-				cellStyle.setAttributeValue(CellStyleAttributes.FOREGROUND_COLOR, GUIHelper.COLOR_DARK_GRAY);
-				configRegistry.registerConfigAttribute(CellConfigAttributes.CELL_STYLE, cellStyle, DisplayMode.NORMAL,
-						DEFAULT_CELL);
```
mx990, line 468: I noticed that this seems to have gotten lost somehow?

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
mx990: That is understandable. I guess we can leave that for a future enhancement.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
mx990: That is understandable. I guess we can leave that for a future enhancement.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java
azoitl: This is now handled by the defaultFGColor

mx990: Without this, I don't notice any difference for default (initial) values, at least on dark mode.

azoitl: oh. I need to bring this back

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: Changing the color of these arrows will be a lot of work. This are images provided by nattable. We would need to either provide own images or write an ICellPainter which would paint the arrows in the foreground color.  I think the later is the only viable solution but I would see this for another PR.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: done

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
azoitl: It was not as easy as I expected but I think the new solution should work now.

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/NatTableWidgetFactory.java
azoitl: done

plugins/org.eclipse.fordiac.ide.ui/src/org/eclipse/fordiac/ide/ui/widget/nattable/FordiacNatTableConfiguration.java
mx990: We could also switch between the provided black and white images depending on the background color.

azoitl: I hoped I could avoid checking the brightness of the background color. But I fear you are right this is most probably the only solution. At least I wrote some years ago an RGB to HSL converted that I can use for that.

mx990 (approved):

sebHollersbacher (approved):

PR #2103, opened 2026-02-03
Fix follow/go-to pin handlers for expanded struct member pins by

oberlehner:
resolving IInterfaceElement#getInterfaceList() recursively via parent pins

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractGoToPinHandler.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractGoToPinHandler.java
@@ -14,7 +14,6 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.handlers;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import org.eclipse.core.commands.AbstractHandler;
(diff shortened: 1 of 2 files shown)


mx990 (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractGoToPinHandler.java
```diff
@@ -89,7 +89,7 @@ private static IInterfaceElement getSelectedInterfaceElement(final ISelection se
 
 	protected static List<IInterfaceElement> getAllInputs(final IInterfaceElement ie) {
 		final List<IInterfaceElement> inputList = new ArrayList<>();
-		final InterfaceList interfaceList = (InterfaceList) ie.eContainer();
+		final InterfaceList interfaceList = ie.getInterfaceList();
 		inputList.addAll(interfaceList.getEventInputs().stream().filter(IInterfaceElement::isVisible).toList());
```
azoitl: This needs to be changed to use `getAllInputs()` otherwise the member access pins will not be listed and navigation will only work on the root pins.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/handlers/AbstractGoToPinHandler.java
```diff
@@ -99,7 +99,7 @@ protected static List<IInterfaceElement> getAllInputs(final IInterfaceElement ie
 
 	protected static List<IInterfaceElement> getAllOutputs(final IInterfaceElement ie) {
 		final List<IInterfaceElement> outputList = new ArrayList<>();
-		final InterfaceList interfaceList = (InterfaceList) ie.eContainer();
+		final InterfaceList interfaceList = ie.getInterfaceList();
 		outputList.addAll(interfaceList.getEventOutputs().stream().filter(IInterfaceElement::isVisible).toList());
```
azoitl: This needs to be changed to use `getAllOutputs()` otherwise the member access pins will not be listed and navigation will only work on the root pins.

azoitl (changes requested):

azoitl (approved):

Issue #2026, opened 2026-01-16
Externalize all User Facing Strings from plugin.xml files


azoitl:
In order to provide 4diac IDE in different languages user facing strings have to be extracted in configuration files. While we have most of these strings removed from the java files we still need to do this for the strings in plugin.xml files. These are mostly menu entries or property sheet tabs.

The tasks to be done are:
 - locate all strings in plugin.xmls
 - generate according plugin.property files
 - configure the plugins to use these strings correctly
 - Bonus: harmonize with the strings extracted from java files to reduce duplicated strings and reduce translation effort

More information on how to prepare Eclipse plugins for translation can be found in this article: https://www.eclipse.org/articles/Article-Internationalization/how2I18n.html

For tracking the plugins in questions:
- [ ] org.eclipse.fordiac.ide 
- [ ] org.eclipse.fordiac.ide.ant
- [x] org.eclipse.fordiac.ide.application
- [ ] org.eclipse.fordiac.ide.attributetypeeditor
- [ ] org.eclipse.fordiac.ide.bulkeditor
- [ ] org.eclipse.fordiac.ide.comgeneration
- [ ] org.eclipse.fordiac.ide.contracts
- [ ] org.eclipse.fordiac.ide.contractspec
- [ ] org.eclipse.fordiac.ide.contractspec.ide
- [ ] org.eclipse.fordiac.ide.contractspec.tests
- [ ] org.eclipse.fordiac.ide.contractspec.ui
- [ ] org.eclipse.fordiac.ide.contractspec.ui.tests
- [ ] org.eclipse.fordiac.ide.datatypeeditor
- [ ] org.eclipse.fordiac.ide.debug
- [ ] org.eclipse.fordiac.ide.debug.replaydebugging
- [ ] org.eclipse.fordiac.ide.debug.replaydebugging.ui
- [ ] org.eclipse.fordiac.ide.debug.st
- [ ] org.eclipse.fordiac.ide.debug.ui
- [ ] org.eclipse.fordiac.ide.debug.ui.st
- [ ] org.eclipse.fordiac.ide.deployment
- [ ] org.eclipse.fordiac.ide.deployment.bootfile
- [ ] org.eclipse.fordiac.ide.deployment.debug
- [ ] org.eclipse.fordiac.ide.deployment.debug.ui
- [ ] org.eclipse.fordiac.ide.deployment.eval
- [ ] org.eclipse.fordiac.ide.deployment.iec61499
[...]

PR #2112, opened 2026-02-04
Externalize User Facing Strings for typemanagement #2026

sshekhar-04:
Part of #2026

Following the "1 plugin per PR" approach discussed with @azoitl , this PR addresses the externalization of user-facing strings for the "org.eclipse.fordiac.ide.typemanagement" plugin.

### Changes:
- **plugin.xml**: Replaced hardcoded labels, names, and descriptions with `%key` references.
- **plugin.properties**: Created a new properties file for this plugin. 

I have ensured that internal identifiers (IDs, class paths, and parameters) remain as hardcoded strings to avoid breaking functional logic. This PR targets the **develop** branch.

Part of https://github.com/eclipse-4diac/4diac-ide/issues/2026

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
@@ -196,3 +196,27 @@ PreviewChange_ReplaceWithMarker=Replace with ErrorMarker
 SafeStructDeletionChange_RootNodeChangeText=Changes for occurrence: {0}{1}
 UpdateUntypedSubappPinChange_0=Pin is not part of untyped subapp
 UpdateFBInstances="Update FB instance"
+Navigator_filters_showonly61499Types_name=Show only IEC 61499 Types
+Navigator_filters_hideAdapters_name=Adapters
+Navigator_filters_hideDevices_name=Devices
+Navigator_filters_hideResources_name=Resources
+Navigator_filters_hideFBs_name=Function Blocks
+Navigator_filters_hideSegments_name=Segments
+
+Navigator_FBTypeLabelProvider_name=FB Type Content
+Navigator_ToolLibraryLabelProvider_name=Toollibrary
+RenameType_name=Rename Type handler
+RenameElement_name=Rename Model Element
+Refactoring_copy_copyTypeParticipant_name=Copy Type Participant
+DeleteFBType_name=Delete FB Type Handler
+Properties_TypeManagementPreferencePage_name=Type Management
+Properties_buildPathPropertyPage_name=4diac Build Path
+NewWizardId=newWizardId
+MoveParticipant1_name=Move To Package Handler
+
+NewTypeWizard_label=New Type
+RenameElement_label=Rename Element
+RepairBrokenConnection_label=Repair broken Connection
+ConnectionsToStruct_label=Connections to Struct
+Refactor_label=Refactor
+TypeDecorator_label=4diac Type Decoration
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
@@ -211,7 +211,6 @@ Refactoring_copy_copyTypeParticipant_name=Copy Type Participant
 DeleteFBType_name=Delete FB Type Handler
 Properties_TypeManagementPreferencePage_name=Type Management
 Properties_buildPathPropertyPage_name=4diac Build Path
-NewWizardId=newWizardId
 MoveParticipant1_name=Move To Package Handler
 
 NewTypeWizard_label=New Type
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
@@ -198,7 +198,7 @@
       <commonFilter
             activeByDefault="true"
             id="org.eclipse.fordiac.ide.typemanagement.navigator.filters.showonly61499Types"
-            name="Show only IEC 61499 Types"
+            name="%Navigator_filters_showonly61499Types_name"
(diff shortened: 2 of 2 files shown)


plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
```diff
@@ -687,11 +687,11 @@
 			locationURI="toolbar:org.eclipse.ui.workbench.file?after=new.group">
         <command
             commandId="org.eclipse.ui.newWizard"
-            label="New Type"
+            label="%Ui.newWizard_label"
             tooltip="Create a new type"
             icon="fordiacimage://ICON_NEW_FUNCTIONBLOCK">
             <parameter
-                name="newWizardId"
```
azoitl, line 694: unfortunately this string shall not be translatable.

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
```diff
@@ -687,11 +687,11 @@
 			locationURI="toolbar:org.eclipse.ui.workbench.file?after=new.group">
         <command
             commandId="org.eclipse.ui.newWizard"
-            label="New Type"
+            label="%Ui.newWizard_label"
```
azoitl: I would prefere to have the new type somewhere captured in the name of the string identifier. Your proposed identifier is to generic and would give translators no hint what the right text would be.

azoitl (changes requested):

sshekhar-04:
Hi @azoitl, thank you for the feedback. I've addressed the issues by:

Reverting the tooltip inside the command parameter to a hardcoded string as it shouldn't be translatable.

Renaming the generic key "Ui.newWizard_label" to a more descriptive "NewTypeWizard_label" to provide better context for translators.

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
```diff
@@ -687,11 +687,11 @@
 			locationURI="toolbar:org.eclipse.ui.workbench.file?after=new.group">
         <command
             commandId="org.eclipse.ui.newWizard"
-            label="New Type"
+            label="%NewTypeWizard_label"
             tooltip="Create a new type"
             icon="fordiacimage://ICON_NEW_FUNCTIONBLOCK">
             <parameter
-                name="newWizardId"
+                name="NewWizardId"
```
azoitl: please keep the caseing as it was

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
```diff
@@ -196,3 +196,29 @@ PreviewChange_ReplaceWithMarker=Replace with ErrorMarker
...
+Navigator_filters_hideFBs_name=Function Blocks
+Navigator_filters_hideSegments_name=Segments
+
+Navigator_FBTypeLabelProvider_name=FB Type Content
+Navigator_ToolLibraryLabelProvider_name=Toollibrary
+RenameType_name=Rename Type handler
+RenameElement_name=Rename Model Element
+Refactoring_copy_copyTypeParticipant_name=Copy Type Participant
+DeleteFBType_name=Delete FB Type Handler
+Preferences_TypeManagementPreferencePage_name=Type Management Preferences
```
azoitl: I just noticed that we have twice the name for a similar thing. I think it should only be one translateable string used for both. Also please remove the Preferences from the printed string.

azoitl (changes requested):

sshekhar-04:
Hi @azoitl, thank you for catching those! I have:

Reverted the casing of newWizardId to match the original.

Harmonized the 'Type Management' strings to use a single key and removed the word 'Preferences' from the translated text as requested.

I've amended and force-pushed the updates. Ready for another look!

plugins/org.eclipse.fordiac.ide.typemanagement/plugin.properties
```diff
@@ -196,3 +196,28 @@ PreviewChange_ReplaceWithMarker=Replace with ErrorMarker
...
+Navigator_FBTypeLabelProvider_name=FB Type Content
+Navigator_ToolLibraryLabelProvider_name=Toollibrary
+RenameType_name=Rename Type handler
+RenameElement_name=Rename Model Element
+Refactoring_copy_copyTypeParticipant_name=Copy Type Participant
+DeleteFBType_name=Delete FB Type Handler
+Properties_TypeManagementPreferencePage_name=Type Management 
+
+Properties_buildPathPropertyPage_name=4diac Build Path
+NewWizardId=newWizardId
```
azoitl: I think this line should also be removed

azoitl (changes requested):
Thanks for your patience. Unfortunately I missed one small thing in my last review. Could you be so kind and fix this.

sshekhar-04:
Good catch, @azoitl! My apologies for leaving that in. I have removed the NewWizardId line from plugin.properties and force-pushed the update. Thank you for your patience during these reviews!

azoitl:
@sshekhar-04 I think something went wrong because I still see the NewWizardId or did I miss anything?

sshekhar-04:
Hi @azoitl, I think there might be some confusion in the line numbers in the diff view.

I have double-checked the current version of the code:

In plugin.xml, I am using the hardcoded name="newWizardId" (lowercase) as it is an internal identifier.

In plugin.properties, there is no entry for NewWizardId or newWizardId.

I've just force-pushed again to ensure the GitHub 'Files Changed' tab is fully synchronized.
the last issue was on line number 694 and the line of code you see is 690.  Could you please check if it looks correct on your end now? Thanks!

azoitl:
To be perfectly sure I just fetched your PR. and also on my local machine it is there. I'll fix it here in gitub.

azoitl (approved):

PR #2119, opened 2026-02-06
update: SetPositionCommandTest now uses IEC61499 coordinate space

Kappuccino111:
@azoitl the test now is successfully updated to only use IEC61499 coordinate space. 

The major change is basically removing the screenToIEC61499() calls. 

Should align with #1985 and #1986.


azoitl (commented):
Thx for the work. Yes with the new API the test got simpler. However should the variable names (`dx`, `dy`) be now changed to something better reflecting the new behavior (`newX`, `newY`) as they are not deltas anymore?

Kappuccino111:
Yes definitely makes sense. Will update and ping 👍

Kappuccino111:
@azoitl this should be good to go. 

On further analysis, I found that simply changing the variable names would be a bit confusing in itself. Since we are now switching to absolute positions themselves, I removed the delta logic and added the absolute logic, making the test more easy to follow.

So basically now instead of `+dx`, `+dy` we have `newX` and `newY` where in `newX = x + dx` and so on. 

Basically we now deal only in absolute positions.

azoitl (approved):
That is an important observation and a very nice clean-up. Thanks.

PR #2127, opened 2026-02-10
feat: Custom plugins may be registered for the error reporting

gravemalte:
This allows us to register custom plugins to the error reporting.

I'm unsure about the name I named the extension point now `errorDialogFilters` but open for suggestions. Also, should I document the whole schema?

I opened the PR to develop now, but I could retarget it to 3.1, or we cherry-pick, depends on what we want.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
+++ b/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
@@ -57,7 +57,14 @@ Usefull when you want to provide a error reporting dialog.
                </documentation>
             </annotation>
          </attribute>
-         <attribute name="issue_url" type="string">
+         <attribute name="issue_url" type="string" use="required">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="preference_qualifier" type="string" use="default" value="org.eclipse.fordiac.ide.issuereport">
             <annotation>
                <documentation>
                   
@@ -83,8 +90,8 @@ Usefull when you want to provide a error reporting dialog.
       <documentation>
          &lt;extension
          point=&quot;org.eclipse.fordiac.ide.errorDialogFilters&quot;&gt;
-  &lt;plugin id=&quot;org.eclipse.ui&quot;/&gt;
-  &lt;plugin id=&quot;org.eclipse.fordiac.ide&quot;/ issue_url=&quot;https://github.com/myuser/myrepo/&quot;&gt;
+  &lt;plugin id=&quot;org.eclipse.ui&quot;/ issue_url=&quot;https://github.com/myuser/myrepo1/&quot;&gt;
+  &lt;plugin id=&quot;org.eclipse.fordiac.ide&quot;/ issue_url=&quot;https://github.com/myuser/myrepo2/&quot;&gt;
 &lt;/extension&gt;
 
 This ensure that the plugin &quot;org.eclipse.ui&quot; and &quot;org.eclipse.fordiac.ide&quot; are registered for filtering.
--- a/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
+++ b/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
@@ -57,6 +57,13 @@ Usefull when you want to provide a error reporting dialog.
                </documentation>
             </annotation>
          </attribute>
+         <attribute name="issue_url" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
       </complexType>
    </element>
 
@@ -76,8 +83,8 @@ Usefull when you want to provide a error reporting dialog.
       <documentation>
          &lt;extension
          point=&quot;org.eclipse.fordiac.ide.errorDialogFilters&quot;&gt;
-		&lt;plugin id=&quot;org.eclipse.ui&quot;/&gt;
-		&lt;plugin id=&quot;org.eclipse.fordiac.ide&quot;/&gt;
+  &lt;plugin id=&quot;org.eclipse.ui&quot;/&gt;
+  &lt;plugin id=&quot;org.eclipse.fordiac.ide&quot;/ issue_url=&quot;https://github.com/myuser/myrepo/&quot;&gt;
 &lt;/extension&gt;
 
 This ensure that the plugin &quot;org.eclipse.ui&quot; and &quot;org.eclipse.fordiac.ide&quot; are registered for filtering.
--- a/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
+++ b/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
@@ -1,0 +1,125 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="org.eclipse.fordiac.ide" xmlns="http://www.w3.org/2001/XMLSchema">
+<annotation>
+      <appInfo>
+         <meta.schema plugin="org.eclipse.fordiac.ide" id="errorDialogFilters" name="Error Dialog Filters"/>
+      </appInfo>
+      <documentation>
+         This schema is used to define custom error dialog filter and to add them to the FordiacLogListener.
+Usefull when you want to provide a error reporting dialog.
+      </documentation>
+   </annotation>
+
+   <element name="extension">
+      <annotation>
+         <appInfo>
+            <meta.element />
+         </appInfo>
+      </annotation>
+      <complexType>
+         <sequence>
+            <element ref="plugin" minOccurs="1" maxOccurs="unbounded"/>
+         </sequence>
+         <attribute name="point" type="string" use="required">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="id" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="name" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+               <appInfo>
+                  <meta.attribute translatable="true"/>
+               </appInfo>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="plugin">
+      <complexType>
+         <attribute name="id" type="string" use="required">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="since"/>
+      </appInfo>
+      <documentation>
+         3.1
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="examples"/>
+      </appInfo>
+      <documentation>
+         &lt;extension
+         point=&quot;org.eclipse.fordiac.ide.errorDialogFilters&quot;&gt;
+		&lt;plugin id=&quot;org.eclipse.ui&quot;/&gt;
+		&lt;plugin id=&quot;org.eclipse.fordiac.ide&quot;/&gt;
+&lt;/extension&gt;
+
+This ensure that the plugin &quot;org.eclipse.ui&quot; and &quot;org.eclipse.fordiac.ide&quot; are registered for filtering.
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="apiinfo"/>
+      </appInfo>
+      <documentation>
+         [Enter API information here.]
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="implementation"/>
+      </appInfo>
+      <documentation>
+         [Enter information about supplied implementation of this extension point.]
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="copyright"/>
+      </appInfo>
+      <documentation>
+         /*******************************************************************************
+ * Copyright (c) 2026 Malte Grave, github.com/gravemalte
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Malte Grave - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+      </documentation>
+   </annotation>
+
+</schema>
--- a/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
+++ b/plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
@@ -57,14 +57,7 @@ Usefull when you want to provide a error reporting dialog.
                </documentation>
             </annotation>
          </attribute>
-         <attribute name="issue_url" type="string" use="required">
-            <annotation>
-               <documentation>
-                  
-               </documentation>
-            </annotation>
-         </attribute>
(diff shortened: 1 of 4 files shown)


mx990:
It looks to me like this creates another extensible infrastructure on top of `ILogListener`. Would it not be easier to just extract the filtering into a method that can then be overridden by a custom implementation? That way, other plugins can register their own log listener with custom filtering.

azoitl:
The advantage of this extension point is that plugins can hook into the exception reporting that we have now in our log listener. The advantage of the extension point is that one does not have to duplicate most code of the log listener to show our reporting dialog / use our reporting infrastructure.

gravemalte:
@mx990 I see the point, but @azoitl explained it quite nicely here. I'm also not the biggest fan of the solution, but other solutions like the OSGI Declarative Services require a lot more work including changing the the plugin to an OSGI plugin, which also includes then autoamtic `MANIFEST.FM`  generation[¹](https://vogella.com/blog/getting-started-with-osgi-declarative-services-2024/).


¹https://vogella.com/blog/getting-started-with-osgi-declarative-services-2024/

mx990:
A subclass of `FordiacLogListener` should still be able to reuse all of the existing implementation for reporting and showing errors, while overriding the filter method.

azoitl:
> A subclass of `FordiacLogListener` should still be able to reuse all of the existing implementation for reporting and showing errors, while overriding the filter method.

This would expose `FordiacLogListener` as API. With @gravemalte's approach we only have an extension point and the internal implementation can change as much as we want. I find this a bit more future prove.

mx990:
The `org.eclipse.fordiac.ide` package is already exposed as API: https://github.com/eclipse-4diac/4diac-ide/blob/9609b450beefda86ac0b27325605c758fc5f9f0b/plugins/org.eclipse.fordiac.ide/META-INF/MANIFEST.MF#L2

If we wanted a more stable API, we could also extract the `LogErrorDialog` and related helper methods into a separate class that can be reused. That would also provide the option to show the error dialog for something that is not logged.

The extension point, in its current form, is also limited to only filter by the plugin id. A filter method that is provided the status could filter for all kinds of things, depending on the use case.

azoitl:
Yes currently it is. But more by accident and not be intention. I still think less API is better then a more stable one. And I think extension points or services are better here.

gravemalte:
I think it depends, I see both points. The biggest pain point I see is how to register a filter for an external plugin that I want to create and hook into the IDE. I need then an Activator which somehow gets a filter registry from 4diac where all filters are in and hook my filter into the registry. Or do I misunderstand you @mx990

mx990:
As I had imagined it, the custom log listener would filter only for its own plugins and is simply hooked into the platform via `Platform.addLogListener(ILogListener)`, just like the `FordiacLogListener` itself. It should thus not need to know about other listeners or filters.

On another note, if we allow to re-use the `FordiacLogListener` in this way, will this not report issues for third-party plugins into the 4diac IDE GitHub repository by default? I believe we then also need a way to specify an alternate URL for issue reporting?

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -123,12 +125,8 @@ protected void createButtonsForButtonBar(final Composite parent) {
 
 	@Override
 	public void logging(final IStatus status, final String plugin) {
-		if ((status.getSeverity() == IStatus.ERROR) && (null != status.getException())
-				&& (status.getPlugin().startsWith(Activator.PLUGIN_ID)
-						|| status.getPlugin().equals(PlatformUI.PLUGIN_ID))
-				// checking/setting the flag must be last, so that we only set it when actually
-				// showing an error dialog and resetting the flag afterwards
```
mx990, line 133: Please do not remove this comment. This is an important implementation detail that was once overlooked already.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -141,6 +139,26 @@ public void logging(final IStatus status, final String plugin) {
 		}
 	}
 
+	private static boolean isPluginInFilterExtension(final String statusPluginId) {
```
mx990: I would suggest caching the filters to avoid re-reading them for each error.

plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
```diff
@@ -0,0 +1,99 @@
...
+   <annotation>
+      <appInfo>
+         <meta.section type="implementation"/>
+      </appInfo>
+      <documentation>
+         [Enter information about supplied implementation of this extension point.]
+      </documentation>
+   </annotation>
+
+
```
mx990, line 131: Please fill out the copyright information for the extension point.

gravemalte:
> As I had imagined it, the custom log listener would filter only for its own plugins and is simply hooked into the platform via `Platform.addLogListener(ILogListener)`, just like the `FordiacLogListener` itself. It should thus not need to know about other listeners or filters.

Now I get it, thanks!


> On another note, if we allow to re-use the `FordiacLogListener` in this way, will this not report issues for third-party plugins into the 4diac IDE GitHub repository? I believe we then also need a way to specify an alternate URL for issue reporting?

Good catch. You can change the URL to specific server AFAIK, but not per custom plugin.

azoitl:
> On another note, if we allow to re-use the `FordiacLogListener` in this way, will this not report issues for third-party plugins into the 4diac IDE GitHub repository by default? I believe we then also need a way to specify an alternate URL for issue reporting?

Hmh. This is a point that is harder to handle. Thx for pointing that out.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: Done.

plugins/org.eclipse.fordiac.ide/schema/errorDialogFilters.exsd
gravemalte: Done.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: I added a caching logic, that was the first thing that came to my mind. If you have a better suggestion please tell me.

mx990: I believe the current implementation will never match any prefix of `pluginID`, since the lookup in `cachedFilters` will return `null` if the `pluginID` is not an exact match.

I would instead suggest using a `TreeSet<String>` and performing the lookup using:
```java
String floor = cachedFilters.floor(pluginID);
return floor != null && pluginID.startsWith(floor);
```

mx990:
If we want to specify an alternate URL for issue reporting, how about making the preference key for the URL part of the extension point and passing the preference key to the `GitIssueCreator`? In that case, I would also see the benefits of using an extension point for this.

I would then suggest using a `TreeMap<String,String>` and performing the lookup using:
```java
Map.Entry<String, String> floor = cachedFilters.floorEntry(pluginID);
if(floor != null && pluginID.startsWith(floor.getKey())) {
  return floor.getValue();   
}
return null;
```

gravemalte:
> If we want to specify an alternate URL for issue reporting, how about making the preference key for the URL part of the extension point and passing the preference key to the `GitIssueCreator`? In that case, I would also see the benefits of using an extension point for this.
> 
> I would then suggest using a `TreeMap<String,String>` and performing the lookup using:
> 
> ```java
> Map.Entry<String, String> floor = cachedFilters.floorEntry(pluginID);
> if(floor != null && pluginID.startsWith(floor.getKey())) {
>   return floor.getValue();   
> }
> return null;
> ```

Very cool idea, like it. I will work it in tomorrow.

azoitl:
> If we want to specify an alternate URL for issue reporting, how about making the preference key for the URL part of the extension point and passing the preference key to the `GitIssueCreator`? In that case, I would also see the benefits of using an extension point for this.

@mx990 I like this idea a lot. Would you also propose to use this for the token based issue creation? 

I would however also allow that implementers of the extension point leave this url open to use the default setup. This would allow people bundling 4diac IDE with own plugins and a different default report configuration in a product.

mx990:
> Would you also propose to use this for the token based issue creation?

Yes, I would propose to provide an alternate qualifier[^1], so that all issue reporting preferences are retrieved against it.

> I would however also allow that implementers of the extension point leave this url open to use the default setup. This would allow people bundling 4diac IDE with own plugins and a different default report configuration in a product.

I would avoid defaulting to the 4diac IDE repository in case it is left empty on accident (or due to lazyness), but it would just be a matter of specifying the current qualifier `org.eclipse.fordiac.ide.issuereport` if this is indeed intentional.

[^1]: see `org.eclipse.fordiac.ide.issuereport.PreferenceConstants.P_BUG_REPORT_PREFERENCE_ID`

gravemalte:
@mx990 Please check again. I added a comment where we change in another PR the behavior of the `GitIssueCreator` I think this should be done separately.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -141,6 +146,37 @@ public void logging(final IStatus status, final String plugin) {
...
+		if (cachedFilters == null) {
+			loadFilterExtensions();
+		}
+		final Map.Entry<String, String> floor = cachedFilters.floorEntry(pluginID);
+		if (floor != null && pluginID.startsWith(floor.getKey())) {
+			// TODO: The issue reporting layer needs to be changed
+			// We have to put the correct issue URL of the filter extension into the
+			// GitIssueCreator, but currently the GitIssueCreator only supports one URL for
+			// all issues.
+			return floor.getValue() != null ? true : false;
```
mx990: The value currently cannot be null.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -141,6 +146,37 @@ public void logging(final IStatus status, final String plugin) {
...
+		final IExtensionRegistry registry = Platform.getExtensionRegistry();
+		final IConfigurationElement[] config = registry
+				.getConfigurationElementsFor("org.eclipse.fordiac.ide.errorDialogFilters"); //$NON-NLS-1$
+		for (final IConfigurationElement e : config) {
+			final String pluginID = e.getAttribute("id"); //$NON-NLS-1$
+			String pluginIssueURL = e.getAttribute("issue_url"); //$NON-NLS-1$
+			// If not plugin reporting URL is not specified, use the default 4diac issue URL
+			if (pluginIssueURL == null) {
+				pluginIssueURL = GitIssueCreator.FORDIAC_IDE_ISSUE_URL;
+			}
```
mx990: As mentioned earlier, I would advise against defaulting to the 4diac IDE repository if the URL is not specified, in order to avoid third-party plugins simply forgetting to specify an alternate URL.

mx990 (changes requested):
I have created a separate PR https://github.com/eclipse-4diac/4diac-ide/pull/2137 that adds the preference qualifier and issue URL parameters.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: Changed to return `true`.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: True, I set it to `""` now. Which is maybe not the best, idea I'm unsure on how the browser would open it.

mx990: I would also set `issue_url` to required in the extension point and skip the configuration element with a warning log message if either `id` or `issue_url` is missing (null or blank).

gravemalte: Okay I see, then ignore my review comment in #2137

gravemalte:
I inserted now the URL of the manual issue creation to the `issue_url`. But I thought it might be better to just insert `https://github.com/myuser/myrepo` instead of the whole body string. The assembly of the whole URL should be done in the `GitIssueCreator`, or what is your opinion? @mx990 @Felix-Schmid.

Also, the question is if it should be done in this PR or if this is another PR.

mx990:
I would suggest rebasing this PR on top of https://github.com/eclipse-4diac/4diac-ide/pull/2137 and providing the `preferenceQualifier` for the filter to the `showErrorDialog` method. The extension point would then specify the preference qualifier instead of the issue URL.

Felix-Schmid:
> The assembly of the whole URL should be done in the `GitIssueCreator`, or what is your opinion?

Yes, that sounds good to me. If the API ever changes, we would only need to change the issue creator class.

gravemalte:
> I would suggest rebasing this PR on top of #2137 and providing the `preferenceQualifier` for the filter to the `showErrorDialog` method. The extension point would then specify the preference qualifier instead of the issue URL.

I adopted the changes. Please test.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -36,9 +44,9 @@
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Link;
 import org.eclipse.swt.widgets.Shell;
-import org.eclipse.ui.PlatformUI;
 
 public class FordiacLogListener implements ILogListener {
+	private static TreeMap<String, String> cachedFilters = null;
```
mx990: I would suggest to initialize the cached filters at declaration to avoid a potential race condition when multiple threads produce errors at the same time:
```java
private static final TreeMap<String, String> cachedFilters = loadFilterExtensions();
```

This also ensures that misconfigured extensions are always identified at startup and not just when the first error is logged.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -137,13 +144,43 @@ public void logging(final IStatus status, final String plugin) {
...
 
+	private static boolean isPluginInFilterMap(final String pluginID) {
+		if (cachedFilters == null) {
+			loadFilterExtensions();
+		}
+		final Map.Entry<String, String> floor = cachedFilters.floorEntry(pluginID);
+		if (floor != null && pluginID.startsWith(floor.getKey())) {
+			return true;
+		}
+		return false;
```
mx990: I believe this can now be simplified to:

```suggestion
		final String floor = cachedFilters.floorKey(pluginID);
		return floor != null && pluginID.startsWith(floor);
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: True, thanks.

plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
gravemalte: Good catch. Didn't think about it.

mx990 (approved):

azoitl (approved):

PR #2135, opened 2026-02-11
Do not consider arrays of struct for configuring F_MOVE blocks

azoitl:


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/AbstractConnectionCreateCommand.java
```diff
@@ -305,7 +305,11 @@ public static boolean isStructManipulatorDefPin(final IInterfaceElement pin) {
 
 	public static boolean shouldStructDataConnCreationBeUsed(final IInterfaceElement pin,
 			final IInterfaceElement other) {
-		return isStructManipulatorDefPin(pin) && other != null && other.getType() instanceof StructuredType;
+		return isStructManipulatorDefPin(pin) && other != null && isSimpleStructPin(other);
+	}
+
+	public static boolean isSimpleStructPin(final IInterfaceElement pin) {
+		return pin.getType() instanceof StructuredType && ((VarDeclaration) pin).getArraySize() == null;
```
mx990:
```suggestion
		return pin.getType() instanceof StructuredType && !((VarDeclaration) pin).isArray();
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/VariableNodeEditPolicy.java
```diff
@@ -71,7 +70,7 @@ protected Command getConnectionCompleteCommand(final CreateConnectionRequest req
 			// if we drag from a struct manipulater but target is not a struct or
 			// configureable F_MOVE pin use normal
 			// data connection creation
-			if (!(pin.getType() instanceof StructuredType)
+			if (!AbstractConnectionCreateCommand.isSimpleStructPin(pin)
```
oberlehner, line 73: i think the AbstractConnectionCreateCommand is a bad place to put this helper function .
shouldtn this be part of the model in  `IInterfaceElement?`

azoitl: I'm not sure if the model is the right place for this. I kept it here because it is the main place which is using it and where currently the checks are.

But I'm open to put it somewhere else.

mx990: How about adding an operation to `ConfigurableFB` to decide if a pin is suitable to connect? An `F_MOVE` could then allow any non-array data pins, while a `STRUCT_(DE)MUX` would allow only structs.

This could then also unify the checks in:
- `AbstractConnectionCreateCommand.shouldStructDataConnCreationBeUsed(IInterfaceElement, IInterfaceElement)`
- `StructDataConnectionCreateCommand.canExecuteConType()`.

azoitl: The check that I changed here are checks that identify if a struct configuring connection creation command or a normal connection creation command should be used. The former shall only be used if the non F_MOVE (I also think mux/demux) pin is a struct. This check is very specific for the command infrastructure and I have a hard time to move it somewhere else, where IMHO it does not belong.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/AbstractConnectionCreateCommand.java
azoitl: Ah missed that one. Fixed.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/AbstractConnectionCreateCommand.java
```diff
@@ -305,7 +305,11 @@ public static boolean isStructManipulatorDefPin(final IInterfaceElement pin) {
 
 	public static boolean shouldStructDataConnCreationBeUsed(final IInterfaceElement pin,
 			final IInterfaceElement other) {
-		return isStructManipulatorDefPin(pin) && other != null && other.getType() instanceof StructuredType;
+		return isStructManipulatorDefPin(pin) && other != null && isSimpleStructPin(other);
+	}
+
+	public static boolean isSimpleStructPin(final IInterfaceElement pin) {
+		return pin.getType() instanceof StructuredType && ((VarDeclaration) pin).isArray();
```
mx990:
```suggestion
		return pin.getType() instanceof StructuredType && !((VarDeclaration) pin).isArray();
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/create/AbstractConnectionCreateCommand.java
azoitl: 🤦 fixed

mx990 (approved):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/policies/VariableNodeEditPolicy.java
oberlehner: ok then i would leave it like this now, and if we need it more often then we can move it later to a different location.

azoitl: Then I'll merge for now. And we will see.

PR #2136, opened 2026-02-12
Insert Adapter Proxy refactoring: generate proxy FB type and rewire adapter connections

oberlehner:
This PR adds the first building block for the Replace Connection with Adapter workflow: automatic adapter-proxy generation + insertion to safely split adapter connections and preserve fan-out/fan-in use sites.

The adapter proxy is an auto-generated Composite FB type named <Adapter>_Proxy that contains exactly one socket (SOCKET1) and one plug (PLUG1) of the selected adapter type. It mirrors the adapter interface in both directions by creating prefixed events/vars (e.g., SOCKET_*, PLUG_*) and wiring them so signals are forwarded and re-exported consistently. When inserted on an adapter connection, it splits one direct connection into two hops via the proxy, preventing fan-in/fan-out issues while preserving existing connection semantics.

**How to test**

In an application editor: right-click an adapter connection → Insert Adapter Proxy → verify a proxy FB instance is inserted and the single connection becomes two adapter connections via the proxy.

In System Explorer: right-click an adapter type (.adp) → Insert Adapter Proxy → verify <Adapter>_Proxy.fbt is generated in the type library.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/plugin.xml
@@ -135,6 +135,12 @@
            id="org.eclipse.fordiac.ide.typemanagement.connectionstostruct"
            name="%command.ConnectionsToStruct">
      </command>
+      <command
+      id="org.eclipse.fordiac.ide.typemanagement.insertAdapterProxy"
+      name="Insert Adapter Proxy"/>
+        <command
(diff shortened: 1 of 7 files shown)


azoitl (approved):

diplfranzhoepfinger:
@oberlehner very interesting. 

do you have some Screenshots or Videos how this should work ? 
i am very interested in this!

PR #2137, opened 2026-02-12
Add preference qualifier parameter and issue URL preference in git issue creator

mx990:
Add preference qualifier and issue URL parameter in git issue creator to specify alternate locations for use in third-party plugins.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
+++ b/plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 Johannes Kepler University Linz
+ * Copyright (c) 2020 Johannes Kepler University Linz, Martin Erich Jobst
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide/src/org/eclipse/fordiac/ide/FordiacLogListener.java
```diff
@@ -79,13 +85,13 @@ public void widgetDefaultSelected(final SelectionEvent e) {
 		private void report() {
 			reportInfo.setText(Messages.FordiacLogListener_ReportInProgress);
 			new Thread(() -> {
-				final Optional<String> url = GitIssueCreator.createIssue(status);
+				final Optional<String> url = GitIssueCreator.createIssue(status, preferenceQualifier, issueURL);
```
gravemalte: ~~I wrote it in already: https://github.com/eclipse-4diac/4diac-ide/pull/2127#discussion_r2799296715~~

~~What if `issueURL` is `""` should we check here if it is `""` and report maybe that if you want to submit an issue to somewhere you have to provide an actual URL?~~

Felix-Schmid (commented):
Perhaps the Git repository for manual reporting should also be specified via the preferences. This would give all 3 reporting modes a consistent approach, and we wouldn't have to deal with the separate issue URL anymore.
I could also add something like this in a future PR if needed.

mx990:
I really like that idea and have changed it accordingly.

Felix-Schmid (approved):

gravemalte (approved):

PR #2144, opened 2026-02-16
Handle referenced projects in library manager

MandKastner:


plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -166,6 +168,17 @@ public static Manifest getManifest(final java.nio.file.Path manifest) {
 		return getManifest(URI.createURI(manifest.toUri().toString()));
 	}
 
+	public static Manifest getReferencedManifest(final IProject project, final String symbolicName) {
+		try {
+			return Stream.of(project.getReferencedProjects()).filter(IProject::isAccessible)
+					.map(ManifestHelper::getContainerManifest).filter(Objects::nonNull)
+					.filter(m -> m.getProduct().getSymbolicName().equals(symbolicName)).findFirst().orElse(null);
```
mx990: I get an NPE here if the `SymbolicName` of the `Product` is not set:
```
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "org.eclipse.fordiac.ide.library.model.library.Product.getSymbolicName()" is null
```

I also noticed that `LibraryManager.collectReferencedDependencies(...)` has a fallback to use the name of the `IProject` in that case, but this does not.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
mx990: I believe you need to return the referenced projects here and also handle deltas of referenced projects above, in order to react to changes in manifests of referenced projects.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -166,6 +166,29 @@ public static Manifest getManifest(final java.nio.file.Path manifest) {
...
 
+	public static Manifest getReferencedManifest(final IProject project, final String symbolicName) {
+		try {
+			final IProject[] projects = project.getReferencedProjects();
+			for (final IProject refProject : projects) {
+				if (!refProject.isAccessible()) {
+					continue;
+				}
+				final Manifest manifest = getContainerManifest(refProject);
+				if (manifest != null && manifest.getProduct() != null) {
```
mx990: I would suggest we also fall back to the project name if it has no manifest or no product.

I would also suggest to extract getting the symbolic name of a project into a separate method to avoid duplicated code with `collectReferencedDependencies(...)` in `LibraryManager`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
mx990: I have added some helper methods and performed a few cleanups. I also found and fixed a problem if no version info was available.

mx990 (approved):

azoitl (approved):

PR #2148, opened 2026-02-17
Externalize User Facing Strings for application plugin #2026

sshekhar-04:
Part of #2026

This PR addresses the externalization of user-facing strings for the `org.eclipse.fordiac.ide.application` plugin. As this is one of the largest and most critical plugins in the IDE, I have performed a thorough review of the `plugin.xml` to ensure a clean separation between translatable labels and internal technical identifiers.

### Key Changes:
- **Internationalization Setup**: Added the `Bundle-Localization: plugin` header to the `MANIFEST.MF`.
- **UI Strings**: Externalized all user-facing `name`, `label`, and `description` attributes for views, menus, and commands.
- **Properties Management**: Created a comprehensive `plugin.properties` using the `application.*` namespace for better organization.
- **Internal Safety**: Verified that all internal identifiers, such as `id`, `commandId`, `parameter name/value`, and `category`, remain hardcoded to prevent any regression in functional logic.

@azoitl , since this is the largest plugin, I have double-checked the diff to ensure no closing tags were accidentally altered. With this submission, I believe the main objectives of #2026 have been met. Please let me know if there are any further plugins you would like me to address.

Part of https://github.com/eclipse-4diac/4diac-ide/issues/2026


azoitl (changes requested):
can you please remove the application prefix. it adds complexity that is not needed, makes it harder for translators and reduces maintainability if these entries are moved to different locations. 

Please also reconsider the names of the strings starting with `jdt...`

sshekhar-04:
@azoitl , I have made the changes as required.

azoitl (approved):
thx for the clean-up

PR #2151, opened 2026-02-17
Fix modulo freeform figure for ECC

Felix-Schmid:
Commit 82c6bd8 appears to have broken the growing background for the System Configuration and ECC diagrams.
This fixes it, although I am not happy with the hard-coded shrink size - maybe someone has a better idea there.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
@@ -73,7 +73,7 @@ public class ModuloFreeformFigure extends AbstractFreeformFigure {
 			final int baseUnit) {
 		final int startExtent = sourceExtent + PADDING + baseOrigin - newOrigin;
 
-		int newExtend = (startExtent / baseUnit) * baseUnit;
+		int newExtend = Math.ceilDiv(startExtent, baseUnit) * baseUnit;
 		if (newExtend < (3 * baseUnit)) {
 			newExtend = 3 * baseUnit;
 		}
@@ -81,11 +81,7 @@ public class ModuloFreeformFigure extends AbstractFreeformFigure {
 	}
 
 	private static int calcAxisOrigin(final int axisPos, final int baseUnit) {
-		if (axisPos < 0) {
-			// when negative we need to go one beyond to have the correct origin
-			return (axisPos / baseUnit) * baseUnit;
-		}
-		return (axisPos / baseUnit) * baseUnit;
+		return Math.floorDiv(axisPos, baseUnit) * baseUnit;
 	}
 
 }


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
```diff
@@ -39,7 +39,7 @@ public ModuloFreeformFigure(final ZoomScalableFreeformRootEditPart zoomScalableF
 	protected Rectangle calculateFreeformExtent() {
 		// adjust size to be a multiple of the base width/height
 		Rectangle contentsExtent = getUnscaledContentsExtent();
-		contentsExtent.shrink(getInsets()); // take any border into our calculation
+		contentsExtent.shrink(1, 1); // do not take shadow into account
```
azoitl: We should definitely also consider the insets here. Because if someone is providing a special border this would not be considered.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
```diff
@@ -83,7 +83,7 @@ private static int calcAxisExtent(final int baseOrigin, final int newOrigin, fin
 	private static int calcAxisOrigin(final int axisPos, final int baseUnit) {
 		if (axisPos < 0) {
 			// when negative we need to go one beyond to have the correct origin
-			return (axisPos / baseUnit) * baseUnit;
+			return (axisPos / baseUnit - 1) * baseUnit;
```
azoitl: This is in prinziple a floor. Should we use floor to make it more explicit?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
```diff
@@ -73,7 +73,7 @@ private static int calcAxisExtent(final int baseOrigin, final int newOrigin, fin
 			final int baseUnit) {
 		final int startExtent = sourceExtent + PADDING + baseOrigin - newOrigin;
 
-		int newExtend = (startExtent / baseUnit) * baseUnit;
+		int newExtend = (startExtent / baseUnit + 1) * baseUnit;
```
azoitl: This is in principle a ceil. should we use that to make the intend more explicit?

azoitl (commented):
I really don't understand why this fixes the problem. 

Should we switch all calculations here to preciseRectangles and doubles like we are doing now more in Draw2d? Maybe this solves the rounding problems at least a bit.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
Felix-Schmid: You are right, I just reverted it to how it was before for a „quick fix“. 

While ceil/floor div is not exactly the same as how it was, it looks like that was more of a bug than a feature, so I changed it to use ceil/floor div now.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java
```diff
@@ -83,7 +83,7 @@ private static int calcAxisExtent(final int baseOrigin, final int newOrigin, fin
 	private static int calcAxisOrigin(final int axisPos, final int baseUnit) {
 		if (axisPos < 0) {
```
azoitl: after seeing the change I think this if should be removed and the floorDiv is here always needed.

azoitl: I took the liberty to fix this myself.

azoitl:
rebased to check merg conflicts.

azoitl (approved):
Now that I see it with ceil and floor div I see where I was wrong with removing the 1 and even more that I should never have used the +1. This is much cleaner now and much better readable.

oberlehner (approved):

PR #2152, opened 2026-02-17
Excluded Further Comments and Non-functional Info From Hash Generation

azoitl:
Version and type information, as well as some comments and subapp sizes where still considered for the hash generation. This is now fixed.


plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -2165,6 +2178,9 @@
       <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
         <details key="ignored" value="true"/>
       </eAnnotations>
+      <eAnnotations source="http:///org/eclipse/fordiac/ide/model/HashMetaData">
+        <details key="ignored" value="true"/>
+      </eAnnotations>
```
mx990, line 2180: It looks like the ignored attribute is now duplicate here?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: thx for noticing. Fixed.

mx990 (approved):

oberlehner (approved):

PR #2158, opened 2026-02-18
Reworked Comment figure to be better aligned with new design

azoitl:
As part of this work also a dark mode version was added and the drawing was simplified improving the performance and reducing garbage during repaints.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/CommentEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/CommentEditPart.java
@@ -13,7 +13,6 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.application.editparts;
 
-import org.eclipse.draw2d.AbstractBorder;
 import org.eclipse.draw2d.Cursors;
 import org.eclipse.draw2d.Figure;
 import org.eclipse.draw2d.Graphics;
@@ -22,8 +21,6 @@ import org.eclipse.draw2d.GridLayout;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.Label;
 import org.eclipse.draw2d.geometry.Dimension;
-import org.eclipse.draw2d.geometry.Insets;
-import org.eclipse.draw2d.geometry.PointList;
 import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.draw2d.text.FlowPage;
 import org.eclipse.draw2d.text.ParagraphTextLayout;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/CommentEditPart.java
```diff
@@ -99,47 +100,53 @@ public void setCommentText(final String newCommentTest) {
...
-			g.fillPolygon(pl);
+
+			updateStickNoteShapePoints(r);
+			g.fillPolygon(STICKY_NOTE_SHAPE_POINTS);
+
+			updateDogEarPoints(r);
+			g.setBackgroundColor(getBorderBolor());
+			g.fillPolygon(STICKY_NOTE_DOG_EAR_POINTS);
+
+			g.setLineWidth(2);
```
mx990: I believe it would look better if all lines had the same width:

Image

compared to the current version:

Image

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/CommentEditPart.java
azoitl: I did that on purpose as it looks a bit better if you stack comments side by side. But I'm not insisting in it.

mx990: I see your point, although I still have a slight preference towards slimmer edges:

image

image

How about using a slightly darker color or a bit of transparency for the edges to make them look a bit softer:

image

azoitl: now that I see your screenshots in dark mode I also think we should reduce the width. edges a bit darker in dark mode is definitely a thing. Let me also check the light version later so that by fixing dark mode we are not breaking light.

azoitl: I played a bit m ore with color and styles and I got this for dark mode:
Screenshot From 2026-02-19 19-45-58


and this for light:
image

I like now both version more as they are less green and warmer. There would be some potential for improvement in the top right corner fold but I'm not sure if it is worth the effort.

mx990: I also like the new colors. I believe the fold is fine as is.

mx990 (approved):

PR #2168, opened 2026-02-22
Improved visual representation of Mapping

azoitl:
Filling the whole FB with the device color was original a very good concept. However it has several issues: readability depends on the device color, it is hard to show mapping status of subapps with FBs mapped to different devices (panned feature) and most important not mapped FBs are very flat and the diagrams sometimes hard to read.

In order to improve this I came up with a concept where we show the mapping status at the bottom of the FB. This first PR migrates the existing mapping features to that concept. The result for a simple dummy app looks as follows. Curios what you think:
image

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractBlockFBNElementEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractBlockFBNElementEditPart.java
@@ -163,7 +163,7 @@ public abstract class AbstractBlockFBNElementEditPart extends AbstractPositionab
 				super.notifyChanged(notification);
 				refreshToolTip(); // TODO add here checks that better define when the tooltip should be refreshed
 				if (notification.getFeature() == LibraryElementPackage.eINSTANCE.getFBNetworkElement_Mapping()) {
-					updateDeviceListener();
+					refreshChildren();
 				}
 			}
 
--- a/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractBlockFBNElementEditPart.java
+++ b/plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editparts/AbstractBlockFBNElementEditPart.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2025 Profactor GmbH, TU Wien ACIN, fortiss GmbH,
- *                          Johannes Kepler University Linz,
- *                          Primetals Technologies Germany GmbH,
- *                          Primetals Technologies Austria GmbH
+ * Copyright (c) 2008 Profactor GmbH, TU Wien ACIN, fortiss GmbH,
+ *                    Johannes Kepler University Linz,
+ *                    Primetals Technologies Germany GmbH,
+ *                    Primetals Technologies Austria GmbH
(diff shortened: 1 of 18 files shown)


mx990 (approved):
I really like the new representation, particularly on dark mode:

Image

azoitl:
@mx990 thx. I hope this will get even better when we finally fill the blocks.

bwiesmayr:
I like the new representation but see an issue with larger subapps, as one is unlikely to see the line at the bottom. Would you consider marking each mapped block? I.e., in your example, all member FBs of SubApp_1

azoitl:
> I like the new representation but see an issue with larger subapps, as one is unlikely to see the line at the bottom. Would you consider marking each mapped block? I.e., in your example, all member FBs of SubApp_1

Hmh you have a point here. The problem I see with marking each contained block is that this PR is also the basis of mapping children of SubApps to different devices. With that user will be allowed to select a Subapp child block and map or unmap it. If we we now show all children of an expanded subapp with the mapped state if the parent subapp is mapped how would users know what was mapped, the individual FBs or the subapp.

azoitl:
I today also did a first prototype for coloring of fbs and untyped subapps:
image

not darkmode yet and the code is ugly, but so that you get an idea where this is going.

bwiesmayr:
> > I like the new representation but see an issue with larger subapps, as one is unlikely to see the line at the bottom. Would you consider marking each mapped block? I.e., in your example, all member FBs of SubApp_1
> 
> Hmh you have a point here. The problem I see with marking each contained block is that this PR is also the basis of mapping children of SubApps to different devices. With that user will be allowed to select a Subapp child block and map or unmap it. If we we now show all children of an expanded subapp with the mapped state if the parent subapp is mapped how would users know what was mapped, the individual FBs or the subapp.

Hm do I need to/want to know? I would have expected that I can map a subapp to a device and still choose to map some individual FBs to other devices (for efficiency, dont want to map 100 components separately just because one element in hierarchy level X needs to go elsewhere). And then it is perfect if I can actually see where each block is running... Either based on its own mapping or, without one, based on the next parent's mapping

azoitl:
> > Hmh you have a point here. The problem I see with marking each contained block is that this PR is also the basis of mapping children of SubApps to different devices. With that user will be allowed to select a Subapp child block and map or unmap it. If we we now show all children of an expanded subapp with the mapped state if the parent subapp is mapped how would users know what was mapped, the individual FBs or the subapp.
> 
> Hm do I need to/want to know? 

You have a good point here. 
 
>I would have expected that I can map a subapp to a device and still choose to map some individual FBs to other devices (for efficiency, dont want to map 100 components separately just because one element in hierarchy level X needs to go elsewhere). And then it is perfect if I can actually see where each block is running... Either based on its own mapping or, without one, based on the next parent's mapping

Recapping I think when we will have to definitely deeper analyze the use-cases when we finally allow to map members of subapplications to different resources and devices. Your proposal sounds very sensible, even though it is the most heavy to implement and get right. Not that this should matter in an UX decision. 

For now I wanted to propose to use the current solution from this PR here, as it leads to cleaner ui and can be much better extended then the previous solution. This would also allow us to collect feedback from our users.

oberlehner (approved):

azoitl:
@bwiesmayr would you be fine that we merge it for the current development cycle to evaluate and improve till 3.2 release?

bwiesmayr:
> @bwiesmayr would you be fine that we merge it for the current development cycle to evaluate and improve till 3.2 release?

sure

azoitl:
great then I'll merge as soon as the build is through.

PR #2185, opened 2026-02-25
Create Product on manifest creation

oberlehner:


plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -260,6 +263,14 @@ public static Manifest createProjectManifest(final IProject project, final Colle
 		return manifest;
 	}
 
+	private static String toSymbolicName(final String name) {
+		final String symbolicName = name.trim().replaceAll("[^A-Za-z0-9_-]", "_"); //$NON-NLS-1$ //$NON-NLS-2$
```
mx990: Should we not also replace `-`, since it is not allowed in a 61131-3 identifier?

oberlehner: You are right, i removed it !

mx990 (approved):

plugins/org.eclipse.fordiac.ide.library.model/src/org/eclipse/fordiac/ide/library/model/util/ManifestHelper.java
```diff
@@ -260,6 +263,14 @@ public static Manifest createProjectManifest(final IProject project, final Colle
 		return manifest;
 	}
 
+	private static String toSymbolicName(final String name) {
+		final String symbolicName = name.trim().replaceAll("[^A-Za-z0-9_]", "_"); //$NON-NLS-1$ //$NON-NLS-2$
+		if (symbolicName.isEmpty()) {
```
ernstblechaPT: I don't fully understand why the fallback for an empty string is returning an empty string - did you have a different fallback in mind?

oberlehner: thx for noticing. thats some leftover code

azoitl (approved):

PR #2195, opened 2026-02-26
Modernize and Harmonize ECC colors, provide dark mode colors

azoitl:
As reference here screenshots of the new design in light and dark mode:

Screenshot From 2026-02-26 14-37-48

image

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/build.properties
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/build.properties
@@ -3,4 +3,5 @@ output.. = bin/
 bin.includes = META-INF/,\
                .,\
                plugin.xml,\
-               plugin.properties
+               plugin.properties,,\
+               css/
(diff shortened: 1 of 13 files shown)


mx990 (approved):
I really like the new look!

bwiesmayr (commented):
I find it good that the colors are getting aligned with the events again (as they are now green) and the design is very sleek and saves space

azoitl:
Thx for the feedback @mx990 @bwiesmayr. It was much harder and took much longer then I original thought it will take to come up with a good solution. Also I'm happy that @mx990 you like the dark mode. I was not sure with the algorithm color. But if you, as our dark mode user, like it then it fits.

gravemalte:
@azoitl  I just look at the dark mode as you mentioned it. The `START` state shared the same background color (at least thats what I see). I think the color of the background of the algorithm would fit better. Or changeing the font color of the `START` state to white. What do you think about it? Now for me its hard to read.

azoitl:
> @azoitl I just look at the dark mode as you mentioned it. The `START` state shared the same background color (at least thats what I see). I think the color of the background of the algorithm would fit better. Or changeing the font color of the `START` state to white. What do you think about it? Now for me its hard to read.

I get your point. Unfortunately I can not easily say if darkmode then use color x and in light y. I have color definitions and the theming api gives me then based on the theme the color. However you gave me an idea that could work.

azoitl:
Based on @gravemalte feedback I changed the center background color of the start state to the background color of the algorithm and also used the algirthm text color. This gives in my opinion the start state an even better visibility and also readability:

Screenshot From 2026-02-27 16-23-03

image

WDYT?

MartinMelikMerkumians:
Looks good. A much more modern feeling

MartinMelikMerkumians (approved):

PR #2196, opened 2026-02-26
Check if file exists before attempting to load from it in type entry

mx990:
This avoids flooding the log with load errors for files that do not exist.


azoitl (commented):
Do we have any chance to reduce the code duplication, e.g, by turning this method into a generic and provide the getters, setters and loaders via lambdas?

mx990:
I agree that de-duplicating this makes sense in general, but would suggest to do that in a separate PR on develop.

azoitl (approved):
I fear your right to move the deduplication back to an own commit on develop. This is most probably bigger then I original thought when I wrote the message.

PR #2197, opened 2026-02-27
Splash screen for 3.1 release

azoitl:
This is a just reusing the 3.0 splash screen for 3.1. As I'm currently not as creative and I saw that Eclipse itself is always using the same splash screen for the 4 release I thought we could do the same. So we would have a new splash screen for the 2026-12 release. But I'm also open to other ideas.


bwiesmayr (approved):
I think it makes sense to keep it for a while

mx990 (approved):

Issue #2205, opened 2026-03-02
Unhandled event loop exception


oberlehner:
Read only File in a Basic Block inside the ECC editor

**4diac IDE version:** 3.1.0.qualifier

**Session ID:** ae8fbde4-6e2a-4541-94e6-887c5bbd65ba

### Stack trace

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.draw2d.Viewport.getClientArea(org.eclipse.draw2d.geometry.Rectangle)" because "port" is null
	at org.eclipse.gef.editparts.ViewportAutoexposeHelper.detect(ViewportAutoexposeHelper.java:85)
	at org.eclipse.gef.AutoexposeHelper$Search.evaluate(AutoexposeHelper.java:98)
	at org.eclipse.gef.ui.parts.GraphicalViewerImpl$1ConditionalTreeSearch.accept(GraphicalViewerImpl.java:180)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:448)
	at org.eclipse.draw2d.Layer.findFigureAt(Layer.java:61)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:444)
	at org.eclipse.draw2d.Layer.findFigureAt(Layer.java:61)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:444)
	at org.eclipse.draw2d.Layer.findFigureAt(Layer.java:61)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:444)
	at org.eclipse.draw2d.Layer.findFigureAt(Layer.java:61)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:444)
	at org.eclipse.draw2d.Layer.findFigureAt(Layer.java:61)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
	at org.eclipse.draw2d.Figure.findFigureAt(Figure.java:444)
	at org.eclipse.draw2d.Figure.findDescendantAtExcluding(Figure.java:407)
[...]

PR #2209, opened 2026-03-02
Fix EC state placement under zoom updated

vikash1703:
Cause:
The mouse location was used without compensating for the current zoom level,
which resulted in incorrect coordinate calculations.

Solution:
Adjusted the coordinate calculation to account for the current zoom factor
before creating the EC state.

Testing:
- Tested state creation at 100% zoom
- Tested state creation at zoom in (150%+)
- Tested state creation at zoom out (<100%)
- Verified transitions still work correctly

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
@@ -27,6 +27,7 @@ import org.eclipse.fordiac.ide.model.libraryElement.ECState;
 import org.eclipse.fordiac.ide.model.libraryElement.Position;
 import org.eclipse.gef.DefaultEditDomain;
 import org.eclipse.gef.EditPartViewer;
+import org.eclipse.gef.GraphicalEditPart;
 import org.eclipse.gef.SharedCursors;
 import org.eclipse.gef.commands.CompoundCommand;
 import org.eclipse.gef.editparts.FreeformGraphicalRootEditPart;
@@ -106,17 +107,21 @@ final class ECCEditorEditDomain extends DefaultEditDomain {
 			handleMove();
 		}
 
-		public void performCreation() {
+		public void performCreation(final EditPartViewer viewer) {
 			final ECState destState = (ECState) getFactory().getNewObject();
+			final Point relativePoint = point.getCopy();
 
-			final Position pos = CoordinateConverter.INSTANCE.createPosFromScreenCoordinates(point.x, point.y);
+			final Object eccEditPart = viewer.getEditPartRegistry().get(getECC());
+			if (eccEditPart instanceof final GraphicalEditPart gep) {
+				gep.getFigure().translateToRelative(relativePoint);
+			}
 
+			final Position pos = CoordinateConverter.INSTANCE.createPosFromScreenCoordinates(relativePoint.x,
+					relativePoint.y);
 			final CreateECStateCommand createStateCommand = new CreateECStateCommand(destState, pos, getECC());
-
 			final CreateTransitionCommand createTransitionCommand = new CreateTransitionCommand(sourceState, destState,
 					null);
-
-			createTransitionCommand.setDestinationLocation(point);
+			createTransitionCommand.setDestinationLocation(relativePoint);
 
 			final CompoundCommand compCom = new CompoundCommand();
 			compCom.add(createStateCommand);
@@ -188,7 +193,7 @@ final class ECCEditorEditDomain extends DefaultEditDomain {
 				transitionStateCreationTool
 						.setLocationActivation(((ECCPanningSelectionTool) getDefaultTool()).getLastLocation());
 				setActiveTool(transitionStateCreationTool);
-				transitionStateCreationTool.performCreation();
+				transitionStateCreationTool.performCreation(viewer);
 				setActiveTool(getDefaultTool());
 				createTransitionAndState = false;
 			}


.gitignore
```diff
@@ -13,3 +13,4 @@
 **/.META-INF_MANIFEST.MF
 **/.tycho-consumer-pom.xml
 **/pom.tycho
+features/.project
```
gravemalte: Please dont add the .project file to the `.gitignore` it ensures that we all use the same configuration. If it bothers you in the diff, put the ignore in your own git config.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -106,24 +105,31 @@ public void setLocationActivation(final Point point) {
...
 
 			final CreateECStateCommand createStateCommand = new CreateECStateCommand(destState, pos, getECC());
-
 			final CreateTransitionCommand createTransitionCommand = new CreateTransitionCommand(sourceState, destState,
 					null);
-
 			createTransitionCommand.setDestinationLocation(point);
 
-			final CompoundCommand compCom = new CompoundCommand();
+			final org.eclipse.gef.commands.CompoundCommand compCom = new org.eclipse.gef.commands.CompoundCommand();
```
gravemalte: Can you import the `CompoundCOmmand` or does it conflict with another import that has the same class name?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -106,24 +105,31 @@ public void setLocationActivation(final Point point) {
 			handleMove();
 		}
 
-		public void performCreation() {
+		public void performCreation(final org.eclipse.gef.EditPartViewer viewer) {
+			if (viewer == null || point == null || sourceState == null) {
+				return;
+			}
+			if (!(viewer.getRootEditPart() instanceof final org.eclipse.gef.editparts.ScalableFreeformRootEditPart root)) {
```
gravemalte: Also here, an import makes this more readable (if the import is possible, but I think it is).

gravemalte:
I think you have to rework the logic. When right-clicking with the context menu and placing a state with more or less than 100% zoom level it doesn't work, the state will be placed somewhere else. ~~When holding ctlr + left-clicking I get a NPE.~~ For an easier debug for you, I recorded my screen and I paste you the NPE.

**UPDATE:** The NPE is not related to your PR and was reported in #2205 


Screen recording:

[Screencast_20260303_131452.webm](https://github.com/user-attachments/assets/800614f7-c4b6-4c9c-b06c-be294056cbf6)

vikash1703:
Hi @gravemalte 
yes you saying right  i make changes and correct the problem ,Thanks for video

azoitl:
> Hi @gravemalte yes you saying right i make changes and correct the problem ,Thanks for video

@vikash1703 as the problem  @gravemalte shows is not caused by your changes, please do not fix it as part of this PR. If you want to do it please do it on top of 3.1.x branch as this is relevant for the 3.1 release.

Also please not that you should regularly update your upstream and rebase your changes on top of upstream as explained in the updated contributing guidelines. Otherwise we will not be able to merge your PR even if it works.

.gitignore
vikash1703: Hi  @gravemalte and @azoitl,

I have successfully rebased my changes on top of the latest upstream/develop and cleaned up the commit history. The PR now contains a single clean commit focused only on the zoom placement fix.

I've also ensured that no unrelated changes from the other issue are included here, as suggested. Ready for your final review. Thank you!

vikash1703:
Hi @azoitl and @gravemalte 

I noticed the build failed in the test.ui bundle. Since the core bundles where I made changes (ECC and Commands) passed successfully, it seems like a pre-existing environment issue in the CI. My refactoring is complete and ready for review.

azoitl:
> Hi @azoitl and @gravemalte
> 
> I noticed the build failed in the test.ui bundle. Since the core bundles where I made changes (ECC and Commands) passed successfully, it seems like a pre-existing environment issue in the CI. My refactoring is complete and ready for review.

@vikash1703 I feear the problem is caused by you. your code is not compiling because you didn't resolve the merge conflicts. you see this with the  `>>>>>` and similar line.

vikash1703:
Hi @azoitl and @gravemalte  now its good to go after it will get successfully merged i want to explore the next problem as you show in the video there is still problem as we can create state while zoom whats your opinion @azoitl

.gitignore
azoitl: I want to point out that @gravemalte is totally right. This update to gitignore is not only unrelated to the PR it is even dangerous. The reason you have a `features/.project` is that you accidently imported the wrong projects. You should delete the `features/.project` file and not mask it from git.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -106,31 +105,39 @@ public void setLocationActivation(final Point point) {
...
-			createTransitionCommand.setDestinationLocation(point);
+			createTransitionCommand.setDestinationLocation(translated);
 
 			final CompoundCommand compCom = new CompoundCommand();
 			compCom.add(createStateCommand);
 			compCom.add(createTransitionCommand);
 
-			setCurrentCommand(compCom);
-			performCreation(1);
+			viewer.getEditDomain().getCommandStack().execute(compCom);
```
azoitl: this change violates the contract of the GEF Tool class.

azoitl (changes requested):
This PR tries to do at least two thing. Maybe more. Therefore this should be split up into several PRs. Furthermore I noticed that you removed our comments. As the code has not changed in this places this is a very bad practice. 

Finally please also be more descriptive in your commit messages.

gravemalte:
Also I just check the PR did you fixed the functionality or just rebased? Because when zooming again in and out and trying to place states with the mouse menu they appear somewhere else, but not at the mouse location.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
vikash1703, line 1: Hi @azoitl and @gravemalte 

1. ECCEditorEditDomain.java 
 Use Draw2D Point for destination location to ensure proper coordinate mapping

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CoordinateConverter.java
vikash1703, line 1: @azoitl and @gravemalte 

CoordinateConverter.java

Precision Update: Refined createPosFromScreenCoordinates by adding Math.round() to the coordinate conversion. This prevents rounding errors when translating screen pixels to IEC 61499 model units.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -1,23 +1,21 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2025 Johannes Kepler University Linz, fortiss GmbH
```
azoitl, line 2: please don't invent copy right holders and years. With the latest Eclipse project handbook rules, only the first year needs to be present, as you did a significant change you can add you at the end of the existing list.  But you can not just change years and add names.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -1,23 +1,21 @@
...
+ * Copyright (c) 2014, 2021, 2026 Johannes Kepler University Linz, 
+ * TU Wien ACIN, Fortiss GmbH
+ * * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
  *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
- *   Alois Zoitl - initial API and implementation and/or initial documentation
```
azoitl, line 11: As allready said seveal times: DONT'T CHANGE UNRELATED COMMENTS!

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/CoordinateConverter.java
```diff
@@ -116,8 +116,8 @@ public double getAverageCharacterWidth() {
 
 	public Position createPosFromScreenCoordinates(final int x, final int y) {
 		final Position pos = LibraryElementFactory.eINSTANCE.createPosition();
-		pos.setX(screenToIEC61499(x));
-		pos.setY(screenToIEC61499(y));
+		pos.setX((int) Math.round(screenToIEC61499(x)));
```
azoitl: This is a very very very bad idea. Position stores the coordinates in double and there is a good reason for that. We need at least two digits after the comma on modern screens.

azoitl (changes requested):
I don not understand why you now change everything to just fix the position. This looks wrong and just to big of a change. The previous states of this PR looked much more promising.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -187,13 +193,13 @@ public void mouseUp(final MouseEvent mouseEvent, final EditPartViewer viewer) {
 					.getTargetEditPart() instanceof FreeformGraphicalRootEditPart) {
 				transitionStateCreationTool
 						.setLocationActivation(((ECCPanningSelectionTool) getDefaultTool()).getLastLocation());
-				setActiveTool(transitionStateCreationTool);
-				transitionStateCreationTool.performCreation();
-				setActiveTool(getDefaultTool());
+
```
azoitl: I would prefere if the setActiveTool  lines. I also think that you then don't need the viewer as paramter.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -187,13 +193,13 @@ public void mouseUp(final MouseEvent mouseEvent, final EditPartViewer viewer) {
 					.getTargetEditPart() instanceof FreeformGraphicalRootEditPart) {
 				transitionStateCreationTool
 						.setLocationActivation(((ECCPanningSelectionTool) getDefaultTool()).getLastLocation());
-				setActiveTool(transitionStateCreationTool);
-				transitionStateCreationTool.performCreation();
-				setActiveTool(getDefaultTool());
+				
+				// Change 3: Calling performCreation with viewer for zoom context
```
azoitl: not helping comment, remove

azoitl (changes requested):
Again you have changes in unrelated files.

vikash1703:
Hi @azoitl, I have reset the branch to the default state to clear all previous changes. Tomorrow, I will submit a minimal PR with the fix specifically in the TransitionStateCreationTool

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
vikash1703, line 1: @azoitl 
Fix transition creation position when creating a new state via drag.

The mouse location was previously used directly in screen coordinates,
which caused incorrect placement when zoom or pan was applied.

Now the point is translated to the editor's relative coordinate system
before creating the state and transition.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditor.java
vikash1703, line 1: Added getContentPane() method to ECCEditor: Exposes the viewer's drawing layer to allow standard coordinate translation.

Standardized Coordinate Mapping: Enables replacing manual viewport calculations with the framework's translateToRelative() method.

azoitl: I would prefere that we are not adding a new interface. in the edit domain and tools you have access to the viewer. You could use the model to edipart mapping table to get the editpart responsible for drawing the ECC and from that the figure for translateToRelative. With that we are more flexible against future changes of the editor and figures. E.g. if we ever get hierarchical state machines it would still work.

azoitl (commented):
I think you lost all your previous code.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
vikash1703, line 1: Hi @azoitl 

The issue was in TransitionStateCreationTool.performCreation() where the coordinates were used directly without accounting for zoom and viewport offset.
I fixed this by passing the viewer to setLocationActivation() and using translateToRelative() on the ECC figure obtained from the viewer's edit part ,I replaced setCurrentCommand/performCreation(1) with a direct commandStack.execute() call to avoid tool activation issues that were causing a Viewport crash when hover.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -91,6 +92,7 @@ public void changeCursor() {
 	private static class TransitionStateCreationTool extends CreationTool {
 		private Point point;
 		private ECState sourceState;
+		private EditPartViewer storedViewer;
```
azoitl: you don't need this.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -101,29 +103,38 @@ public void setSourceState(final ECState state) {
 			this.sourceState = state;
 		}
 
-		public void setLocationActivation(final Point point) {
+		public void setLocationActivation(final Point point, final EditPartViewer viewer) {
 			this.point = point;
-			handleMove();
+			this.storedViewer = viewer;
+			// handleMove();
```
azoitl: this should definitly not be commented out.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -101,29 +103,38 @@ public void setSourceState(final ECState state) {
 			this.sourceState = state;
 		}
 
-		public void setLocationActivation(final Point point) {
+		public void setLocationActivation(final Point point, final EditPartViewer viewer) {
```
azoitl: this should not be needed. if it is the problem is somewhere else

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -101,29 +103,38 @@ public void setSourceState(final ECState state) {
...
+			this.storedViewer = viewer;
+			// handleMove();
 		}
 
 		public void performCreation() {
 			final ECState destState = (ECState) getFactory().getNewObject();
+			final Point relativePoint = point.getCopy();
+
+			if (storedViewer != null && sourceState != null) {
+				final Object eccEditPart = storedViewer.getEditPartRegistry().get(getECC());
```
azoitl: the editpart registry holds already an editpart and not an object

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -101,29 +103,38 @@ public void setSourceState(final ECState state) {
...
-			createTransitionCommand.setDestinationLocation(point);
+			createTransitionCommand.setDestinationLocation(relativePoint);
 
 			final CompoundCommand compCom = new CompoundCommand();
 			compCom.add(createStateCommand);
 			compCom.add(createTransitionCommand);
 
-			setCurrentCommand(compCom);
-			performCreation(1);
+			if (storedViewer != null) {
```
azoitl: please replace with the original version.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
```diff
@@ -185,9 +196,7 @@ public void mouseUp(final MouseEvent mouseEvent, final EditPartViewer viewer) {
 		if (transition) {
 			if (createTransitionAndState && ((AdvancedPanningSelectionTool) getDefaultTool())
 					.getTargetEditPart() instanceof FreeformGraphicalRootEditPart) {
-				transitionStateCreationTool
-						.setLocationActivation(((ECCPanningSelectionTool) getDefaultTool()).getLastLocation());
-				setActiveTool(transitionStateCreationTool);
+				transitionStateCreationTool.setLocationActivation(new Point(mouseEvent.x, mouseEvent.y), viewer);
```
azoitl: the original version was definitely correcter.

vikash1703:
> Also I just check the PR did you fixed the functionality or just rebased? Because when zooming again in and out and trying to place states with the mouse menu they appear somewhere else, but not at the mouse location.

Hi, I tested it locally with zoom enabled. The fix is in ECCEditorEditDomain where I replaced the direct coordinate usage with translateToRelative() via the viewer's edit part. States are now placed at the correct position with zoom.
here is the attach reference video 

https://github.com/user-attachments/assets/a0ed7ce7-a9f5-42b0-9242-95c8ad3865d4

azoitl (changes requested):
@vikash1703 after seeing that you are spaming us with this PR and going in circles.
I'm not sure if you will be able to solve this problem with Sonet. I noticed when using AI tools to improve 4diac IDE the best is to use their input as reference and get ideas how to do it. However manually fixing and improving things worked sofar always better.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditorEditDomain.java
vikash1703: > the original version was definitely correcter.

Hi @azoitl,


1. I have addressed all your feedback on ECCEditorEditDomain
2. I have Removed the storedViewer field it was not needed but i added.
3. handleMove() is back I was wrongly commented out before
4. setLocationActivation() is back to original signature 
5. setCurrentCommand(compCom) and performCreation(1) are back now it ensures GEF contract
6. mouseUp() uses getLastLocation() to original version as you said was correct
7. I have passes  viewer directly from mouseUp() to performCreation(viewer)  this is the only change. Since TransitionStateCreationTool is a static inner class it cannot access the outer class, so passing viewer as parameter to performCreation() was the minimal way to fix this.

azoitl (approved):
@vikash1703 thank you for cleaning up and taking my feedback in consideration. I really hoped that we could get rid of the viewer parameter in the `performCreation()` method. However it looks like that we are not yet correctly using the command. For now this is fine.

The reason I mentioned that is that for fixing the problem of switching the creation modes (press ctrl/release ctrl) we will need to come back to that code. But maybe that is anyhow a bigger refinement.  In the FBNetworkEditor we solved that inside of the command. Maybe we will need to do that anyhow there.

vikash1703:
> @vikash1703 thank you for cleaning up and taking my feedback in consideration. I really hoped that we could get rid of the viewer parameter in the `performCreation()` method. However it looks like that we are not yet correctly using the command. For now this is fine.
> 
> The reason I mentioned that is that for fixing the problem of switching the creation modes (press ctrl/release ctrl) we will need to come back to that code. But maybe that is anyhow a bigger refinement. In the FBNetworkEditor we solved that inside of the command. Maybe we will need to do that anyhow there.

Hi @azoitl
At that time I did not think of that but you are absolutely right, we need to fix this. I will look into how FBNetworkEditor solved this inside the command and try to apply the same approach to remove the viewer parameter from performCreation() in a future PR.

PR #2216, opened 2026-03-03
Library import rework: unified wizard + GitLab secure endpoints 

oberlehner:
This PR modernizes library import in 4diac by consolidating multiple entry points into one Unified Library Import wizard and by introducing multi-endpoint GitLab configuration.

**One wizard to import libraries from:**
- Workspace Libraries
- File System Libraries
- GitLab endpoints (configured in preferences)

**GitLab Endpoints preference page:**
- Manage multiple endpoints (name/url/token)
- Test connection from the UI
- Tokens stored securely in Equinox Secure Storage; only name/url persisted in plain prefs
- Migration from legacy gitLabURL/gitLabToken preferences

**Cleanup:**

1. Removes old GitLab import wizard and the legacy single-endpoint preference page
2. Improves manifest load diagnostics
3. Deletes downloaded GitLab ZIP folder after successful extraction


plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitLabEndpointsPreferencePage.java
```diff
@@ -0,0 +1,301 @@
...
+		// validate all
+		for (final GitLabEndpoint ep : endpoints) {
+			if (ep == null || !ep.isValid()) {
+				setErrorMessage(Messages.GitLabEndpointsPreferencePage_all_endpoints);
+				return false;
+			}
+		}
+		setErrorMessage(null);
+		GitLabEndpointsStore.saveEndpoints(endpoints);
+		return returnValueOnSuccess;
```
mx990: Why not always return true on success?

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitLabEndpointsStore.java
```diff
@@ -0,0 +1,272 @@
...
+				}
+				final String name = Objects.toString(ep.name(), "").trim(); //$NON-NLS-1$
+				final String token = Objects.toString(ep.token(), "").trim(); //$NON-NLS-1$
+				if (!name.isEmpty() && !token.isEmpty()) {
+					putTokenSecure(name, token);
+				}
+			}
+		}
+
+		prefs.put(PreferenceConstants.P_GITLAB_ENDPOINTS, encodeNameUrlOnly(endpoints));
```
mx990: How about storing each endpoint url under a separate node, same as for the tokens, instead of encoding all endpoints into a single string?

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitLabEndpointsStore.java
```diff
@@ -0,0 +1,272 @@
...
+					.node(endpointNodeId(endpointName));
+			return node.get(SECURE_TOKEN_KEY, "");
+		} catch (final Exception e) {
+			return "";
+		}
+	}
+
+	/**
+	 * Secure storage node names are path components. Encode name to avoid '/'
+	 * issues.
```
mx990: How about we just disallow special characters in endpoint names and use the name as-is?

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/Messages.java
```diff
@@ -28,6 +28,66 @@ public final class Messages extends NLS {
...
+	public static String GitLabEndpointDialog_token;
+
+	public static String GitLabEndpointDialog_token_note_empty;
+
+	public static String GitLabEndpointDialog_url;
+
+	public static String GitLabEndpointDialog_url_not_empty;
+	public static String GitLabEndpointsPreferencePage_16;
+
+	public static String GitLabEndpointsPreferencePage_20;
```
mx990: Please use meaningful message keys.

plugins/org.eclipse.fordiac.ide.gitlab/plugin.xml
```diff
@@ -1,32 +1,24 @@
...
+            name="GitLab Endpoints">
       </page>
    </extension>
+   <extension point="org.eclipse.ui.preferencePages">
+  <page
+      id="org.eclipse.fordiac.ide.gitlab.preferences.gitlabEndpoints"
+      name="GitLab Endpoints"
+      class="org.eclipse.fordiac.ide.gitlab.preferences.GitLabEndpointsPreferencePage"
+      category="org.eclipse.fordiac.ide">
+  </page>
```
mx990: It looks like this is duplicate somehow?

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/GitLabEndpointSource.java
mx990, line 1: Would this not better fit in `org.eclipse.fordiac.ide.gitlab`?

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/ILibrarySource.java
```diff
@@ -0,0 +1,42 @@
...
+
+	boolean isSelectableLeaf(Object element);
+
+	void install(IProject targetProject, Collection<?> selectedLeafElements, IProgressMonitor monitor) throws Exception;
+
+	default void dispose() {
+		// optional
+	}
+
+	default String exclusiveVersinSelectionKey(final Object element) {
```
mx990:
```suggestion
	default String exclusiveVersionSelectionKey(final Object element) {
```

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/ILibrarySource.java
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.library.ui.sources;
+
+import java.util.Collection;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.swt.widgets.Composite;
+
+public interface ILibrarySource {
```
mx990, line 21: This interface seems to mix both UI and non-UI concerns. Could we separate the non-UI concerns (loading and installing) into a separate interface or add this into `IArchiveDownloader`?

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/LibrarySourceUIComponents.java
```diff
@@ -0,0 +1,29 @@
...
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.library.ui.sources;
+
+import java.util.Objects;
+
+import org.eclipse.jface.viewers.ILabelProvider;
+import org.eclipse.jface.viewers.ITreeContentProvider;
+
+public record LibrarySourceUIComponents(ITreeContentProvider contentProvider, ILabelProvider labelProvider,
+		Object input, Object context) {
```
mx990: How about using a common `WorkbenchContentProvider` and `WorkbenchLabelProvider` to provide the individual information via an `IWorkbenchAdapter`, instead of specifying explicit providers here?

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/Messages.java
```diff
@@ -39,6 +39,84 @@ public class Messages extends NLS {
...
+
+	public static String LibrarySourceBuilder_2;
+
+	public static String LibrarySourceBuilder_3;
+
+	public static String LibrarySourceBuilder_4;
+
+	public static String LibrarySourceBuilder_5;
+
+	public static String LibrarySourceBuilder_6;
```
mx990: Please use meaningful message keys.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -510,10 +510,15 @@ private DownloadResult<java.net.URI> libraryDownload(final String symbolicName,
...
 		FordiacLogHelper.logInfo("Attempting to download library " + symbolicName + " with version " + versionRange //$NON-NLS-1$ //$NON-NLS-2$
-				+ " preferring " + preferred); //$NON-NLS-1$
+				+ " preferring " + preferred + " Project: " + project != null ? project.getName() : ""); //$NON-NLS-1$
 
-		final List<IArchiveDownloader> downloaders = TypeLibraryManager.listExtensions(DOWNLOADER_EXTENSION,
+		List<IArchiveDownloader> downloaders = TypeLibraryManager.listExtensions(DOWNLOADER_EXTENSION,
 				IArchiveDownloader.class);
+
+		if (downloaders.size() == 1 && downloaders.get(0).hasMultipleEndpoints()) {
+			downloaders = downloaders.get(0).convertEndpointsToDownloader();
```
mx990, line 519: What if we have more that one downloader with multiple endpoints?

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/wizard/New4diacProjectWizard.java
```diff
@@ -83,16 +81,21 @@ public boolean performFinish() {
 				@Override
 				protected void execute(final IProgressMonitor monitor) {
 					createProject(monitor != null ? monitor : new NullProgressMonitor());
+					//
```
mx990: What is this comment for?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
oberlehner: The downloader should anyway be restructured, since this solution is not so nice. 
 It will be refatored, so that a downloader can have multiple endpoints.
 But this will be part of a follow up PR. 
 I will add a comment so that we prevent misuse.

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/LibrarySourceUIComponents.java
oberlehner: Good point, i changed that !

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitLabEndpointsStore.java
oberlehner: Done

plugins/org.eclipse.fordiac.ide.gitlab/src/org/eclipse/fordiac/ide/gitlab/preferences/GitLabEndpointsStore.java
oberlehner: Done

plugins/org.eclipse.fordiac.ide.library.ui/src/org/eclipse/fordiac/ide/library/ui/sources/ILibrarySource.java
oberlehner: I agree that the interface currently mixes UI and operational concerns. For this PR, I’d prefer to keep it as-is to avoid a broader refactoring, since splitting it would touch all current source implementations and the wizard integration without changing the actual behavior.

mx990 (approved):

PR #2217, opened 2026-03-03
fix: stabilize parameterized test names in ValueConverterTest

gravemalte:
Prevents GitHub PR noise where renamed tests were reported as 73 additions and 73 deletions. This was caused by the default JUnit display name including the ValueConverter's memory address (hashcode).

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/value/StringValueConverter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/value/StringValueConverter.java
@@ -127,6 +127,11 @@ public final class StringValueConverter implements ValueConverter<String> {
 		return result.toString();
 	}
 
+	@Override
+	public String toString() {
+		return getClass().getSimpleName();
+	}
+
 	private static void escape(final int value, final StringBuilder result) {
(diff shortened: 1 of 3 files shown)


mx990 (changes requested):
The other value converters have corresponding `toString()` methods that produce a stable output, see also https://github.com/eclipse-4diac/4diac-ide/blob/cde6986de47b6e75c15828a0cf841040840950f5/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/value/NumericValueConverter.java#L108-L113 and https://github.com/eclipse-4diac/4diac-ide/blob/cde6986de47b6e75c15828a0cf841040840950f5/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/value/BoolValueConverter.java#L48-L51

I would prefer to add the missing `toString()` methods for `StringValueConverter` and `WStringValueConverter` rather than not show the name of the converter at all.

gravemalte:
> I would prefer to add the missing `toString()` methods for `StringValueConverter` and `WStringValueConverter` rather than not show the name of the converter at all.

Also fine for me I change that.

gravemalte:
@mx990 Ready to rereview.

mx990 (approved):

PR #2226, opened 2026-03-05
[ECC] Fix zoom-aware coordinate translation in NewStateAction

vikash1703:
Fix coordinate translation when creating ECC states via right-click context menu with zoom enabled.

Previously, the code manually calculated diagram coordinates using viewport 
offset and zoom factor (getViewLocation() + scale(1/zoom)). This caused newly 
created states to appear at incorrect positions when the canvas was zoomed.

The fix uses GEF's built-in translateToRelative() method on the ECC EditPart's 
figure, which correctly handles zoom, viewport offset, and scrolling. The ECC 
EditPart is looked up from the viewer's edit part registry via the IAdaptable 
interface, avoiding any direct dependency on editor internals.

Affected: NewStateAction.run()

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
@@ -17,13 +17,14 @@ import org.eclipse.draw2d.FigureCanvas;
 import org.eclipse.draw2d.geometry.Point;
 import org.eclipse.fordiac.ide.fbtypeeditor.ecc.Messages;
 import org.eclipse.fordiac.ide.fbtypeeditor.ecc.commands.CreateECStateCommand;
-import org.eclipse.fordiac.ide.fbtypeeditor.ecc.editors.ECCEditor;
 import org.eclipse.fordiac.ide.fbtypeeditor.ecc.editors.StateCreationFactory;
 import org.eclipse.fordiac.ide.model.CoordinateConverter;
+import org.eclipse.fordiac.ide.model.libraryElement.ECC;
 import org.eclipse.fordiac.ide.model.libraryElement.ECState;
 import org.eclipse.fordiac.ide.model.libraryElement.Position;
 import org.eclipse.fordiac.ide.ui.imageprovider.FordiacImage;
-import org.eclipse.gef.editparts.ZoomManager;
+import org.eclipse.gef.GraphicalEditPart;
+import org.eclipse.gef.GraphicalViewer;
 import org.eclipse.gef.ui.actions.WorkbenchPartAction;
 import org.eclipse.ui.IWorkbenchPart;
 
@@ -38,7 +39,6 @@ public class NewStateAction extends WorkbenchPartAction {
 	private static StateCreationFactory stateFactory = new StateCreationFactory();
 	private FigureCanvas viewerControl;
 	private org.eclipse.swt.graphics.Point pos = new org.eclipse.swt.graphics.Point(0, 0);
-	private ZoomManager zoomManager;
 
 	public NewStateAction(final IWorkbenchPart part) {
 		super(part);
@@ -54,10 +54,6 @@ public class NewStateAction extends WorkbenchPartAction {
 		}
 	}
 
-	public void setZoomManager(final ZoomManager zoomManager) {
-		this.zoomManager = zoomManager;
-	}
-
 	@Override
 	protected boolean calculateEnabled() {
 		return true; // we can always be enabled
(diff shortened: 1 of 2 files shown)


azoitl (commented):
If you remove methods you also need to check for any callers. I guess you are not compiling and testing on your side.

vikash1703:
@azoitl 
You are right, I apologize. I have fixed it by also removing the setZoomManager() 
call from ECCEditor.configureGraphicalViewer() where it was being called. 
I will make sure to check for all callers before removing any method in future changes.

vikash1703:
> If you remove methods you also need to check for any callers. I guess you are not compiling and testing on your side.

I thik CI failure in fix/ecc-new-state is because develop still has the old ECCEditor.java with the setZoomManager() call. The fix/ecc-editor-adapter PR removes that call. If it merges fix/ecc-editor-adapter the i think it will pass all tests

vikash1703:
> If you remove methods you also need to check for any callers. I guess you are not compiling and testing on your side.

In NewStateAction:  I have removed  the manual zoom calculation (getViewLocation() and  scale(1/zoom)). 
Now  we are using translateToRelative() on the ECC figure from the viewer's edit part and I did not added any extra 
things which violate ECC and 4diac rule.


In ECCEditor: I have removed setZoomManager() call as it is no longer needed and I have added EditPartViewer in getAdapter() It is required because NewStateAction using editor.getAdapter(EditPartViewer.class) to access the viewer.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
```diff
@@ -66,17 +62,19 @@ protected boolean calculateEnabled() {
 	@Override
 	public void run() {
 		final ECCEditor editor = (ECCEditor) getWorkbenchPart();
+		final Point clickPos = new Point(pos.x, pos.y);
 
-		final Point location = viewerControl.getViewport().getViewLocation();
-		final Point realPos = new Point(pos.x + location.x, pos.y + location.y);
-		realPos.scale(1.0 / zoomManager.getZoom());
+		final EditPartViewer viewer = editor.getAdapter(EditPartViewer.class);
```
azoitl: our Graphical editors are adapting on `GraphicalViewer` using this gives you the same API without the need to change the ECCEditor.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
```diff
@@ -66,17 +62,19 @@ protected boolean calculateEnabled() {
 	@Override
 	public void run() {
 		final ECCEditor editor = (ECCEditor) getWorkbenchPart();
```
azoitl: I would prefere not to have this cast. This will help us later improving the code and better decouple editor from handlers and actions. See below for the solution.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
```diff
@@ -66,17 +62,19 @@ protected boolean calculateEnabled() {
...
+		final Point clickPos = new Point(pos.x, pos.y);
 
-		final Point location = viewerControl.getViewport().getViewLocation();
-		final Point realPos = new Point(pos.x + location.x, pos.y + location.y);
-		realPos.scale(1.0 / zoomManager.getZoom());
+		final EditPartViewer viewer = editor.getAdapter(EditPartViewer.class);
+		if (viewer != null) {
+			final Object eccEditPart = viewer.getEditPartRegistry().get(editor.getModel());
```
azoitl: our ECCEditor adapts on ECC. So using `editor.getAdapter(ECC.class)` should work here and would make our code more maintainable.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/actions/NewStateAction.java
vikash1703: > our ECCEditor adapts on ECC. So using `editor.getAdapter(ECC.class)` should work here and would make our code more maintainable.

Hi @azoitl 
Now I have Removed ECCEditor cast, now using getWorkbenchPart().getAdapter(GraphicalViewer.class) for viewer and getWorkbenchPart().getAdapter(ECC.class) for ECC model. Also removed EditPartViewer case from ECCEditor.getAdapter().

azoitl: Thx for fixing my concerns. Looks good now.

azoitl (approved):

PR #2227, opened 2026-03-05
[ECC] Add EditPartViewer adapter to ECCEditor for zoom-aware coordina…

vikash1703:
@azoitl and @gravemalte 
I have removed zoomManager() method as it is no longer needed as i replace with  translateToRelative() as suggested and Added EditPartViewer as an adaptable type in getAdapter() so that NewStateAction can access the viewer without any direct dependency on editor internals.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditor.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditor.java
@@ -49,6 +49,7 @@ import org.eclipse.gef.DefaultEditDomain;
 import org.eclipse.gef.DragTracker;
 import org.eclipse.gef.EditPart;
 import org.eclipse.gef.EditPartFactory;
+import org.eclipse.gef.EditPartViewer;
 import org.eclipse.gef.KeyHandler;
 import org.eclipse.gef.KeyStroke;
 import org.eclipse.gef.LayerConstants;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditor.java
```diff
@@ -286,6 +286,9 @@ public <T> T getAdapter(final Class<T> adapter) {
 		if (adapter == ECC.class) {
 			return adapter.cast(getModel());
 		}
+		if (adapter == EditPartViewer.class) {
```
azoitl, line 289: I don't see why this should be here please remove

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/editors/ECCEditor.java
```diff
@@ -126,7 +127,6 @@ protected void configureGraphicalViewer() {
 		// position for state creation
 		final IAction action = getActionRegistry().getAction(NewStateAction.CREATE_STATE);
 		((NewStateAction) action).setViewerControl((FigureCanvas) viewer.getControl());
-		((NewStateAction) action).setZoomManager(getZoomManger());
```
azoitl, line 129: please move this to your PR #2226 where it belongs to as otherwise  #2226 is not complete and this here is not a working independent PR.

azoitl (commented):
As this is not an independent PR and the change here is partly wrong and the other part belongs to #2226 I'm closing it.

PR #2229, opened 2026-03-08
Remove explicit versions from Junit dependencies in Target Platform

azoitl:


mx990 (approved):
I used to get lots of errors when not specifying the exact version, but it appears to be working now.

azoitl:
> I used to get lots of errors when not specifying the exact version, but it appears to be working now.

We had a maven dependency that I recently removed. Maybe that fixed that as well.

PR #2237, opened 2026-03-09
Updated deployment interpreter Version for 3.2 Release Cycle

cochicde:


cochicde:
I just copied what you did in your commit @azoitl

azoitl (commented):
@cochicde sorry for the unclear specification. You should:
  - for every plugin where you added larger features increase the minor version i.e., 3.2
  -  in my PR I allready should have updated to maintainence version. This is for small changes that fixe bugs.

please also consider the other plugins where your two big PRs I just merged had changes.

cochicde:
> @cochicde sorry for the unclear specification. You should:
> 
> * for every plugin where you added larger features increase the minor version i.e., 3.2
> * in my PR I allready should have updated to maintainence version. This is for small changes that fixe bugs.
> 
> please also consider the other plugins where your two big PRs I just merged had changes.

I see. What about now?

azoitl (approved):
LGTM

azoitl:
What I don't know is what version number we should give new plugins. In my just submitted draft PR #2241 I used 3.0 but I'm not sure what is there a good practice. Open for suggestions.

cochicde:
> What I don't know is what version number we should give new plugins. In my just submitted draft PR #2241 I used 3.0 but I'm not sure what is there a good practice. Open for suggestions.

I'm not sure about  a good practice for eclipse plugins. For me, if it doesn't hurt much, I'd keep them kind of the same. Since the amount of plugin is big, having different version for each one of them sounds like a nightmare.

azoitl:
> > What I don't know is what version number we should give new plugins. In my just submitted draft PR #2241 I used 3.0 but I'm not sure what is there a good practice. Open for suggestions.
> 
> I'm not sure about a good practice for eclipse plugins. For me, if it doesn't hurt much, I'd keep them kind of the same. Since the amount of plugin is big, having different version for each one of them sounds like a nightmare.

I'm not a fan of that. Because we would loose the information which plugins where not touched in a long time or do only have maintenance work.

PR #2239, opened 2026-03-09
Remove instantiation of runtimes outside of the functionality

cochicde:
This PR started as runtimes were slipping away from my interpreter, so I moved the creation of runtimes out of the functionality into three functions:

`FBRuntimeAbstract getOrCreateRuntime(final FBNetworkRuntime fBNetworkRuntime, final FBNetworkElement element)`
`FBNetworkRuntime getOrCreateNetworkRuntime(final FBNetworkRuntime fBNetworkRuntime, final UntypedSubApp uSubApp)`
`FBNetworkRuntime getOrCreateOuterNetworkRuntime(final FBNetworkRuntime fBNetworkRuntime, final UntypedSubApp uSubApp) `

I let a comment on the only line that's actually changing the behaviour, where a copy of the runtime is assigned instead of the actual runtime.

I also left comments on places where I think runtimes will be moving out from network runtimes. The example I have is an application which has a SubAppliacation and inside the SubApplication there's a composite FB. So I have:
- Network runtime -> App
- Network Runtime -> SubApp
- Network Runtime -> CFB
- Other runtimes

Image this scenario when reading the TODO comments

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
+++ b/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
@@ -42,23 +42,22 @@ public class FBNetworkDefaultInterpreter extends FBWithNetworkDefaultInterpreter
 			return switchNetwork(eventOccurrence.getEvent(), EcoreUtil.copy(fBNetworkRuntime));
 		}
 		if (eventOccurrence.getParentFB() instanceof final UntypedSubApp uSubApp) {
-			return new UntypedSubApplicationDefaultInterpreter(eventOccurrence).run(fBNetworkRuntime, uSubApp);
+			return new UntypedSubApplicationDefaultInterpreter(eventOccurrence, uSubApp).run(fBNetworkRuntime);
 		}
 
-		// run FB Type to get the output events for the instance in the network
-		FBRuntimeAbstract runtime = fBNetworkRuntime.getTypeRuntimes().get(eventOccurrence.getParentFB());
-		if (runtime == null) {
-			final FBType copiedType = EcoreUtil.copy(eventOccurrence.getParentFB().getType());
-			runtime = RuntimeFactory.createFrom(copiedType);
-			fBNetworkRuntime.getTypeRuntimes().put(eventOccurrence.getParentFB(), runtime);
-		}
+		final FBRuntimeAbstract runtime = RuntimeFactory.getOrCreateRuntime(fBNetworkRuntime,
+				eventOccurrence.getParentFB());
 
 		// sampling input & writing output is special for composite types
 		if (runtime instanceof final CompositeFBTypeRuntime compTypeRT) {
 			if (compTypeRT.getNetworkRuntime().getOuterNetworkRuntime() == null) {
+				// TODO: won't this move the fBNetworkRuntime from a possible
+				// composite/network parent runtime of it? or do we need to create a copy of it?
 				compTypeRT.getNetworkRuntime().setOuterNetworkRuntime(fBNetworkRuntime);
 				// put the composite runtime into the inner network, so we will find our way
 				// back to the outer network
+				// TODO: At this point compTypeRT is inside fBNetworkRuntime.getTypeRuntimes(),
+				// won't this move the compTypeRT from it?
 				compTypeRT.getNetworkRuntime().getTypeRuntimes().put(eventOccurrence.getParentFB(), compTypeRT);
 			}
 			return DefaultRunFBType.runFBType(runtime, eventOccurrence);
@@ -215,7 +214,7 @@ public class FBNetworkDefaultInterpreter extends FBWithNetworkDefaultInterpreter
 		if (!(dest instanceof Event)) {
 			throw new IllegalArgumentException("cannot trigger FB with pin " + dest.getName()); //$NON-NLS-1$
 		}
-		final EventOccurrence destEO = EventOccFactory.createFrom((Event) dest, null);
+		final EventOccurrence destEO = EventOccFactory.createFrom((Event) dest);
 
 		// if the destination EO does not have a parent, it might be the outgoing
 		// connection for the network inside a composite
--- a/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
+++ b/plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
@@ -51,13 +51,9 @@ public class FBNetworkDefaultInterpreter extends FBWithNetworkDefaultInterpreter
 		// sampling input & writing output is special for composite types
 		if (runtime instanceof final CompositeFBTypeRuntime compTypeRT) {
 			if (compTypeRT.getNetworkRuntime().getOuterNetworkRuntime() == null) {
-				// TODO: won't this move the fBNetworkRuntime from a possible
-				// composite/network parent runtime of it? or do we need to create a copy of it?
 				compTypeRT.getNetworkRuntime().setOuterNetworkRuntime(fBNetworkRuntime);
 				// put the composite runtime into the inner network, so we will find our way
 				// back to the outer network
-				// TODO: At this point compTypeRT is inside fBNetworkRuntime.getTypeRuntimes(),
-				// won't this move the compTypeRT from it?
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBWithNetworkDefaultInterpreter.java
```diff
@@ -64,14 +64,15 @@ protected static Event getEquivalentEventTypePin(final EventOccurrence sourceEve
...
+			final FBNetworkElement parentFB, final IInterfaceElement pin) {
 		final var equivalentFb = runtime.getFbnetwork().getFBNamed(parentFB.getName());
 		return InterfacePinUtils.findPinInInterface(equivalentFb, pin);
 	}
 
 	protected EList<EventOccurrence> switchNetwork(final Event event, final FBNetworkRuntime runtime) {
 		final EList<Connection> outputs = ConnectionUtils.getOutputConnections(event);
-		final EventOccurrence outputEO = EventOccFactory.createFrom(eventOccurrence.getEvent(), runtime);
+		final EventOccurrence outputEO = EventOccFactory.createFrom(eventOccurrence.getEvent(),
+				EcoreUtil.copy(runtime));
```
cochicde: this is the only line that's actually changing some behaviour. The rest is refactoring

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/api/RuntimeFactory.java
```diff
@@ -124,6 +124,47 @@ public static FBNetworkRuntime createRecursiveFrom(final FBNetwork network) {
...
+
+		FBNetworkRuntime runtime = (FBNetworkRuntime) fBNetworkRuntime.getTypeRuntimes().get(uSubApp);
+		if (runtime != null) {
+			return runtime;
+		}
+
+		runtime = RuntimeFactory.createFrom(uSubApp.getSubAppNetwork());
+		// TODO: won't this move the fBNetworkRuntime from a possible
+		// composite/network parent runtime of it? or do we need to create a copy of it?
+		runtime.setOuterNetworkRuntime(fBNetworkRuntime);
```
bwiesmayr, line 150: Maybe this was supposed to happen only when the runtime is created, not every time?

bwiesmayr: we probably also didn't want to "put" the runtime if it already existed

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
```diff
@@ -42,23 +42,22 @@ public EList<EventOccurrence> run(final FBNetworkRuntime fBNetworkRuntime) {
...
-			fBNetworkRuntime.getTypeRuntimes().put(eventOccurrence.getParentFB(), runtime);
-		}
+		final FBRuntimeAbstract runtime = RuntimeFactory.getOrCreateRuntime(fBNetworkRuntime,
+				eventOccurrence.getParentFB());
 
 		// sampling input & writing output is special for composite types
 		if (runtime instanceof final CompositeFBTypeRuntime compTypeRT) {
 			if (compTypeRT.getNetworkRuntime().getOuterNetworkRuntime() == null) {
+				// TODO: won't this move the fBNetworkRuntime from a possible
+				// composite/network parent runtime of it? or do we need to create a copy of it?
```
bwiesmayr: that's why we ask here whether it has an outer network runtime already? and if not, we try to fix it?

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
```diff
@@ -42,23 +42,22 @@ public EList<EventOccurrence> run(final FBNetworkRuntime fBNetworkRuntime) {
...
 				compTypeRT.getNetworkRuntime().setOuterNetworkRuntime(fBNetworkRuntime);
 				// put the composite runtime into the inner network, so we will find our way
 				// back to the outer network
+				// TODO: At this point compTypeRT is inside fBNetworkRuntime.getTypeRuntimes(),
+				// won't this move the compTypeRT from it?
```
bwiesmayr: I think the map has no containment and adding it into a map should not cause any issues here

bwiesmayr (commented):
here some of my comments. I think @Felix-Schmid knows better than me 

I have tested quite extensively with composite FBs (which were fine) and figured that there was some issue(s) with untyped subapps, especially in the combination composite FBs and subapps.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/api/RuntimeFactory.java
cochicde: I checked again and I think this is actually ok, as the `outerNetworkRuntime` member has `containment=false`, so I assume that if `fBNetworkRuntime` belonged somewhere else it will remain there

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
cochicde: Same here, I think is fine,  as `outerNetworkRuntime` member has `containment=false`,  I assume that if `fBNetworkRuntime` belonged somewhere else it will remain there

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
cochicde: Actually I think here's where I found the issue. I see in the model that `typeRuntimes` has `containment=true`. In my example, my SubApp runtime is contained in the `typeRuntimes` of the main application Runtime, and the CFB is contained in the `typeRuntimes` of the SubApp.

When executing the CFB inside this SubApp, the code at this point will assign itself ot its `typeRuntime` leaving my SubApp `typeRuntimes` without it. 

Maybe I missed something.

cochicde:
I removed the TODO comments in the code as they were meant to start a discussion, but not to focus on that, as this PR is mainly refactoring for clearer separation of responsabilities of the affected classes

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBWithNetworkDefaultInterpreter.java
Felix-Schmid: Unfortunately, this change breaks specifically for composite types. When trying to save the execution trace, an exception is thrown, because an EventImpl is not contained within a resource.
Steps to reproduce:
1) add a composite type (e.g., E_CYCLE) to a network
2) select an input event pin (e.g., START)
3) try to record the execution trace

Felix-Schmid (changes requested):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBWithNetworkDefaultInterpreter.java
cochicde: thanks for checking and for the example to reproduce it. I can reproduce but after some hours debugging I could not find a proper solution.
I remove the change and left it as it was, since this PR was intended as a refactoring, so that change was not even supposed to be there.

I'll check again in the future the problem, or even if I still need the change.

Felix-Schmid (approved):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBWithNetworkDefaultInterpreter.java
Felix-Schmid: Ok, sounds good. I also could not think of a quick solution for now...

PR #2240, opened 2026-03-09
Set value of input/outputs only once. Otherwise just copy the raw value

cochicde:


plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
```diff
@@ -145,8 +149,13 @@ private static void writeDataOutputsToConnections(final EventOccurrence eo, fina
 
 		final EList<VarDeclaration> networkVarsSample = getAssociatedDataPins(eo, runtime);
 
-		networkVarsSample.forEach(variable -> variable.getOutputConnections().stream().forEach(
-				outputConnection -> map.put(outputConnection, EcoreUtil.copy(getOutputValue(variable, runtime)))));
+		networkVarsSample.forEach(variable -> variable.getOutputConnections().stream().forEach(outputConnection -> {
+			if (map.get(outputConnection) == null) {
```
azoitl: by using computeIfAbsent I think this could be solved cleaner and better readable. In the lambda I would just create a plain Value instance and always use setValue. Would make the code better readable.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
```diff
@@ -97,7 +97,11 @@ private void sampleDataInput(final FBRuntimeAbstract runtime, final FBNetworkRun
 			}
 			final VarDeclaration typeVarDec = getEquivalentDataPinFromType(runtime, varDec);
 			if (!VariableUtils.isEmptyValue(value)) {
-				typeVarDec.setValue(EcoreUtil.copy(value));
+				if (typeVarDec.getValue() == null) {
+					typeVarDec.setValue(EcoreUtil.copy(value));
```
azoitl: If you instead of EcureUtil.copy  create an empty Value object you can remove the else and have in both case value.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
cochicde: I couldn't use `computeIfAbsent` as map here is of type EMap, which doesn't offer that method. I couldn't find anything similart. I changed to the other approach though, to set an empty value and remove the else.

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
cochicde: done

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
azoitl: ah I only saw map. Still I like this more. Especially when you have so simple structures doing it manually it is much more performant then `ECoreUtill.copy()`,

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
```diff
@@ -145,8 +148,12 @@ private static void writeDataOutputsToConnections(final EventOccurrence eo, fina
 
 		final EList<VarDeclaration> networkVarsSample = getAssociatedDataPins(eo, runtime);
 
-		networkVarsSample.forEach(variable -> variable.getOutputConnections().stream().forEach(
-				outputConnection -> map.put(outputConnection, EcoreUtil.copy(getOutputValue(variable, runtime)))));
+		networkVarsSample.forEach(variable -> variable.getOutputConnections().stream().forEach(outputConnection -> {
+			if (map.get(outputConnection) == null) {
+				map.put(outputConnection, LibraryElementFactory.eINSTANCE.createValue());
```
azoitl: put returns the value inserted. So if you change to:
```
var value = map.get(outputConnection)
if  value == null) {
    value = map.put(outputConnection, LibraryElementFactory.eINSTANCE.createValue());
}
value.setValue(getOutputValue(variable, runtime).getValue());
```
this saves you twice the get.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.fb.interpreter/src/org/eclipse/fordiac/ide/fb/interpreter/FBNetworkDefaultInterpreter.java
cochicde: you're right! Much better.

azoitl (approved):

PR #2241, opened 2026-03-09
Prototype AsciiDocEditor for Type Documentation

azoitl:
This needs currently the mylyn nightly update site as there is a bug fix for duplicated action handler registration.


azoitl:
Tested with Eclipse Mylyn 2026-06 M2. With all updates on both sides it finally builds and is ready for review.

oberlehner (approved):

PR #2242, opened 2026-03-09
Allow multiple timelines and forcing states at any point in time

cochicde:
Ok, so this is the big one. What this does:
- From a timeline of events, new timelines can spawn if the user manually triggers an event
- The whole state of the system is applied to the interpreter before triggering
- Values can be forced as in forte, which is also applied to the interpreter
- Current state of BFB are also watched (still not possible to force though)
- Everything is done using the GEF framework.

The main idea is as follow:
1. From the resource network, create all runtimes at the beginning and make sure all Value objects exist.
2. Iterate over the runtimes and retrieve all the Values from the data input/outputs, all input/output events, all connections Values, and from the Basic internal variabbles and current state.
3. Create different Map<String, ...> where the key is the watchpoint (App.FB.Interface) and the value is either a Value object, Event or BFBRuntime. 
4. Get/Set the state of the interpreter using the State representation from the ReplayNavigation (Map<String, String> where the key is the watchpotin and the value is the string representation of the value). With this set/get, any state in the replay debugging can be replicated on the interpreter and the execution continue from that point on.

This is how it looks:

Screenshot 2026-03-09 21 34 07


cochicde:
last push was just to bring the branch to the latest develop state

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/LaunchConfigurationTab.java
```diff
@@ -77,7 +77,7 @@ public class LaunchConfigurationTab extends AbstractLaunchConfigurationTab {
 	private static final String SELECT_PATH_DIALOG_TEXT = "Select trace directory"; //$NON-NLS-1$
```
azoitl, line 77: these texts here should not be marked with non nls, they should be moved to message.properties

azoitl (commented):
not sure if this is for another PR but user facing string must not be marked with non-nls and should not be extracted to constants but should follow the internationalization guidelines and es such be extracted into message.properties sot hat they can be translated.

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/LaunchConfigurationTab.java
cochicde: you're right. I thought in doing it in this PR, but I think it will be cleaner to do it separately. I still need to implement other things. I could open an issue an assign myself already so I don't forget about it

azoitl: Is this PR is rather big I agree for next time. For the future please only mark strings that are really not to be translated. We better go with warnings then with wrongly supressed ones.

azoitl (approved):

PR #2260, opened 2026-03-13
Fix duplicates in data type recursion check

mx990:
The data type recursion check did not remove the type when performing its depth-first search, which led to false positives if a struct was used multiple times as a member.


Felix-Schmid (approved):
Thanks for the fix!

azoitl (approved):

PR #2263, opened 2026-03-16
Reworked layout of interface editor

azoitl:
In my endevor of improving look and feel of 4diac IDE graphical editors I worked on the interface editor. With this change the connector lines are now independent editparts which will allow in the future to use them for with editing. Furhtermore the columns for comments and types are better aligned making the editor more calm. finally the colors for the with boxes and lines are toned done a bit.

The first commitin this chain is a bit larger couldn't get a smaller MVP.


mx990:
It seems the FB shape is missing in the *FB Debug* view:

image

mx990:
It looks like the *With* lines are a bit off-center and also go across the connectors:

image

Should there be a purple line going into `SOCKET1`?

plugins/org.eclipse.fordiac.ide.fbtypeeditor/css/dark_if_colors.css
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Alois Zoitl - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+IEclipsePreferences#org-eclipse-ui-workbench:org-eclipse-fordiac-fbtypeeditor {
+	preferences:
+		'org.eclipse.fordiac.ide.fbtypeeditor.WithLineColor=180,180,180'
+		'org.eclipse.fordiac.ide.fbtypeeditor.WithBoxColor=130,130,130'
```
mx990: I would suggest making the dark mode colors a bit darker:
```suggestion
		'org.eclipse.fordiac.ide.fbtypeeditor.WithLineColor=150,150,150'
		'org.eclipse.fordiac.ide.fbtypeeditor.WithBoxColor=100,100,100'
```

Image

mx990 (changes requested):
I really like the new layout together with the colored input and output lines and filled *With* connectors.

oberlehner (approved):
I like it !

azoitl:
> It looks like the _With_ lines are a bit off-center and also go across the connectors:
> image
> 

I also had the impression that we are of but thought more in the other direction. I measured and changed it to the following. Is that better?

image



> Should there be a purple line going into `SOCKET1`?

My idea was that only pins that can be used for with will have a connector line. Therefore I on purpose didn't show lines for plugs and sockets. I thought this is especially interesting for the future when we could allow to drag with connections from these connections.

plugins/org.eclipse.fordiac.ide.fbtypeeditor/css/dark_if_colors.css
azoitl: I have the same issue in dark mode. Will tune both colors more.

azoitl:
> It seems the FB shape is missing in the _FB Debug_ view:
> image

I'll have to investigate I fear this comes from some changes in the figure that I'm using now.

azoitl:
I had a problem in my with boxes this is now the real one:
image
as you can see in order to adjust to the line width (vertical with and horizontal connectors) it is not fully squared. Is this an issue?

mx990:
I do not belive it is much of an issue, but maybe 10x9 would look better than 8x9, since the horizontal line is thicker than the vertical one?

mx990:
> > Should there be a purple line going into `SOCKET1`?
> 
> My idea was that only pins that can be used for with will have a connector line. Therefore I on purpose didn't show lines for plugs and sockets. I thought this is especially interesting for the future when we could allow to drag with connections from these connections.

Could we not simply disallow dragging *Withs* to/from adapter lines, just as with the little pin decoration we have currently?

azoitl:
> I do not belive it is much of an issue, but maybe 10x9 would look better than 8x9, since the horizontal line is thicker than the vertical one?

I tried 9:10 but it looks a bit heavy:

Screenshot From 2026-03-19 00-01-30


and 7:8 to small:
image


Edit:

I noticed there is a bug in my 9:8 screenshot so here a fixed version:

image

azoitl:
> > > Should there be a purple line going into `SOCKET1`?
> > 
> > 
> > My idea was that only pins that can be used for with will have a connector line. Therefore I on purpose didn't show lines for plugs and sockets. I thought this is especially interesting for the future when we could allow to drag with connections from these connections.
> 
> Could we not simply disallow dragging _Withs_ to/from adapter lines, just as with the little pin decoration we have currently?

Yes we could, but to indirectly cite Jurassic Park: not sure if we should.

azoitl:
I further tuned the with colors in light and dark mode, the interface boxes now have rounded corners and are a bit more vertical, furthermore the with connections are drawn in an order bottom to top so that the vertical lines are always beneath the boxes and pin type changes are updating the connection color.

I'm currently in a lucky situation that I'm in an workshop with different 1499 users. Today I had the possibility to do with 5 experienced 1499 users (mostly with other tools) an interview about the new interface editor. I asked several questions to mask the main intent: finding out if adapters should have connections or not. I asked from different sides and very neutral. The feedback was that 4 out of this five did not expect that adapters would have connections. 1 person said it would be nice. Not sure if that is significant but I wanted to see what real users say. 

They definitely pointed out that the comments are hard to associate with pins. For this I have a prototype which I would turn into a PR when we are done with this one.

azoitl:
today I talked with more people. I'm still unsure but decided to add the connections also for adapters

mx990 (approved):

PR #2266, opened 2026-03-17
Add optional FB status to response for FB Lists

azoitl:


ruspl-afed:
I have a question: how do you track the `.ecore` changes @azoitl ?

Scenario to consider (this particular modification is just an example to illustrate the problem):
1) UserA creates model instance in IDE v3.2, specifies statuses carefully
2) UserB opens model instance in IDE v3.1 : maybe everything is fine (didn`t check TBH)
3) UserB does another valuable modification for model instance and saves it
=> all the statuses dropped from the model instance at this point

azoitl:
> I have a question: how do you track the `.ecore` changes @azoitl ?
> 
> Scenario to consider (this particular modification is just an example to illustrate the problem):
> 
>     1. UserA creates model instance in IDE v3.2, specifies statuses carefully
> 
>     2. UserB opens model instance in IDE v3.1 : maybe everything is fine (didn`t check TBH)
> 
>     3. UserB does another valuable modification for model instance and saves it
>        => all the statuses dropped from the model instance at this point

@ruspl-afed in order to overcome this problem we have for quite some time decoupled the persistence format from the EMF model. Our persistence model is based on the XML format defined in IEC 61499-2 and has not really changed in the last years. Most model changes we perform is to improve how we work on the user's data. We have some enhancements what we store in our model between 4diac IDE 2.0 and 3.0 but we are able to open an 2.0 project in 3.x versions. The other way is not always possible. Therefore we normally do not need to track and take care about most model changes. However what we should start and what we currently not have is to have a set of xml files and automated tests that will check if new versions can still open them and provide all required data.

mx990:
When I try this together with https://github.com/eclipse-4diac/4diac-forte/pull/822, I get the following exception:
```
!ENTRY org.eclipse.fordiac.ide.deployment.iec61499 2 0 2026-03-19 16:52:22.214
!MESSAGE Feature 'Status' not found. (, 3, 75)
<Response ID="0">
  <FBList>
    <FB name="EMB_RES" type="iec61499::system::EMB_RES" Status="RUNNING"/>

  </FBList>
</Response>
!STACK 0
org.eclipse.emf.ecore.resource.Resource$IOWrappedException: Feature 'Status' not found. (, 3, 75)
	at org.eclipse.emf.ecore.xmi.impl.XMLLoadImpl.handleErrors(XMLLoadImpl.java:77)
	at org.eclipse.emf.ecore.xmi.impl.XMLLoadImpl.load(XMLLoadImpl.java:268)
	at org.eclipse.emf.ecore.xmi.impl.XMLResourceImpl.doLoad(XMLResourceImpl.java:888)
	at org.eclipse.emf.ecore.xmi.impl.XMLResourceImpl.load(XMLResourceImpl.java:852)
	at org.eclipse.fordiac.ide.deployment.iec61499.executors.DeploymentExecutor.parseResponse(DeploymentExecutor.java:436)
	at org.eclipse.fordiac.ide.deployment.iec61499.executors.DeploymentExecutor.queryResources(DeploymentExecutor.java:371)
	at org.eclipse.fordiac.ide.deployment.interactors.AbstractDelegatingDeviceManagementInteractor.queryResources(AbstractDelegatingDeviceManagementInteractor.java:154)
	at org.eclipse.fordiac.ide.deployment.interactors.DeviceManagementExecutorService.lambda$53(DeviceManagementExecutorService.java:392)
[...]
Caused by: org.eclipse.emf.ecore.xmi.FeatureNotFoundException: Feature 'Status' not found. (, 3, 75)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.reportUnknownFeature(XMLHandler.java:2055)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.handleUnknownFeature(XMLHandler.java:2019)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.setAttribValue(XMLHandler.java:2771)
	at org.eclipse.emf.ecore.xmi.impl.SAXXMLHandler.handleObjectAttribs(SAXXMLHandler.java:77)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.createObjectFromFactory(XMLHandler.java:2257)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.handleFeature(XMLHandler.java:1947)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.processElement(XMLHandler.java:1058)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.startElement(XMLHandler.java:1036)
	at org.eclipse.emf.ecore.xmi.impl.XMLHandler.startElement(XMLHandler.java:725)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.startElement(AbstractSAXParser.java:518)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.AbstractXMLDocumentParser.emptyElement(AbstractXMLDocumentParser.java:183)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanStartElement(XMLDocumentFragmentScannerImpl.java:1387)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:2726)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:605)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanDocument(XMLDocumentFragmentScannerImpl.java:542)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:889)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:825)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XMLParser.parse(XMLParser.java:141)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.parse(AbstractSAXParser.java:1224)
	at java.xml/com.sun.org.apache.xerces.internal.jaxp.SAXParserImpl$JAXPSAXParser.parse(SAXParserImpl.java:637)
	at java.xml/com.sun.org.apache.xerces.internal.jaxp.SAXParserImpl.parse(SAXParserImpl.java:326)
	at org.eclipse.emf.ecore.xmi.impl.XMLLoadImpl.load(XMLLoadImpl.java:258)
	... 12 more
```

It seems the EMF attribute lookup is case-sensitive, because it works if I change the response to use lowercase `status` in `createFBResponseMessage` of https://github.com/eclipse-4diac/4diac-forte/pull/822.

ruspl-afed:
> It seems the EMF attribute lookup is case-sensitive

Yes, it normally is. 
From the above story, I concluded that the 4diac is completely independent of the EMF XMI, but there seem to be some details.

ruspl-afed:
> we have for quite some time decoupled the persistence format from the EMF model. Our persistence model is based on the XML format defined in IEC 61499-2 and has not really changed in the last years.

@azoitl what I can see in the [code](https://github.com/eclipse-4diac/4diac-ide/blob/23af93fcd0f17e0118af93286c55ee80c99b6e64/plugins/org.eclipse.fordiac.ide.deployment.iec61499/src/org/eclipse/fordiac/ide/deployment/iec61499/executors/DeploymentExecutor.java#L434-L435) is usage of classic EMF XML 
```
                final XMLResource xmlResource = new XMLResourceImpl();
                xmlResource.load(source, respMapping.getLoadOptions());
```
so my concern regarding ecore changes seems to be valid, but in a bit different scenario.

azoitl:
Interesting I did it the same as for name and type. Need to dig deeper what we did there.

ruspl-afed:
> Need to dig deeper what we did there.

Generally it looks like a requirement for exact match between `4diac-forte` and `4diac-ide` versions. Do we declare/desire such a limitation?

azoitl:
>>    Need to dig deeper what we did there.

> Generally it looks like a requirement for exact match between 4diac-forte and 4diac-ide versions. Do we declare/desire such a limitation?

In general yes. We are thinking about it but not yet have a good solution. However for this specific change it is not as the field is marked optional with a default value. I just messed up the EMF model configuration. I had the same issue the last time I had to fix something in this model.

mx990:
> Interesting I did it the same as for name and type. Need to dig deeper what we did there.

As you can see in the log message above, `name` and `type` are sent in lower case by FORTE, so it works for those.

plugins/org.eclipse.fordiac.ide.deployment/model/devResponse.ecore
```diff
@@ -8,12 +8,16 @@
...
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="status" eType="#//FBStatus"
+        defaultValueLiteral="UNKNOWN" unsettable="true"/>
   </eClassifiers>
   <eClassifiers xsi:type="ecore:EClass" name="FB">
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="name" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="type" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
     <eStructuralFeatures xsi:type="ecore:EReference" name="ports" upperBound="-1"
         eType="#//Port" containment="true"/>
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="status" eType="#//FBStatus"
+        defaultValueLiteral="UNKNOWN" unsettable="true"/>
```
mx990: We could also specify an `ExtendedMetaData` annotation with the XML attribute name:

```suggestion
    <eStructuralFeatures xsi:type="ecore:EAttribute" name="status" eType="#//FBStatus"
        defaultValueLiteral="UNKNOWN" unsettable="true">
      <eAnnotations source="http:///org/eclipse/emf/ecore/util/ExtendedMetaData">
        <details key="name" value="Status"/>
        <details key="kind" value="attribute"/>
      </eAnnotations>
    </eStructuralFeatures>
```

azoitl:
I just noticed that `name` and  `type` are wrongly sent from 4diac FORTE. They should also be upper case. I would fix that now. @mx990 do you know if with your annotation then both version `name` and `Name` will work? Or should I test?

azoitl:
Did some experiments and checked a bit. Unfortunately with annotations we will not get a backwards compatible solution if i change the wrong attributes in 4diac FORTE. What I found is that having a solution with both ways of writing is not that straight forward. @mx990 what would do you think we should do?

mx990:
We could enable the load option `XMLResource.OPTION_LAX_FEATURE_PROCESSING`. While this does not make the name lookup case-insensitive, it allows to fall back to the EMF attribute name even when specifying `ExtendedMetaData`.

In the mid to long term, we should probably consider writing our own parser, since we may also start running into performance issues when monitoring larger applications.

ruspl-afed:
> our own parser

We may start from providing our own `EStructuredFeature` resolution, this functionality is above parsing. 
However, I still do not understand what we are going to fight with, and have an impression that we are trying to hide a problem instead of formulating and solving it.

azoitl:
@mx990 thx for pointing to `XMLResource.OPTION_LAX_FEATURE_PROCESSING`. This is exactly what helps to have a quick solution for backwards compatibility at this point. But I agree that sooner or later we will need our own XML parser. Especially when we will switch per default to OPC UA deployment where we can have much larger messages then in the TCP based deployment.

@ruspl-afed the EMF model in discussion here was introduced to simplify handling the data that 4diac FORTE or IEC 61499 devices in general send back to an tool like 4diac IDE during deployment of an application. The XML protocol used is originally based on the following definitions:https://holobloc.com/doc/ita/index.htm (see section on Configurability). What I found out yesterday that we wrongly sent messages from 4diac FORTE using lower case attribute names.  This should definitely be fixed and improved. However as you point out just changing the EMF model for 3.2 will break any backwards compatibility (e.g., people using older 4diac FORTE instances). Therefore we needed a solution that works for both cases. While `XMLResource.OPTION_LAX_FEATURE_PROCESSING`  is a bit of a hack it is a very low key and quick solution solution for our problem.

ruspl-afed:
> What I found out yesterday that we wrongly sent messages from 4diac FORTE using lower case attribute names. This should definitely be fixed and improved. 

Agree, since this is a defect in our implementation of the protocol.

> However as you point out just changing the EMF model for 3.2 will break any backwards compatibility (e.g., people using older 4diac FORTE instances). Therefore we needed a solution that works for both cases.

Well, we have a choice how to handle this. What I would start from is "fixing schema versions".
1) 4diac FORTE should provide its signature in a message, so we can understand, that this is "4diac FORTE" of version "3.2.0" is talking with us. "No signature" will mean "4diac FORTE 3.1.0" or earlier.
2) EMF NS URI should be versioned, since NS URI is a key to distinguish one ecore (version) from another. All the readers of the XML should consult NS URI value first (and this is what EMF does during `Factory` instance resolution).
3) After having identifications for source and target we can judge the best approach to not fail during transformation.

mx990:
I agree that this solution would offer us the greatest flexibility for future changes while maintaining backwards compatibility. However, I also believe that would be overkill for this particular use case. We have had 8 commits that touched `devResponse.ecore` in the last 8 years since it was created. There were only 2 incompatible changes and both of them could have easily been avoided.

Conversely, our main concern here is performance, since the monitoring should be capable of handling tens or hundreds of messages every second, depending on the polling interval specified by the user. When communicating via OPC UA, there is also no hard limit on the amount of data received in each message.

Furthermore, the only purpose for this particular model is to parse ephemeral XML responses, so even using EMF is essentially overkill here. That is why we are considering moving away from EMF entirely and using a custom parser together with Java records in the future.

ruspl-afed:
> We have had 8 commits that touched `devResponse.ecore` in the last 8 years since it was created.

Thank you for your answer @mx990  Of cause you have more context to select better solution.

PR #2267, opened 2026-03-18
Remove `javax.activation` from target platform

ruspl-afed:
We do not use this (deprecated) bundle explicitly so it is better to remove it from the list of 3rd party.


azoitl (commented):
In previous releases we had issues not having it there. But I don't remember who added it and why.

mx990 (approved):
I had added it for Eclipse Milo, but it looks like they since switched to `jakarta.activation`.

@ruspl-afed Thanks for identifying and getting rid of this legacy dependency.

azoitl (approved):

PR #2270, opened 2026-03-18
Detect output events in simple blocks which are not used in ECStates

oberlehner:


ruspl-afed:
I'm not familiar with all the historical context of 4diac-ide code base, but they usually try to keep EMF-generated code as clean from everything else as possible. The reason to do so is to simplify future support effort, since 
1) embedding Java statements to `.ecore` make us kind of unarmed in the face of refactoring. 
2) having non-generated code in `src-gen` breaks the contract of `-gen` suffix (we have regular `src` for this)
3) `EOperation` is something that one need to introduce carefully, may be with a help of corresponding registry.

azoitl:
> I'm not familiar with all the historical context of 4diac-ide code base, but they usually try to keep EMF-generated code as clean from everything else as possible. The reason to do so is to simplify future support effort, since
> 
>     1. embedding Java statements to `.ecore` make us kind of unarmed in the face of refactoring.
> 
>     2. having non-generated code in `src-gen` breaks the contract of `-gen` suffix (we have regular `src` for this)
> 
>     3. `EOperation` is something that one need to introduce carefully, may be with a help of corresponding registry.

In general I would agree with you. Therefore we also avoid the use of generated not. Not fully there yet. However we noticed that by carefully introducing EOperations with dedicated helper classes in the src package we can make working with our model much easier.

ruspl-afed:
> Therefore we also avoid the use of generated not

me too!

I prefer to introduce `EOperation` via registry if really needed and to modify EMF code generation templates.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
```diff
@@ -75,4 +81,43 @@ static boolean validateInternalFBs(final BaseFBType fbType, final DiagnosticChai
 		}
 		return true;
 	}
+
+	static boolean validateEventUsage(final BaseFBType fbType, final DiagnosticChain diagnostics,
+			final Map<Object, Object> context) {
+		if (fbType instanceof final SimpleFBType simpleFBType) {
```
mx990: Please separate this into `SimpleFBTypeAnnotations` and move the model invariant to `SimpleFBType`, since it specifically targets simple FB types.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
```diff
@@ -75,4 +81,43 @@ static boolean validateInternalFBs(final BaseFBType fbType, final DiagnosticChai
...
+			final DiagnosticChain diagnostics, final String messagePattern) {
+		final Set<Event> unusedEventOutputs = new LinkedHashSet<>(fbType.getInterfaceList().getEventOutputs());
+		unusedEventOutputs.removeAll(usedEventOutputs);
+		return addEventWarnings(unusedEventOutputs, diagnostics, messagePattern);
+	}
+
+	private static boolean addEventWarnings(final Set<Event> events, final DiagnosticChain diagnostics,
+			final String messagePattern) {
+		if (diagnostics != null) {
+			events.stream().filter(Objects::nonNull)
```
mx990: I believe the filter is unnecessary here?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
```diff
@@ -75,4 +81,43 @@ static boolean validateInternalFBs(final BaseFBType fbType, final DiagnosticChai
...
+					.forEach(event -> diagnostics.add(new BasicDiagnostic(Diagnostic.WARNING,
+							LibraryElementValidator.DIAGNOSTIC_SOURCE,
+							LibraryElementValidator.BASE_FB_TYPE__VALIDATE_EVENT_USAGE,
+							MessageFormat.format(messagePattern, event.getName()), FordiacMarkerHelper
+									.getDiagnosticData(event, LibraryElementPackage.Literals.INAMED_ELEMENT__NAME))));
+			return true;
+		}
+		return false;
```
mx990: This currently returns true or false depending on if the diagnostics are null, but it should return false only if there are unused events.

mx990 (changes requested):

oberlehner:
> I'm not familiar with all the historical context of 4diac-ide code base, but they usually try to keep EMF-generated code as clean from everything else as possible. The reason to do so is to simplify future support effort, since
> 
> 1. embedding Java statements to `.ecore` make us kind of unarmed in the face of refactoring.
> 2. having non-generated code in `src-gen` breaks the contract of `-gen` suffix (we have regular `src` for this)
> 3. `EOperation` is something that one need to introduce carefully, may be with a help of corresponding registry.

We usually rarely do that, but if we do it we use a pattern for that.
The pattern is that we always introduce a dedicated annotation class, which makes maintenance of the injected code easier.

For Validation tasks, we need to add an eOperation, because the validator needs an invariant otherwise, the builder is not able to perform validations.

We don't use `generated not` for validation operations.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
oberlehner: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
oberlehner: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/BaseFBTypeAnnotations.java
oberlehner: done

ruspl-afed:
> The pattern is that we always introduce a dedicated annotation class, which makes maintenance of the injected code easier.

Thank you for the explanation @oberlehner , now the change is more clear for me.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/SimpleFBTypeAnnotations.java
```diff
@@ -0,0 +1,66 @@
...
+		if (diagnostics != null) {
+			events.stream()
+					.forEach(event -> diagnostics.add(new BasicDiagnostic(Diagnostic.WARNING,
+							LibraryElementValidator.DIAGNOSTIC_SOURCE,
+							LibraryElementValidator.SIMPLE_FB_TYPE__VALIDATE_EVENT_USAGE,
+							MessageFormat.format(messagePattern, event.getName()), FordiacMarkerHelper
+									.getDiagnosticData(event, LibraryElementPackage.Literals.INAMED_ELEMENT__NAME))));
+			return !events.isEmpty();
+		}
+		return true;
```
mx990: Please return true when the validation succeeds, that is, there are no unused events and also independent of whether the diagnostics is null:
```suggestion
		}
		return events.isEmpty();
```

mx990 (changes requested):

mx990 (approved):

PR #2278, opened 2026-03-19
Launch configuration should use `org.eclipse.fordiac.ide.product`

ruspl-afed:
Otherwise all the `fordiacimage` URIs are not resolved

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.contractspec/.launch/Launch Runtime Eclipse.launch
+++ b/plugins/org.eclipse.fordiac.ide.contractspec/.launch/Launch Runtime Eclipse.launch
@@ -1,34 +1,38 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.pde.ui.RuntimeWorkbench">
-<booleanAttribute key="append.args" value="true"/>
-<booleanAttribute key="askclear" value="true"/>
-<booleanAttribute key="automaticAdd" value="true"/>
-<booleanAttribute key="automaticValidate" value="false"/>
-<stringAttribute key="bad_container_name" value="/org.eclipse.fordiac.ide.contractspec/.launch/"/>
-<stringAttribute key="bootstrap" value=""/>
-<stringAttribute key="checked" value="[NONE]"/>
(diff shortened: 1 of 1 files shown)


ruspl-afed:
BTW why do we need our custom URI schema for images?

There is no chance that the PDE will ever find out about this and stop issuing warnings about missing image resources.

azoitl:
> BTW why do we need our custom URI schema for images?

This uri schema is based on an old article that I read in a Eclipse Magazine long time ago. The great advantage of this approach is that we decouple image file names from icons. And second we have all icons together in one fragment bundle and there a mapping properties file that maps icon names to image file names. This allows us to easier exchange and update images. And it allows users who want to brand label or change icons just to provide their own fragment. 

Until the Eclipse Platform does not have an own approach for icon themes I still see this the better option. Especially if you see what is going on in the Eclipse UI best practices repository to track down all the duplicated images and how to exchange them for new icons.  

> 
> There is no chance that the PDE will ever find out about this and stop issuing warnings about missing image resources.

Yes and that is sad.

ruspl-afed:
> Until the Eclipse Platform does not have an own approach for icon themes I still see this the better option.

Understood. Yes, this is better option for today. May be I'll suggest a better place for registering URI handler to make it more robust for `-product` changes.

azoitl:
> > Until the Eclipse Platform does not have an own approach for icon themes I still see this the better option.
> 
> Understood. Yes, this is better option for today. May be I'll suggest a better place for registering URI handler to make it more robust for `-product` changes.

Yes the place is not the best. The original article targeted a pure e4 application and there was a dedicated startup service. This I could not get working for our mixed e3/e4 approach. Improvements are appreciated.

ruspl-afed:
oops. I introduced a lot of line ending changes in initial commit. fixed.

ruspl-afed:
@Felix-Schmid any concerns here?

Felix-Schmid (approved):
Sorry for the delay. Looks good to me.

ruspl-afed:
> Sorry for the delay

no problem

> Looks good to me.

Thank you for review @Felix-Schmid !

azoitl (approved):

Issue #2277, opened 2026-03-19
Improve the shape of included Eclipse Milo artifacts


ruspl-afed:
We have a dedicated feature derived from `.target` definition with required Eclipse Milo libraries. 

Although it was common practice to isolate 3rd party libs in this way for many years, current practice is a bit different.
The main goal is to enable ability to update for a certain 3rd party in case of the security need. To achieve this, a number of rules should be applied:

1) 3rd party library bundle should not be a part of any feature
2) 3rd party library bundle should be references in a form of `Import-Package` directive only
3) 3rd party library bundle manifests should be normalized (see Eclipse Orbit as a reference)

Let's Improve the shape of included Eclipse Milo artifacts as a first step to this direction.

azoitl:
@ruspl-afed great that you found this issue and know how to improve our dependency handling here. Please note that Eclipse Milo had by now its graduation release and maybe before doing this upgrading to a post 1.0 Eclipse Milo release would be the more efficient approach. However we sofar haven't found time to do it and Milo changed significantly so it is unfortunately not a straight forward migration.

ruspl-afed:
> upgrading to a post 1.0 Eclipse Milo

This is definitely something we need to consider. However, such an update implies a change in the target platform.
So, I wouldn't expect these to be two separate independent changes.

azoitl:
> > upgrading to a post 1.0 Eclipse Milo
> 
> This is definitely something we need to consider. However, such an update implies a change in the target platform. So, I wouldn't expect these to be two separate independent changes.

Yes but then maybe updating to 1.0 and fixing the target platform and dependencies may be more efficient then first fixing the dependencies and then updating to fix them again.

ruspl-afed:
Would you like to update to [the latest Eclipse Milo we have for today](https://github.com/eclipse-milo/milo/releases/tag/v1.1.1)?

m-meingast:
> Would you like to update to [the latest Eclipse Milo we have for today](https://github.com/eclipse-milo/milo/releases/tag/v1.1.1)?

@ruspl-afed When migrating to 1.0 we faced some issues as described in https://github.com/eclipse-milo/milo/issues/1688. However, a solution for Milo was found and it should be available soon so that we can finally upgrade.

m-meingast:
As soon as the new Milo version releases, we can merge #2024

ruspl-afed:
Thank you for your comments @m-meingast 
[...]

PR #2281, opened 2026-03-19
Reduce the surface of Milo dependencies before upgrade

ruspl-afed:
The initial part of work for #2277 : switch from bundle to concrete packages to easier locate types in 1.0+


plugins/org.eclipse.fordiac.ide.deployment.opcua/META-INF/MANIFEST.MF
```diff
@@ -26,3 +22,10 @@ Require-Bundle: org.apache.log4j,
  org.eclipse.fordiac.ide.deployment,
  org.eclipse.fordiac.ide.deployment.iec61499
 Export-Package: org.eclipse.fordiac.ide.deployment.opcua
+Import-Package: org.eclipse.milo.opcua.sdk.client;version="[1.1.0,2.0.0)",
+ org.eclipse.milo.opcua.stack.core;version="[1.1.0,2.0.0)",
+ org.eclipse.milo.opcua.stack.core.types;version="1.1.2",
```
azoitl: is there a reason why you pint the version  for the `org.eclipse.milo.opcua.stack.core.types` and not use a range as for the others?

ruspl-afed: Good catch! 
The reason was in the tooling: this very first import I added manually to configure QuickFix, and then I added others via QuickFix,

I need a scenario to test it on runtime, because it may need more packages than compiler.

azoitl:
@m-meingast as you have a test setup for the Milo use case and it is a bit elaborate to set-up could you test this PR?

m-meingast:
I've tried the deployment with the Blink Tutorial application. Looks good to me.

m-meingast (approved):

azoitl (commented):
@m-meingast thx for testing.

@ruspl-afed thx for the improvement.

PR #2284, opened 2026-03-19
More defensively handle selection changed events in DiagramEditor

azoitl:


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/DiagramEditorWithFlyoutPalette.java
```diff
@@ -501,8 +501,11 @@ protected void createActions() {
 
 	@Override
 	public void selectionChanged(final IWorkbenchPart part, final ISelection selection) {
-		super.selectionChanged(part, selection);
-		updateActions(getSelectionActions());
+		// If not the active editor, ignore selection changed.
+		if (getSite() != null && getSite().getPage() != null && this.equals(getSite().getPage().getActiveEditor())) {
```
oberlehner, line 505: Why don't you use org.eclipse.fordiac.ide.ui.editors.EditorUtils.getCurrentActiveEditor() ?

azoitl: As we are inside of an editor we can use the editor infrastructure. This code is based on the code that Eclipse GEF Classic uses inside of the GraphicalEditor. 
I do not even have a clue why getSite or getPage may return null but I've seen this in the log files. I expect that we are not correctly de-registering from the selection services. Maybe something we need to check in our breadcrumb editor. Nonetheless a check here to not do anything is not a bad idea.

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/DiagramEditorWithFlyoutPalette.java
mx990: It seems this check is too strict in some cases. When I am in the `ApplicationEditor`, the call to `getSite().getPage().getActiveEditor()` returns the outer `AutomationSystemEditor`. This means actions/selections inside the `ApplicationEditor` are no longer updated. I am thus unable, for example, to delete any FB instances anymore.

azoitl: Thx for this feedback. This is in all cases to strict. We need to do something that I already did for the AbstractTypeEditorPage and forgot. Will create a PR soon.

azoitl: See #2312.

PR #2289, opened 2026-03-20
Show resource status in monitoring

mx990:


plugins/org.eclipse.fordiac.ide.deployment.debug/src/org/eclipse/fordiac/ide/deployment/debug/DeploymentDebugDevice.java
```diff
@@ -99,14 +100,29 @@ protected void terminated() {
...
+	}
+
+	private void updateResource(final org.eclipse.fordiac.ide.deployment.devResponse.Resource devResource) {
+		final DeploymentDebugResource resource = resources.computeIfAbsent(devResource.getName(), this::createResource);
+		if (resource != null) {
+			resource.setStatus(devResource.getStatus());
+		}
+	}
+
+	private DeploymentDebugResource createResource(final String name) {
```
ruspl-afed, line 119: What do you think about `Optional<DeploymentDebugResource>` return type for this method? Personally I fight with `null` everywhere, but may be you have other preferences.

mx990: Normally, I would agree, but `java.util.Map.computeIfAbsent(String, Function<? super String, ? extends DeploymentDebugResource>)` in line 113 expects a `null` value so no mapping is recorded.

ruspl-afed: Yes, with `computeIfAbsent` is is not so convenient to use `Optional`.
And it seems that the "problem" starts a little bit earlier, when we return `null` from device.

Thank you for sharing your opinion!

azoitl (approved):

PR #2292, opened 2026-03-21
Split FordiacTypeResource into Individual Resources

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	@Override
+	public T getLibraryElement() {
+		final LibraryElement content = LibraryElementResource.super.getLibraryElement();
+		return (typeClass.isInstance(content)) ? typeClass.cast(content) : null;
+	}
+
+	private IFile getTypeFile() throws IOException {
+		IFile typeFile = null;
```
mx990:
```suggestion
		final IFile typeFile;
```

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
```diff
@@ -0,0 +1,133 @@
...
+			throw new IOWrappedException(e);
+		}
+	}
+
+	@Override
+	protected void doSave(final OutputStream outputStream, final Map<?, ?> options) throws IOException {
+		final T content = getLibraryElement();
+		if (content == null) {
+			throw new IOException(
+					MessageFormat.format(Messages.FordiacTypeResource_UnsupportedContent, contents.get(0).toString()));
```
mx990: This will trigger an `IndexOutOfBoundsException` if the resource has no contents.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/LibraryElementResource.java
```diff
@@ -0,0 +1,23 @@
...
+ ********************************************************************************/
+package org.eclipse.fordiac.ide.model.resource;
+
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+
+public interface LibraryElementResource extends Resource {
+
+	default LibraryElement getLibraryElement() {
+		return (getContents().get(0) instanceof final LibraryElement le) ? le : null;
```
mx990: This will trigger an `IndexOutOfBoundsException` if the resource has no contents. I would suggest:

```suggestion
		return (!getContents().isEmpty() && getContents().getFirst() instanceof final LibraryElement le) ? le : null;
```

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
```diff
@@ -10,23 +9,48 @@
...
+		case TypeLibraryTags.DATA_TYPE_FILE_ENDING -> new DataTypeResourceImpl(uri);
+		case TypeLibraryTags.DEVICE_TYPE_FILE_ENDING -> new DeviceTypeResourceImpl(uri);
+		case TypeLibraryTags.FC_TYPE_FILE_ENDING -> new FunctionTypeResourceImpl(uri);
+		case TypeLibraryTags.FB_TYPE_FILE_ENDING -> new FBTypeResourceImpl(uri);
+		case TypeLibraryTags.GLOBAL_CONST_FILE_ENDING -> new GlobalConstTypeResourceImpl(uri);
+		case TypeLibraryTags.RESOURCE_TYPE_FILE_ENDING -> new ResourceTypeResourceImpl(uri);
+		case TypeLibraryTags.SEGMENT_TYPE_FILE_ENDING -> new SegmentTypeResourceImpl(uri);
+		case TypeLibraryTags.SUBAPP_TYPE_FILE_ENDING -> new SubappTypeResourceImpl(uri);
+		case TypeLibraryTags.SYSTEM_TYPE_FILE_ENDING -> new SystemResourceImpl(uri);
+		default -> null;
```
mx990: I believe we should throw an exception (e.g., `UnsupportedOperationException`) here instead, since callers will typically not expect `createResource` to return null.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
ruspl-afed: I agree that we should throw from create method, rather than return `null`. But may be `CoreException` may fit here better than `UnsupportedOperationException` since operation is in fact supported, but cannot be performed in particular conditions.

Another option is to change signature to `public Optional<LibraryElementResource> findResource(final URI uri)`
``

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/LibraryElementResource.java
ruspl-afed: I would avoid introducing new nullable API 
"create" method. We should either throw in case of any issues or convert it to `Optional<T> findT` method

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
```diff
@@ -0,0 +1,133 @@
...
+		final LibraryElement content = LibraryElementResource.super.getLibraryElement();
+		return (typeClass.isInstance(content)) ? typeClass.cast(content) : null;
+	}
+
+	private IFile getTypeFile() throws IOException {
+		IFile typeFile = null;
+		if (uri.isPlatformResource()) {
+			typeFile = ResourcesPlugin.getWorkspace().getRoot().getFile(new Path(this.uri.toPlatformString(true)));
+		} else if (uri.isFile()) {
+			typeFile = ResourcesPlugin.getWorkspace().getRoot().getFile(new Path(uri.toFileString()));
```
ruspl-afed, line 119: This doesn't look as the right approach, because it mixes `path` and `location` that will not always work. Perhaps we have more places like this already introduced in the sources base, that we need to review (and may be rework). Should we create a separate issue for this topic @azoitl @mx990 ?

azoitl: This code I moved from the previous FordiacTypeResource. i didn't want to do to many changes as this is just the first step for making EMF Resources the base layer of our persistence infrastructure. Currently we solely rely on IFiles and our TypeLibrary. Making this here better is something I'm very open but would see this in a separate PR.

ruspl-afed: This definitely deserves a separate discussion. Perhaps, you can suggest a suitable place for it.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
oberlehner: i would suggest to throw an exception as previously mentioned. This case should never happen, and therefore we want to know it early

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
azoitl: I guess a new discussion in 4diac IDE discussion for how and where to use EMF resources and what we should consider in that process would help. From the experience so far this will most probably anyhow be a longer process as loading and storing our models is currently one of our most performance sensitive area where we will need to get even better then we are.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
azoitl: `CoreException` is not allowed as the `Factory`  interface does not declare an exceptions. It needs to be an unchecked one.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/LibraryElementResource.java
azoitl: For now I would like to keep it nullable.

azoitl: added isEmpty check.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
azoitl: fixed

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/impl/AbstractLibraryElementResource.java
azoitl: fixed

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
```diff
@@ -10,23 +9,53 @@
...
+		case TypeLibraryTags.DATA_TYPE_FILE_ENDING -> new DataTypeResourceImpl(uri);
+		case TypeLibraryTags.DEVICE_TYPE_FILE_ENDING -> new DeviceTypeResourceImpl(uri);
+		case TypeLibraryTags.FC_TYPE_FILE_ENDING -> new FunctionTypeResourceImpl(uri);
+		case TypeLibraryTags.FB_TYPE_FILE_ENDING -> new FBTypeResourceImpl(uri);
+		case TypeLibraryTags.GLOBAL_CONST_FILE_ENDING -> new GlobalConstTypeResourceImpl(uri);
+		case TypeLibraryTags.RESOURCE_TYPE_FILE_ENDING -> new ResourceTypeResourceImpl(uri);
+		case TypeLibraryTags.SEGMENT_TYPE_FILE_ENDING -> new SegmentTypeResourceImpl(uri);
+		case TypeLibraryTags.SUBAPP_TYPE_FILE_ENDING -> new SubappTypeResourceImpl(uri);
+		case TypeLibraryTags.SYSTEM_TYPE_FILE_ENDING -> new SystemResourceImpl(uri);
+		default -> throw new InvalidParameterException(
```
mx990: Did you mean `java.lang.IllegalArgumentException`?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/FordiacTypeResourceFactory.java
azoitl: yes this is what I meant. Fixed.

mx990 (approved):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/resource/LibraryElementResource.java
azoitl: @ruspl-afed I had a bit a hectic day on Monday, I think I should add more reasoning. The usage of this method will either later never return 0 or will be wrapped into something that will take care of this. The reason is that we have dedicated error types in our model. This will ensure that even if files can not be loaded or are missing that the models of our users are always openable.

PR #2294, opened 2026-03-21
Remove reload type from ChangeStructCommand

azoitl:
Since quite some time the TypeEntry is preserved therefore searching for a new one does not make sense and using TypeEntry.getType shall always deliver the latest type.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/move/MoveTypeRefactoringParticipant.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/move/MoveTypeRefactoringParticipant.java
@@ -33,14 +33,10 @@ import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.model.IdentifierVerifier;
-import org.eclipse.fordiac.ide.model.commands.change.ChangeStructCommand;
-import org.eclipse.fordiac.ide.model.data.StructuredType;
 import org.eclipse.fordiac.ide.model.helpers.PackageNameHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.BlockFBNetworkElement;
 import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableFB;
-import org.eclipse.fordiac.ide.model.libraryElement.StructManipulator;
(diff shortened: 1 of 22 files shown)


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/move/MoveTypeRefactoringParticipant.java
```diff
@@ -179,7 +179,7 @@ public UpdateInstanceModelEdit(final BlockFBNetworkElement instance, final TypeE
 		protected Command createCommand(final BlockFBNetworkElement element) {
 			if (element instanceof final StructManipulator demux
 					&& typeEntry.getType() instanceof final StructuredType structuredType) {
-				return new ChangeStructCommand(demux, structuredType, visibleChildrenString, true);
+				return new ChangeStructCommand(demux, structuredType, visibleChildrenString);
```
mx990: Can we then also avoid passing the visible children, since that seems to be already handled by `ChangeStructCommand(StructManipulator, DataType)`? That should make this identical to the super implementation and we could get rid of this class entirely?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/move/MoveTypeRefactoringParticipant.java
azoitl: Thx for noticing this. Yes we can. However during testing I found a bug in the ` LibraryElementDependencyUpdater`, which I fixed in #2304.  This was maybe even the root cause for the problem I investigated and for which I created this PR.

azoitl:
I cherry picked to 3.1.x branch to get it for a 3.1.1. release. However it looks like github does  not let me switch branches for PRs anymore. Will make a new one.

mx990:
It seems to be a bit hidden now: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

azoitl:
> It seems to be a bit hidden now: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

I knew this solution. However Github removed the edit button for me.

azoitl:
Ah found it again. The edit button for the PR title allows also to change the branch

PR #2304, opened 2026-03-22
LibraryElementDependencyUpdater used old name to search in typelib

azoitl:
@mx990 does the dependency updater even need to search the typelib again? As we maintain typeentries couldn't we just use type entries as I do in PR #2294?

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
+++ b/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
@@ -64,7 +64,7 @@ public class LibraryElementDependencyUpdater extends LibraryElementDependencyTra
 				// react to TYPE_ENTRY_FULL_TYPE_NAME_FEATURE for renamed dependencies
 				if (TypeEntry.TYPE_ENTRY_FULL_TYPE_NAME_FEATURE.equals(notification.getFeature())) {
 					Display.getDefault()
-							.asyncExec(() -> updateDependency(dependency, notification.getOldStringValue()));
+							.asyncExec(() -> updateDependency(dependency, notification.getNewStringValue()));
 				}
 			}
 		} else {


mx990:
This is intentional. We want to get an error type (entry) with the old name, because this is what users would get if they were to close and re-open the editor.

azoitl:
> This is intentional. We want to get an error type (entry) with the old name, because this is what users would get if they were to close and re-open the editor.

Ok this is then bad as it breaks package name and type name refactoring if an editor is open.

azoitl:
I spent now more time to better understand what is going on and I think I can trace it down to the following line in 

https://github.com/eclipse-4diac/4diac-ide/blob/599986b566463e700ec50162809c96eadb7ffe2e/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java#L67

When the editor is open the model changes are executed on the display thread using `syncExec`. Although the dependency updates are triggered before the model changes on the affected types the model changes are performed before and the file is correctly saved. However then the waiting dependency update comes along and updates the editor model from the valid new type to an error type. Errors are shown to the user and if the users continues and saves a wrong model is saved. Also further refactorings do not work as the model in the editor has an error type entry. 

A quick fix that worked in a first experiment would be to change the `async` to a `sync` but I found that to dangerous to even propose that. 

Should certain refactoring block dependency updates as they are handled by the refactoring?

mx990:
> A quick fix that worked in a first experiment would be to change the async to a sync but I found that to dangerous to even propose that.

As you might have already guessed, that would then break name updates outside of refactorings.

> Should certain refactoring block dependency updates as they are handled by the refactoring?

I believe this would also be problematic, because if a user de-selects certain updates during refactoring, we will still need to show the resulting errors in the editor as the references then remain untouched.

I have now created a PR #2330 that explicitly runs the event loop before performing the refactoring in an editor when already in the UI thread to process any pending dependency updates. I am also considering changing the processing of dependency updates further to make them both more efficient and robust.

azoitl:
thx for the feedback and the PR. As commented at #2330 it still is not perfect but a great step forward.

PR #2324, opened 2026-03-24
Fix block positioning in nested unfolded subapps

oberlehner:
This fixes wrong block placement when moving elements inside nested unfolded subapps. The issue was caused by follow-up resize logic comparing child bounds and container bounds in different coordinate systems, which could trigger an incorrect container reposition after an otherwise valid move.

The patch normalizes these bounds before resize checks and updates the related trim, convert-to-subapp, and subapp interface resize paths to use the same bounds-based resize helper for consistent behavior.


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ResizeGroupOrSubappCommand.java
```diff
@@ -183,20 +186,40 @@ private static AbstractChangeContainerBoundsCommand checkAndCreateResizeInterfac
...
+			return null;
+		}
+		currentBounds.x += newBounds.x - oldBounds.x;
+		currentBounds.y += newBounds.y - oldBounds.y;
+		currentBounds.width += newBounds.width - oldBounds.width;
+		currentBounds.height += newBounds.height - oldBounds.height;
+		return FBNetworkXYLayoutEditPolicy.createChangeBoundsCommand(container, currentBounds);
+	}
+
+	private static Rectangle getCurrentBounds(final FBNetworkElement container) {
```
azoitl, line 208: I'm not happy with this method. You could easily use the on screenfigure from the container's edit part.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ResizeGroupOrSubappCommand.java
```diff
@@ -154,15 +159,13 @@ private AbstractChangeContainerBoundsCommand checkAndCreateResizeCommand(final G
...
+			final Rectangle fbBounds = getAbsoluteFBBounds(containerEP, children);
+			final Rectangle containerBounds = getAbsoluteContainerBounds(containerEP);
 			if (fbBounds != null && !containerBounds.contains(fbBounds)) {
 				fbBounds.union(containerBounds);
 				final FBNetworkElement container = getContainer(containerEP);
 				if (container != null) {
-					return createChangeBoundsCommand(container,
-							((GraphicalEditPart) containerEP.getParent()).getFigure().getBounds(), containerBounds,
-							fbBounds);
+					return createChangeBoundsCommand(container, containerBounds, fbBounds);
```
azoitl, line 168: as you translated the fb and containerBounds to absolute I think it would be better to translate them back to the relative to the containers figure.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/commands/ResizeGroupOrSubappCommand.java
```diff
@@ -243,6 +266,22 @@ private Rectangle getFBBounds(final List<FBNetworkElement> children) {
...
+		final Rectangle fbBounds = getFBBounds(children);
+		if (fbBounds != null) {
+			containerEP.getFigure().translateToAbsolute(fbBounds);
+		}
+		return fbBounds;
+	}
+
+	private static Rectangle getAbsoluteContainerBounds(final GraphicalEditPart containerEP) {
+		final Rectangle containerBounds = ContainerContentLayoutPolicy.getContainerAreaBounds(containerEP);
+		if (containerEP.getParent() instanceof final GraphicalEditPart parentEP) {
```
azoitl, line 279: please replace with:
`containerEP.getFigure().translateToAbsolute(containerBounds);`

azoitl (changes requested):

PR #2330, opened 2026-03-25
Fix dependency update during move refactoring

mx990:
There was a problem when refactorings were executed entirely within the UI thread, such as when using drag&drop to move a resource. This delayed UI updates, particularly dependency updates for edited library elements, until after the refactoring had completed.

This fixes the problem by explicitly running the event loop before performing changes to ensure any pending dependency updates have finished before proceeding.


azoitl (approved):
This is an improvement for move type refactorings. 

However I still have the problem that because of the dependency update that struct demuxers loose there visible children. Will need to check how to preserve that. But that is a separate PR.

PR #2332, opened 2026-03-25
Adds F_UPPER_BOUND, F_LOWER_BOUND and F_LEN_ARRAY FB definitions

MartinMelikMerkumians:
This is the IDE commit to provide F_UPPER_BOUND, F_LOWER_BOUND and F_LEN_ARRAY functionality as FBs


azoitl (commented):
Should we provide a default value for the dimension so that users that want to work with 1 dimensional arrays do not need to set a value. I think this is the main use case and would make the usage of the block much easier.

MartinMelikMerkumians:
That's a very good idea and I absolutely agree, I will set it to 1 (after I checked with the standard that 1 is correct)

azoitl (approved):

franz-hoepfinger-4diac:
F_LEN_ARRAY needs a FORTE FB right ? 
AND: should it have a IN only not a IN_OUT, to be same Interface as F_UPPER_BOUND, F_LOWER_BOUND ? 
(make it dual-use ? )

The Package Name is wrong, so it must be touched anyway ... 

https://github.com/eclipse-4diac/4diac-ide/pull/2379

PR #2334, opened 2026-03-26
Copy type didn't consider unloadable types.

azoitl:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/impl/AbstractTypeEntryImpl.java
```diff
@@ -372,8 +372,10 @@ private LibraryElement loadType() {
 	@Override
 	public LibraryElement copyType() {
 		final LibraryElement copy = EcoreUtil.copy(getType());
-		encloseInResource(copy);
-		copy.setTypeEntry(this);
+		if (copy != null) {
```
mx990, line 375: `getType()` no longer returns `null` when the type cannot be loaded.

azoitl: Ah thx for letting me know. missed that the error type creation is now inside getType.

Issue #2047, opened 2026-01-23
IDE Generation Error


franz-hoepfinger-4diac:
it is:
`#include "forte/datatypes/ANY_DERIVED_dtp.h"`
`CIEC_ANY_DERIVED_VARIANT`

it should be: 
`#include "forte/datatypes/forte_any_derived.h"`

4DIAC FORTE Export Filter 3.0.0.202512231800!

PR #2340, opened 2026-03-28
Fix incorrect handling of ANY_DERIVED in FORTE export

AryanMishra1789:
## Description
ANY_DERIVED was incorrectly treated as a derived datatype, resulting in ANY_DERIVED_dtp.h and CIEC_ANY_DERIVED_VARIANT. This happened because it was handled by generic cases instead of as a built-in type.

## Fix
Added a type-safe check using GenericTypes.ANY_DERIVED to map it to forte_any_derived.h and CIEC_ANY_DERIVED. This avoids incorrect _dtp and _VARIANT generation while preserving behavior for other types.

#2047

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
+++ b/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
@@ -210,6 +210,8 @@ final class ForteNgExportUtil {
 			AdapterType: '''«type.generateTypeNamespace»::FORTE_«type.generateTypeNamePlain»'''
 			ArrayType:
 				generateArrayTypeName(type.subranges, type.baseType)
+			// match ANY_DERIVED exclusively
+			DataType case type === GenericTypes.ANY_DERIVED: '''CIEC_ANY_DERIVED'''
 			// match generic types (must be before other data types)
 			DataType case GenericTypes.isAnyType(type): '''CIEC_«type.generateTypeNamePlain»_VARIANT'''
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
```diff
@@ -210,6 +210,8 @@ final class ForteNgExportUtil {
 			AdapterType: '''«type.generateTypeNamespace»::FORTE_«type.generateTypeNamePlain»'''
 			ArrayType:
 				generateArrayTypeName(type.subranges, type.baseType)
+			// match ANY_DERIVED exclusively
+			DataType case type === GenericTypes.ANY_DERIVED: '''CIEC_ANY_DERIVED'''
```
mx990, line 214: The `CIEC_ANY_DERIVED_VARIANT` was actually correct. It was simply missing in 4diac FORTE. I created PR https://github.com/eclipse-4diac/4diac-forte/pull/825 to add it.

plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend
```diff
@@ -327,6 +329,8 @@ final class ForteNgExportUtil {
 					"forte_array_fixed"
 			AdapterType:
 				type.name + "_adp"
+			AnyDerivedType case type === GenericTypes.ANY_DERIVED:
+				"forte_any_derived"
```
mx990, line 333: In this case, `forte_any_derived_variant` would be correct, which can be achieved by reordering the switch cases.

mx990 (changes requested):

PR #686, opened 2024-10-31
Add new event type E_MOVE to typelibrary with detailed information.

franz-hoepfinger-4diac:

PR #2341, opened 2026-03-29
Add configurable DataType support for E_MOVE block

franz-hoepfinger-4diac:
E_MOVE now uses ConfigurableMoveFB like F_MOVE, enabling the
DataType dropdown in the Properties View. This allows users to
configure the data type for E_MOVE pins at runtime, matching
the behavior of F_MOVE.

F_MOVE is a "forward always" Block, 
E_MOVE is a "sample and hold" Block introduced with this PR: https://github.com/eclipse-4diac/4diac-ide/pull/686

- Add TYPENAME_EMOVE and PACKAGE_NAME_EMOVE constants to LibraryElementTags
- Extend BlockInstanceFactory to create ConfigurableMoveFB for E_MOVE
- Update E_MOVE.fbt type definitions with new VersionInfo entries

Contributed on behalf of HR Agrartechnik GmbH.
Signed-off-by: Franz Höpfinger <f.hoepfinger@hr-agrartechnik.de>

### DIFF
--- a/data/typelibrary/events-1.0.0/typelib/E_MOVE.fbt
+++ b/data/typelibrary/events-1.0.0/typelib/E_MOVE.fbt
@@ -6,6 +6,8 @@
 	</VersionInfo>
 	<VersionInfo Organization=" HR Agrartechnik GmbH" Version="1.1" Author="Franz Höpfinger" Date="2024-10-31" Remarks="Copy F_MOVE to E_MOVE and adjust Comments">
 	</VersionInfo>
+	<VersionInfo Organization="HR Agrartechnik GmbH" Version="1.2" Author="Franz Höpfinger" Date="2026-03-29" Remarks="Added configurable DataType for E_MOVE similar to F_MOVE">
+	</VersionInfo>
 	<InterfaceList>
 		<EventInputs>
 			<Event Name="REQ" Type="Event" Comment="Normal Execution Request">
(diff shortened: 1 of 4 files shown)


azoitl (changes requested):
We do not have this block in 4diac FORTE. Now that I see it again I don't even understand what it does. Therefore I would like to delete it as I do not see any value compared to the F_MOVE block.

franz-hoepfinger-4diac:
@azoitl 

- do we need E_MOVE ? 
- should we better name it ? 

ma Favourite was to name it somehow like E_D_FF, so E_D_FF_BYTE and so on .. as Generic FB would fit right ? 

or do we have the SAME/SIMILAR functionality in the rtevents lib ? 
i am trying to dig a bit deeper, but unfortunately there is less Documentation. 
https://github.com/eclipse-4diac/4diac-ide/discussions/2395#discussioncomment-16587579
but, if i interpret TMIN correct, this is already a Minimum Inter-Arrival Time protecting subsequent Systems from Event-Storms... 
however E_MOVE should do something similar (only fire a Event when Data is changed) it is similar in the Goal right ?

franz-hoepfinger-4diac:
ALTERNATIVE: https://github.com/eclipse-4diac/4diac-ide/pull/2398
my clear preference would be to close this and merge https://github.com/eclipse-4diac/4diac-ide/pull/2398

franz-hoepfinger-4diac:
> ALTERNATIVE: #2398 my clear preference would be to close this and merge #2398

@azoitl Ready for review. if you also preffer #2398 i would close this !

azoitl:
I also like the name in #2398 more and would go with that one. Therefore closing this here.

PR #2352, opened 2026-03-31
Added CommandStackListener to update PlantHierarchy on Comment changes

sebHollersbacher:


ruspl-afed:
I'm not sure if we're using the right solution here. BTW, what problem are we solving with this fix and why is it so special that it deserves a separate listener instance?

oberlehner:
I would use the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` since this is specially invented for all changes that affect the plant hierachy.

oberlehner:
> I'm not sure if we're using the right solution here. BTW, what problem are we solving with this fix and why is it so special that it deserves a separate listener instance?

This fixes the problem, if we change the comment of an untyped subapp, we also want to update the plant hierachy view, which is a feature of  4diac IDE !
You can test this feature by selecting a project -> right click -> show in plant hierchachy !

sebHollersbacher:
> I would use the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` since this is specially invented for all changes that affect the plant hierachy.

Using the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` would create a new dependency for the application plug-in. I was also thinking of using the `org.eclipse.fordiac.ide.model.commands.QualNameChangeListenerManager`, however, I don't like to use the Qualified Name Listener infrastructure for comments

oberlehner:
> > I would use the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` since this is specially invented for all changes that affect the plant hierachy.
> 
> Using the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` would create a new dependency for the application plug-in. I was also thinking of using the `org.eclipse.fordiac.ide.model.commands.QualNameChangeListenerManager`, however, I don't like to use the Qualified Name Listener infrastructure for comments


- Refreshing is not working after changing a comment, i have to manually refresh the node
-Adding a new comment is not displayed in the hierachy:
image

sebHollersbacher:
> > > I would use the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` since this is specially invented for all changes that affect the plant hierachy.
> > 
> > 
> > Using the `org.eclipse.fordiac.ide.hierarchymanager.ui.listeners.HierarchyManagerUpdateListener` would create a new dependency for the application plug-in. I was also thinking of using the `org.eclipse.fordiac.ide.model.commands.QualNameChangeListenerManager`, however, I don't like to use the Qualified Name Listener infrastructure for comments
> 
> * Refreshing is not working after changing a comment, i have to manually refresh the node
>   -Adding a new comment is not displayed in the hierachy:

The commit had to be rebases as it relied on #2351 
It should work now

ruspl-afed:
> This fixes the problem, if we change the comment of an untyped subapp, we also want to update the plant hierachy view, which is a feature of 4diac IDE !

thank you for the explanation @sebHollersbacher , I plan to have a look on this area more closely, since I still have an impression that we are re-inventing EMF `ViewerNotification` here.

oberlehner:
> > This fixes the problem, if we change the comment of an untyped subapp, we also want to update the plant hierachy view, which is a feature of 4diac IDE !
> 
> thank you for the explanation @sebHollersbacher , I plan to have a look on this area more closely, since I still have an impression that we are re-inventing EMF `ViewerNotification` here.

i am not sure if we can use `ViewerNotification` since we also need to update the .plant.hier XML file.
But i am open to a better solution for synchronsing the different models

oberlehner (approved):

PR #2362, opened 2026-04-02
Fix pasting files from outside of 4diac

Felix-Schmid:
E.g., copying files in the system file explorer and pasting them into 4diac.


azoitl (approved):
LGTM

PR #2366, opened 2026-04-03
Add rename during copy refactoring

Felix-Schmid:
Display a rename prompt when resources are pasted into the container of the original file.


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
```diff
@@ -58,4 +69,63 @@ private ExistsResolve queryUserOverwrite(final IResource file, final IContainer
 		default -> ExistsResolve.CANCEL_ALL;
 		};
 	}
+
+	private ExistsResolve queryUserOverwriteSelf(final IResource file, final IContainer destination) {
+		final StringBuilder msg = new StringBuilder(Messages.Copy_RenameDialog_Message);
+		msg.append('\'').append(file.getName()).append('\'');
```
mx990: Please use `MessageFormat.format(...)`. You can escape single quotes by doubling them, e.g., `Enter a new name for ''{0}''`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
```diff
@@ -58,4 +69,63 @@ private ExistsResolve queryUserOverwrite(final IResource file, final IContainer
...
+
+		final IInputValidator validator = string -> {
+			final IWorkspace workspace = file.getWorkspace();
+			if (file.getName().equals(string)) {
+				return Messages.Copy_RenameDialog_MustBeNewName;
+			}
+			final IStatus status = workspace.validateName(string, file.getType());
```
mx990, line 82: Please also check if the name is a valid identifier using `IdentifierVerifier.verifyIdentifier(String)`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
```diff
@@ -58,4 +69,63 @@ private ExistsResolve queryUserOverwrite(final IResource file, final IContainer
...
+		String fileNameWithoutExt = oldName;
+		if (lastIdx > 0) {
+			fileExt = oldName.substring(lastIdx);
+			fileNameWithoutExt = oldName.substring(0, lastIdx);
+		}
+
+		final Matcher m = ENDS_WITH_NUMBER.matcher(fileNameWithoutExt);
+		if (m.find()) {
+			try {
+				final BigDecimal newNumber = new BigDecimal(m.group()).add(BigDecimal.ONE);
```
mx990: When making multiple copies, this currently keeps suggesting the same name, which already exists after the first copy. I would instead suggest to just append incrementing numbers until a file with the computed name does not exist yet.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
Felix-Schmid: Should we really enforce the use of valid identifier names for all files a user might copy inside a project? Otherwise I could add this check to the `CopyTypeParticipant`.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
Felix-Schmid: Good catch, done.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/UserCopyRefactoringQueries.java
oberlehner: maybe we can validate only for files with a 4diac file ending. see org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags

azoitl: As the goal is to get that infrastructure upstreamed to Eclipse Platform we should find a generic solution where we do not have 4diac IDE specific things.

mx990: > Should we really enforce the use of valid identifier names for all files a user might copy inside a project? Otherwise I could add this check to the `CopyTypeParticipant`.

Ah, thanks for clarifying. In that case, I also believe we should add this check to the `CopyTypeParticipant`.

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.systemmanagement.ui/src/org/eclipse/fordiac/ide/systemmanagement/ui/systemexplorer/SystemExplorerPasteAction.java
```diff
@@ -193,6 +194,9 @@ private static void startCopyRefactoring(final IResource[] files, final IContain
 			final Change change = changeOp.getChange();
 			if (change != null) {
 				change.perform(pm);
+			} else if (changeOp.getConditionCheckingFailedSeverity() >= RefactoringStatus.ERROR) {
+				ErrorDialog.openError(shell, null, null, Status.error(
+						changeOp.getConditionCheckingStatus().getMessageMatchingSeverity(RefactoringStatus.ERROR)));
```
mx990: I would suggest to use `RefactoringUI.createLightWeightStatusDialog(...)` here.

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #2369, opened 2026-04-05
Implement ECStateConnectionAnchor for better transition routing and s…

vikash1703:
[ECC Transition Routing and Anchor Implementation.docx](https://github.com/user-attachments/files/26490076/ECC.Transition.Routing.and.Anchor.Implementation.docx)

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2026 Johannes Kepler University Linz
+ * Copyright (c) 2026 Vikash Kumar Sinha
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -8,17 +8,14 @@
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
- *   Vikash Kumar - initial implementation
+ *   Vikash Kumar Sinha - initial implementation
  *******************************************************************************/
 package org.eclipse.fordiac.ide.fbtypeeditor.ecc.figures;
 
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.List;
 
 import org.eclipse.draw2d.AbstractConnectionAnchor;
-import org.eclipse.draw2d.IFigure;
-import org.eclipse.draw2d.Label;
 import org.eclipse.draw2d.geometry.Point;
 import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.fordiac.ide.model.libraryElement.ECState;
@@ -33,92 +30,75 @@ public class ECStateConnectionAnchor extends AbstractConnectionAnchor {
 	private final ECState state;
 	private final ECTransition transition;
 
-	public ECStateConnectionAnchor(final IFigure owner, final ECState state, final ECTransition transition) {
+	public ECStateConnectionAnchor(final ECStateFigure owner, final ECState state, final ECTransition transition) {
 		super(owner);
 		this.state = state;
 		this.transition = transition;
 	}
 
 	@Override
+	public ECStateFigure getOwner() {
+		return (ECStateFigure) super.getOwner();
+	}
+
+	@Override
 	public Point getReferencePoint() {
+		if (transition != null && transition.getPosition() != null) {
+			return transition.getPosition().toScreenPoint();
+		}
 		final Rectangle bounds = getOwner().getBounds().getCopy();
 		getOwner().translateToAbsolute(bounds);
 		return bounds.getCenter();
 	}
 
-	@Override
-	public Point getLocation(final Point reference) {
-		final Rectangle totalBounds = getOwner().getBounds().getCopy();
-		getOwner().translateToAbsolute(totalBounds);
-
-		final Point center = totalBounds.getCenter();
-
-		final Point ref = (reference != null) ? reference : center;
-
-		final Edge edge = edgeForReference(center, ref);
-
-		if (edge == Edge.RIGHT) {
-			return applySpacing(totalBounds, new Point(totalBounds.x + totalBounds.width, totalBounds.getCenter().y),
-					false, edge);
-		}
-
-		if (getOwner() instanceof ECStateFigure) {
-			final Label nameLabel = ((ECStateFigure) getOwner()).getNameLabel();
-			final Rectangle nameBounds = nameLabel.getBounds().getCopy();
-
-			nameLabel.translateToAbsolute(nameBounds);
-
-			return switch (edge) {
-			case TOP -> applySpacing(nameBounds, nameBounds.getTop(), true, edge);
-			case LEFT -> applySpacing(nameBounds, nameBounds.getLeft(), false, edge);
-			case BOTTOM -> applySpacing(nameBounds, nameBounds.getBottom(), true, edge);
-			default -> applySpacing(totalBounds, new Point(totalBounds.x + totalBounds.width, center.y), false, edge);
-			};
-		}
-
-		return center;
-	}
-
 	private Edge edgeForReference(final Point center, final Point ref) {
 		final int dx = ref.x - center.x;
 		final int dy = ref.y - center.y;
-		final boolean hasAction = (state != null) && !state.getECAction().isEmpty();
 
-		if (dx >= 0 && dy > 0 && hasAction) {
-			return Edge.RIGHT;
-		}
 		if (Math.abs(dy) > Math.abs(dx)) {
 			return (dy >= 0) ? Edge.BOTTOM : Edge.TOP;
 		}
+		if (dx >= 0 && state != null && !state.getECAction().isEmpty()) {
+			return (dy >= 0) ? Edge.BOTTOM : Edge.TOP;
+		}
+
 		return (dx >= 0) ? Edge.RIGHT : Edge.LEFT;
 	}
 
-	private Point applySpacing(final Rectangle bounds, final Point base, final boolean verticalEdge, final Edge edge) {
+	@Override
+	public Point getLocation(final Point reference) {
+		final Rectangle nameBounds = getOwner().getNameLabel().getBounds().getCopy();
+		getOwner().getNameLabel().translateToAbsolute(nameBounds);
+
+		final Point ref = (reference != null) ? reference : nameBounds.getCenter();
+		final Edge edge = edgeForReference(nameBounds.getCenter(), ref);
+
+		return applySpacing(nameBounds, edge);
+	}
+
+	private Point applySpacing(final Rectangle bounds, final Edge edge) {
 		final List<ECTransition> ordered = getTransitionsOnEdgeOrdered(edge);
 		final int index = indexOf(ordered, transition);
 		final int count = Math.max(1, ordered.size());
 
-		final Point p = base.getCopy();
-
-		if (verticalEdge) {
+		return switch (edge) {
+		case TOP -> {
 			final int slotWidth = bounds.width / (count + 1);
-			p.x = bounds.x + (index + 1) * slotWidth;
-		} else {
+			yield new Point(bounds.x + (index + 1) * slotWidth, bounds.y);
+		}
+		case BOTTOM -> {
+			final int slotWidth = bounds.width / (count + 1);
+			yield new Point(bounds.x + (index + 1) * slotWidth, bounds.y + bounds.height);
+		}
+		case LEFT -> {
 			final int slotHeight = bounds.height / (count + 1);
-			p.y = bounds.y + (index + 1) * slotHeight;
+			yield new Point(bounds.x, bounds.y + (index + 1) * slotHeight);
 		}
-
-		if (edge == Edge.RIGHT) {
-			p.x = bounds.x + bounds.width;
-		} else if (edge == Edge.LEFT) {
-			p.x = bounds.x;
-		} else if (edge == Edge.BOTTOM) {
-			p.y = bounds.y + bounds.height;
-		} else if (edge == Edge.TOP) {
-			p.y = bounds.y;
+		case RIGHT -> {
+			final int slotHeight = bounds.height / (count + 1);
+			yield new Point(bounds.x + bounds.width, bounds.y + (index + 1) * slotHeight);
 		}
-
-		return p;
+		};
 	}
 
 	private List<ECTransition> getTransitionsOnEdgeOrdered(final Edge edge) {
@@ -131,13 +111,30 @@ public class ECStateConnectionAnchor extends AbstractConnectionAnchor {
 				? state.getOutTransitions()
 				: state.getInTransitions();
 
+		final Rectangle nameBounds = getOwner().getNameLabel().getBounds().getCopy();
+		getOwner().getNameLabel().translateToAbsolute(nameBounds);
+		final Point center = nameBounds.getCenter();
+
 		for (final ECTransition t : list) {
-			if (edgeForTransition(t) == edge) {
+
+			if (t.getPosition() == null) {
+				continue;
+			}
+
+			final Point bendPoint = t.getPosition().toScreenPoint();
+			if (edgeForReference(center, bendPoint) == edge) {
 				result.add(t);
 			}
 		}
+		result.sort((t1, t2) -> {
+			final Point p1 = t1.getPosition().toScreenPoint();
+			final Point p2 = t2.getPosition().toScreenPoint();
+			if (edge == Edge.TOP || edge == Edge.BOTTOM) {
+				return Integer.compare(p1.x, p2.x);
+			}
+			return Integer.compare(p1.y, p2.y);
+		});
 
-		result.sort(Comparator.comparingDouble(this::angleForTransitionNormalized));
 		return result;
 	}
 
@@ -151,55 +148,5 @@ public class ECStateConnectionAnchor extends AbstractConnectionAnchor {
 			}
 		}
 		return 0;
-	}
-
-	private Edge edgeForTransition(final ECTransition t) {
-		if (t == null || state == null) {
-			return Edge.RIGHT;
-		}
-		final ECState other = (t.getSource() == state) ? t.getDestination() : t.getSource();
-		if (other == null) {
-			return Edge.RIGHT;
-		}
-
-		final Rectangle bounds = getOwner().getBounds().getCopy();
-		getOwner().translateToAbsolute(bounds);
-		final Point center = bounds.getCenter();
-
-		final Point otherPos = other.getPosition().toScreenPoint();
-		final int halfW = bounds.width / 2;
-		final int halfH = bounds.height / 4; // ← 12 hata diya
-		final Point otherCenter = new Point(otherPos.x + halfW, otherPos.y + halfH);
-
-		final int dx = otherCenter.x - center.x;
-		final int dy = otherCenter.y - center.y;
-		final boolean hasAction = !state.getECAction().isEmpty();
-
-		if (dx >= 0 && dy > 0 && hasAction) {
-			return Edge.RIGHT;
-		}
-		if (Math.abs(dy) > Math.abs(dx)) {
-			return (dy >= 0) ? Edge.BOTTOM : Edge.TOP;
-		}
-		return (dx >= 0) ? Edge.RIGHT : Edge.LEFT;
-	}
-
-	private double angleForTransitionNormalized(final ECTransition t) {
-		if (t == null || state == null) {
-			return 0.0;
-		}
-		final ECState other = (t.getSource() == state) ? t.getDestination() : t.getSource();
-		if (other == null) {
-			return 0.0;
-		}
-
-		final double dx = other.getPosition().getX() - state.getPosition().getX();
-		final double dy = other.getPosition().getY() - state.getPosition().getY();
-
-		double angle = Math.atan2(dy, dx);
-		if (angle < 0) {
-			angle += Math.PI * 2.0;
-		}
-		return angle;
 	}
 }
--- a/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
+++ b/plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
@@ -1,0 +1,205 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Johannes Kepler University Linz
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Vikash Kumar - initial implementation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.fbtypeeditor.ecc.figures;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+import org.eclipse.draw2d.AbstractConnectionAnchor;
+import org.eclipse.draw2d.IFigure;
+import org.eclipse.draw2d.Label;
+import org.eclipse.draw2d.geometry.Point;
+import org.eclipse.draw2d.geometry.Rectangle;
+import org.eclipse.fordiac.ide.model.libraryElement.ECState;
+import org.eclipse.fordiac.ide.model.libraryElement.ECTransition;
+
+public class ECStateConnectionAnchor extends AbstractConnectionAnchor {
+
+	private enum Edge {
+		TOP, BOTTOM, LEFT, RIGHT
+	}
+
+	private final ECState state;
+	private final ECTransition transition;
+
+	public ECStateConnectionAnchor(final IFigure owner, final ECState state, final ECTransition transition) {
+		super(owner);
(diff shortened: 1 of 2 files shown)


vikash1703:
@azoitl 

Hi this is not for actual merge i wanted you to test the changes and is there is any conflicts then i try to resolve i have attached documentation for the same.

Thanks
vikas

azoitl:
@vikash1703 thx for letting us know. BTW it would be great if you could add the content of your docx file as comment of the PR. This would make it more easier for others to read and is also more secure.

vikash1703:
> @vikash1703 thx for letting us know. BTW it would be great if you could add the content of your docx file as comment of the PR. This would make it more easier for others to read and is also more secure.

1. 
Overview 
This implementation addresses the issue where multiple transitions in the ECC (Event Control Chart) would overlap when connected to the same state. By following the GEF MVC architecture, we have decoupled the routing logic from the figure and implemented a dynamic spacing system.

2. 
Key Components & Implementation Details 

ECStateConnectionAnchor.java (The Logic): This class acts as the "Traffic Manager" for transitions, calculating exact coordinates on the state border. 

Smart Edge Selection: Dynamically determines the best edge (TOP, BOTTOM, LEFT, RIGHT) based on the relative position of the target state. 

Priority Logic: Transitions prioritize the BOTTOM or RIGHT edges if a state contains Actions (Algorithms/Output Events) to avoid text overlap. 

Dynamic Spacing: Separates multiple transitions on the same edge with a 15-pixel gap (SAME_EDGE_GAP = 15). 

The Margin Guard: A 4-pixel margin acts as an "Invisible Safety Wall" at corners, preventing anchors from sliding onto sharp corners or falling off the edge. 

Angle-Based Sorting: Transitions are sorted by their normalized angle to ensure they remain in a consistent sequence and do not cross.

ECStateEditPart.java (The Controller): Manages the lifecycle of the State and its connections. 

Decoupling: Replaced the default ChopboxAnchor with the custom ECStateConnectionAnchor. 

Integration: Overrode getSourceConnectionAnchor and getTargetConnectionAnchor so the editor automatically uses the new spacing logic.

3. 
Core Benefits 

Anti-Overlap: Clear separation makes complex state machines readable. 

Visual Stability: The 15px gap provides "breathing room" for labels. 

Corner Safety: Prevents visual glitches at corner tips. 

Architectural Integrity: Keeps Figure (UI) classes clean by keeping logic in the Anchor classes.

4. 
Testing & Verification Procedure 
Test 1 (Crowding): Connected 5+ transitions to one edge. Result: Lines spread out in a fan shape with a 15px gap and no overlap. 

Test 2 (Corner/Margin): Moved state to extreme diagonal angles. Result: Anchor point locked 4 pixels before reaching the corner. 

Test 3 (Action Protection): Added an Action to a state. Result: Transitions automatically shifted toward the BOTTOM edge to preserve readability. 

Test 4 (Dynamic Re-routing): Dragged state in a circular motion. Result: Anchor smoothly jumped between edges while maintaining sorted order.

vikash1703:
> @vikash1703 thx for letting us know. BTW it would be great if you could add the content of your docx file as comment of the PR. This would make it more easier for others to read and is also more secure.

please test once if anything you cant figure it out i will put a video tomorrow.

vikash1703:
> @vikash1703 thx for letting us know. BTW it would be great if you could add the content of your docx file as comment of the PR. This would make it more easier for others to read and is also more secure.

@azoitl 

I want to confirm that we have the feature where the arrow change automatically when the state are moved as i am confused when i setup my 4diac ide again it still show me the same that i show you on zoom meet?

azoitl:
> > @vikash1703 thx for letting us know. BTW it would be great if you could add the content of your docx file as comment of the PR. This would make it more easier for others to read and is also more secure.
> 
> @azoitl
> 
> I want to confirm that we have the feature where the arrow change automatically when the state are moved as i am confused when i setup my 4diac ide again it still show me the same that i show you on zoom meet?

Per default GEF rotates the connection styles based on the line segments given to it. So if your connection getLines returns the start/end segment with the same trajectory as your spline you should automatically get the right orientation of the arrow.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,186 @@
+package org.eclipse.fordiac.ide.fbtypeeditor.ecc.figures;
```
oberlehner, line 13: please add a copyright header. Otherwise the build will fail

vikash1703: > please add a copyright header. Otherwise the build will fail



> please add a copyright header. Otherwise the build will fail

@oberlehner 

Thanks for guidence i will do the same.

vikash1703: @azoitl and @oberlehner 

I have made some changes and try to follow all rules of GFE classic.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,205 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Johannes Kepler University Linz
```
azoitl: Please note that you or your employer is the copyright holder for any contributions that you are doing. Please refere also the Eclipse project handbook on more details on that.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,205 @@
...
+		final Rectangle totalBounds = getOwner().getBounds().getCopy();
+		getOwner().translateToAbsolute(totalBounds);
+
+		final Point center = totalBounds.getCenter();
+
+		final Point ref = (reference != null) ? reference : center;
+
+		final Edge edge = edgeForReference(center, ref);
+
+		if (edge == Edge.RIGHT) {
```
azoitl: why is right here handled special?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,205 @@
...
+			return applySpacing(totalBounds, new Point(totalBounds.x + totalBounds.width, totalBounds.getCenter().y),
+					false, edge);
+		}
+
+		if (getOwner() instanceof ECStateFigure) {
+			final Label nameLabel = ((ECStateFigure) getOwner()).getNameLabel();
+			final Rectangle nameBounds = nameLabel.getBounds().getCopy();
```
azoitl: why do you need the nameLabel bounds? I think this breaks for the start state. Can you not just take the bounds of the host figure?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,205 @@
...
+
+			nameLabel.translateToAbsolute(nameBounds);
+
+			return switch (edge) {
+			case TOP -> applySpacing(nameBounds, nameBounds.getTop(), true, edge);
+			case LEFT -> applySpacing(nameBounds, nameBounds.getLeft(), false, edge);
+			case BOTTOM -> applySpacing(nameBounds, nameBounds.getBottom(), true, edge);
+			default -> applySpacing(totalBounds, new Point(totalBounds.x + totalBounds.width, center.y), false, edge);
```
azoitl: if you have a switch case on an enum I would recommend to not add a default. This here would be the right case. Therefore you shold also mention right here.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,205 @@
...
+
+	public ECStateConnectionAnchor(final IFigure owner, final ECState state, final ECTransition transition) {
+		super(owner);
+		this.state = state;
+		this.transition = transition;
+	}
+
+	@Override
+	public Point getReferencePoint() {
+		final Rectangle bounds = getOwner().getBounds().getCopy();
```
azoitl, line 49: Shouldn't you use the transition's middle point from the model as reference point?

azoitl (changes requested):
Your new connection anchor is quite a complex code. I tried to go through it and mark a few first things that I noted. I think more clean-up then what I found is needed.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: you are right about the start state. I have updated getConnectionBounds() in ECStateFigure to use the figure's own bounds. When there are no actions it returns the full bounds, and when actions are present it limits the width and height to the name label area. ECStateConnectionAnchor only calls getConnectionBounds() — it has no direct knowledge of nameLabel at all.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: I have removed the default

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: you are right. I have updated getReferencePoint() to return the transition's middle point from the model using transition.getPosition().toScreenPoint(). When transition is null, it falls back to the figure's center.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: The RIGHT edge was handled separately because when a state has actions, transitions on the right side should attach to the full figure's right edge including the action area, not just the name label bounds. But you are right — with getConnectionBounds() now handling this logic internally, the special case is no longer needed. I will remove it and let the switch handle all edges uniformly.

azoitl: If `edgeForReference()` will check for that and return Top or Bottom in this case the whole code would be cleaner and the switch case below would be always correct.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
azoitl: I'm sorry I was wrong. For calculating the start position you need the name label bounds.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateFigure.java
```diff
@@ -219,4 +222,14 @@ public Insets getInsets(final IFigure figure) {
 		}
 
 	}
+
+	public Rectangle getConnectionBounds() {
```
azoitl: I'm sorry that I created confusion with the name label. As stated above the name label bounds are the better solutions that work on all cases without introducing to much new code.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,207 @@
...
+public class ECStateConnectionAnchor extends AbstractConnectionAnchor {
+
+	private enum Edge {
+		TOP, BOTTOM, LEFT, RIGHT
+	}
+
+	private final ECState state;
+	private final ECTransition transition;
+
+	public ECStateConnectionAnchor(final IFigure owner, final ECState state, final ECTransition transition) {
```
azoitl: When you change the `IFigure` here to an `ECStateFigure` an instance of this connection anchor can only be created with an `ECStateFigure` with that information you can override `getOwner`and return an `ECStateFigure`. Then you can remove all `instanceof ECStateFigure` checks making your code much better readable.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateFigure.java
vikash1703: > I'm sorry that I created confusion with the name label. As stated above the name label bounds are the better solutions that work on all cases without introducing to much new code.



> I'm sorry that I created confusion with the name label. As stated above the name label bounds are the better solutions that work on all cases without introducing to much new code.



> I'm sorry that I created confusion with the name label. As stated above the name label bounds are the better solutions that work on all cases without introducing to much new code.

but it can throw nullpointererror in some cases i think if we add new method it is more scalable no problem now let me deeply analyse as i am understanding now what you want.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: > When you change the `IFigure` here to an `ECStateFigure` an instance of this connection anchor can only be created with an `ECStateFigure` with that information you can override `getOwner`and return an `ECStateFigure`. Then you can remove all `instanceof ECStateFigure` checks making your code much better readable.



> When you change the `IFigure` here to an `ECStateFigure` an instance of this connection anchor can only be created with an `ECStateFigure` with that information you can override `getOwner`and return an `ECStateFigure`. Then you can remove all `instanceof ECStateFigure` checks making your code much better readable.



> When you change the `IFigure` here to an `ECStateFigure` an instance of this connection anchor can only be created with an `ECStateFigure` with that information you can override `getOwner`and return an `ECStateFigure`. Then you can remove all `instanceof ECStateFigure` checks making your code much better readable.

Hmm let me also think i that what you are thinking but for now yes its better to make our code more readable and less lengthy as i also understand this implementation in more complex.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateFigure.java
azoitl: every state figure has a name label and the namelabel is the label a connection should attach to. Introducing a new method with a name that is somehow not related to states is IMHO not a good idea.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,189 @@
...
+		}
+		for (int i = 0; i < list.size(); i++) {
+			if (list.get(i) == t) {
+				return i;
+			}
+		}
+		return 0;
+	}
+
+	private Edge edgeForTransition(final ECTransition t) {
```
azoitl: why do you need a second method for defining which edge to use?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,189 @@
...
+
+		if (verticalEdge) {
+			final int slotWidth = bounds.width / (count + 1);
+			p.x = bounds.x + (index + 1) * slotWidth;
+		} else {
+			final int slotHeight = bounds.height / (count + 1);
+			p.y = bounds.y + (index + 1) * slotHeight;
+		}
+
+		if (edge == Edge.RIGHT) {
```
azoitl: why is this if else cascade not an switch

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,189 @@
...
+			return transition.getPosition().toScreenPoint();
+		}
+		final Rectangle bounds = getOwner().getBounds().getCopy();
+		getOwner().translateToAbsolute(bounds);
+		return bounds.getCenter();
+	}
+
+	@Override
+	public Point getLocation(final Point reference) {
+		final Rectangle totalBounds = getOwner().getBounds().getCopy();
```
azoitl: as you ownly use the center you con also use the name label or the connection bounds, reducing the complexity of the code.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,189 @@
...
+		final int dx = ref.x - center.x;
+		final int dy = ref.y - center.y;
+
+		if (Math.abs(dy) > Math.abs(dx)) {
+			return (dy >= 0) ? Edge.BOTTOM : Edge.TOP;
+		}
+		return (dx >= 0) ? Edge.RIGHT : Edge.LEFT;
+	}
+
+	private Point applySpacing(final Rectangle bounds, final Point base, final boolean verticalEdge, final Edge edge) {
```
azoitl: it is unclear to me why you need the bounds and the base, as well as the verticalEdge and the edge.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
vikash1703: > If `edgeForReference()` will check for that and return Top or Bottom in this case the whole code would be cleaner and the switch case below would be always correct.

Hi @azoitl, my intention is that transitions should exit from the right edge of the figure when the state has actions. This is because the action area is on the right side and I want transitions to attach there with proper spacing. That is why I kept the RIGHT edge case separate with totalBounds. I have attached an image to show the expected behavior. Is this approach acceptable or do you suggest a different way to achieve this?

image

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateFigure.java
vikash1703: You're right. I've removed the getConnectionBounds() method and updated the anchor to use getNameLabel().getBounds() directly to keep the code clean and avoid unnecessary methods.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
azoitl: No in this case the transitions going to the states 1 - 3 should start on top and bottom and the whole right edge is a forbidden area.

vikash1703: > No in this case the transitions going to the states 1 - 3 should start on top and bottom and the whole right edge is a forbidden area.

@azolt

Ok then i will do it the same - what i was thinking and this part taken my most of the time and confusion because  was thinking when we write action on the right side and then draw the transtion it overlaps thats why i was making it like that no problem i will revert the changes.

vikash1703: > No in this case the transitions going to the states 1 - 3 should start on top and bottom and the whole right edge is a forbidden area.

@azoitl 
image
done

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,166 @@
...
+
+		for (final ECTransition t : list) {
+			final ECState other = (t.getSource() == state) ? t.getDestination() : t.getSource();
+			if (other == null) {
+				continue;
+			}
+
+			final Point otherPos = other.getPosition().toScreenPoint();
+
+			if (edgeForReference(center, otherPos) == edge) {
```
azoitl: for consistency you should use here the same point as you use in the getReference method. This would be the the transition bend point.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -0,0 +1,166 @@
...
+		}
+		for (int i = 0; i < list.size(); i++) {
+			if (list.get(i) == t) {
+				return i;
+			}
+		}
+		return 0;
+	}
+
+	private double angleForTransitionNormalized(final ECTransition t) {
```
azoitl: wouldn't a simple y or x position based on the transition's position and the edge be sufficient for not getting crossed transitions. Again using the other state here seems to be wrong.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
azoitl: nice

azoitl (approved):
Thx for all the work you put into this. It is really nice and works smoothly.

Two things I noticed that we could consider for future work is:

1. In `public ConnectionAnchor getSourceConnectionAnchor(final Request request) {` you currently pass null as transition. This method is called when a new transition is created. We may want a different connection ancor for that case. The advantage is that the ECStateConnectionAnchor would not allow null for transitions and many null checks could be removed.
2. For self-transitions currently a FixedAnchor is used. This anchor now does not fit well to your new one. Maybe we can also come up with a different anchor for that.

PR #2386, opened 2026-04-14
Refactored BulkEditor into multiple classes

sebHollersbacher:
Split code of BulkEditor plugin into multiple classes to make it more readable and easier to extend/change


plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorSettings.java
```diff
@@ -82,8 +80,7 @@ public BulkEditorSubSettings getSubSettings(final String ref) {
 	public void saveState(final IMemento memento) {
 		final IMemento childMemento = memento.createChild(TAG_BULKEDITOR_SETTINGS);
 		subSettingsMap.values().forEach(subSetting -> subSetting.saveState(childMemento));
-		childMemento.putInteger(MODE_TAG, modeSelection);
-		childMemento.putBoolean(ADVANCED_MODE_TAG, advancedMode);
+		childMemento.putInteger(MODE_TAG, modeSelection.ordinal());
```
mx990: If we are already making an incompatible change, I would suggest to save the enum `name()` instead of its `ordinal()` to be more robust in case of future changes.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorStateListener.java
```diff
@@ -0,0 +1,58 @@
...
+ *   Martin Erich Jobst - initial API and implementation and/or initial documentation
+ *   Sebastian Hollersbacher - Extracted to own class
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.bulkeditor.editors;
+
+import org.eclipse.fordiac.ide.model.ui.editors.LibraryElementProvider;
+import org.eclipse.fordiac.ide.model.ui.editors.LibraryElementStateListener;
+import org.eclipse.ui.IEditorInput;
+
+public class BulkEditorStateListener implements LibraryElementStateListener {
```
mx990: This class is closely tied to the internals of the bulk editor and extracting it exposes lots of functionality as public methods. I would therefore suggest leaving it as a nested class.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorOperationContextUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.eclipse.core.commands.operations.IOperationHistoryListener;
+import org.eclipse.core.commands.operations.IUndoContext;
+import org.eclipse.core.commands.operations.ObjectUndoContext;
+import org.eclipse.core.commands.operations.OperationHistoryEvent;
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.commands.ScopedOperation;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+
+public class BulkEditorOperationContextUpdater implements IOperationHistoryListener {
```
mx990: I believe this class is now independent from the bulk editor. I would suggest renaming it to `MultiLibraryElementOperationContextUpdater` and moving it to `org.eclipse.fordiac.ide.model.ui.editors`.

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/nattable/BulkEditorNatTable.java
```diff
@@ -0,0 +1,267 @@
...
+
+	private BulkEditorMode currentMode;
+	private NatTable natTable;
+
+	// Exactly one of the two providers is non-null at any given time; which one is
+	// valid is determined by currentMode / dynamicTable.
+	private ChangeableListDataProvider<VarDeclaration> varDeclProvider;
+	private ChangeableListDataProvider<Attribute> attributeProvider;
+	private SorterModel<VarDeclaration> varDeclarationSorterModel;
+	private SorterModel<Attribute> attributeSorterModel;
```
mx990: Can we refactor this so we do not have to distinguish between variables and attributes here? Maybe we could have each builder return a common interface that allows to access these in a generic manner?

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/search/FilterRecord.java
```diff
@@ -0,0 +1,46 @@
...
+
+	final Filter nameFilter;
+	final Filter typeFilter;
+	final Filter commentFilter;
+
+	final Pattern namePattern;
+	final Pattern typePattern;
+	final Pattern commentPattern;
+
+	FilterRecord(final boolean selected, final Filter nameFilter, final Filter typeFilter, final Filter commentFilter) {
```
mx990: Should this be package protected?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/editors/BulkEditorSettings.java
sebHollersbacher: I used `ordinal()` so I could still use `childMemento.getInteger(MODE_TAG)` and only lose the `ADVANCED_MODE_TAG`.

mx990: In that case, I believe we should remain fully backwards-compatible by looking for a `MODE_TAG` of type `String` with the `name()` first and then falling back to the previous version.

mx990 (approved):

azoitl (approved):

PR #2406, opened 2026-04-20
Validate version range in manifest

MandKastner:


plugins/org.eclipse.fordiac.ide.library.model/META-INF/MANIFEST.MF
```diff
@@ -18,3 +18,4 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.core.resources,
  org.eclipse.fordiac.ide.ui
 Bundle-ActivationPolicy: lazy
+Import-Package: org.eclipse.fordiac.ide.model.errormarker
```
oberlehner: why are you importing? the other packages are under Require-Bundle

oberlehner (approved):

PR #2407, opened 2026-04-21
Fix: Add PV to REQ event for CTU function blocks

franz-hoepfinger-4diac:
The preset value (PV) input was missing from the event-associated variables for the normal execution request (REQ) event. Including PV ensures it is correctly processed during the execution of all CTU counter variants.


franz-hoepfinger-4diac:
here the 1.0.0 should not be altered right ?

azoitl:
Yes as commented already several times to your PRs version 1.0 is only here for people how have projects from an 4diac IDE version between 2.0.1 and 3.0. Therefore this lib will be removed with one of the next releases. Most probably the 2026-12 release. Projects from 2.0.1 and earlier have anyhow a copy of there lib and do not need them and you should not make new projects with a 1.0 lib.

azoitl (approved):

PR #2408, opened 2026-04-21
Fix: Associate QO output with RT_E_CYCLE EO event

franz-hoepfinger-4diac:


azoitl:
This change is not only wrong it is a very bad idea. This would mean that the output data associated with the init thread would also be used by the thread that is executing the RT task.

franz-hoepfinger-4diac:
Understand. 
What Event shoud QO be bound to ?

azoitl:
good question. maybe it should just not be there

diplfranzhoepfinger:
> good question. maybe it should just not be there

I believe so. I researched some Publications from Vyatkin and Zoitl and both independently do not have this QO in the Pictures.

azoitl:
Also I would not see any case when this variable could be set to a sensible value.

PR #2410, opened 2026-04-21
Fix: Add configuration parameters to FB INIT events

franz-hoepfinger-4diac:


azoitl:
This PR mixes several things. Some are plain wrong and for none an explanation is given. The biggest problems are:

 1. updates on the 1.0 lib which we are not doing
 2. The Port block does not have an init event and should not have one. 

For ADS I don't know. If you think this is important please split and provide more reasoning in new PRs.

data/typelibrary/io-1.0.0/typelib/eliteboard/Port.fbt
```diff
@@ -5,6 +5,30 @@
...
+				<With Var="Pin14"/>
+				<With Var="Pin15"/>
+			</Event>
+		</EventInputs>
+		<EventOutputs>
+			<Event Name="INITO" Type="EInit" Comment="Initialization Confirm">
+			</Event>
+		</EventOutputs>
 		<InputVars>
 			<VarDeclaration Name="Pin0" Type="STRING" Comment="Name of the Pin" />
```
franz-hoepfinger-4diac, line 33: How can you connect this if there is no WITH ?

azoitl: it is not intended to be connected

franz-hoepfinger-4diac: OK. in many cases OK, you just put a constant String there. 
but what if you have the String coming from a *.ini File or something like this ? 

Somebody makes 1 Software with 2 different Configurations ?

azoitl: but then also the block in 4daic FORTE needs to be updated, because currently it is not capable of that feature. So sneaking this into a bigger PR is a bad idea,

PR #2415, opened 2026-04-27
Clear stale type library markers after successful type registration

oberlehner:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -283,24 +283,34 @@ public void addTypeEntry(final TypeEntry entry) {
...
-		if (isProgramTypeEntry(entry) && !addProgramTypeEntry(entry)) {
-			handleDuplicateTypeName(entry);
+		boolean programTypeEntryAdded = true;
+		if (isProgramTypeEntry(entry)) {
+			programTypeEntryAdded = addProgramTypeEntry(entry);
+			if (!programTypeEntryAdded) {
+				handleDuplicateTypeName(entry);
+			}
+		}
+		if (typeEntryAdded && programTypeEntryAdded) {
```
azoitl, line 312: shouldn't this be an or?

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
oberlehner: no because `programmTypeEntry `stays true, if `isProgramTypeEntry(entry)` returns false. So we also apply the deletion of stale markers to nonProgramTypeEntries

azoitl: ah now I see it. Feels a bit strange. Thx for the clarifiation

azoitl (approved):

PR #2418, opened 2026-04-28
Reworked how visible works for VarInOut

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/nat/VarDeclarationVisibleEditableRule.java
```diff
@@ -0,0 +1,57 @@
...
+		final VarDeclarationTableColumn column = columns.get(columnIndex);
+		final VarDeclaration varDecl = dataProvider.getRowObject(rowIndex);
+
+		if (column == VarDeclarationTableColumn.VISIBLE || column == VarDeclarationTableColumn.VISIBLEIN) {
+			return !isVisibleAndConnected(varDecl);
+		}
+		if (varDecl.isInOutVar() && column == VarDeclarationTableColumn.VISIBLEOUT) {
+			return !isVisibleAndConnected(varDecl.getInOutVarOpposite());
+		}
+		return VarDeclarationTableColumn.DEFAULT_EDITABLE.contains(column);
```
mx990: I would suggest turning this into a switch expression on `column`.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/properties/EditUntypedSubappVarInOutSection.java
```diff
@@ -72,7 +73,7 @@ public Command createCommand(final VarDeclaration rowObject, final VarDeclaratio
 				new VarDeclarationConfigLabelAccumulator(inputProvider, this::getAnnotationModel, columns));
 		final NatTableColumnProvider<VarDeclarationTableColumn> columnProvider = new NatTableColumnProvider<>(columns);
 		inputTable = NatTableWidgetFactory.createRowNatTable(parent, inputDataLayer, columnProvider,
-				getSectionEditableRule(), null, this, true);
+				new VarDeclarationVisibleEditableRule(inputProvider, columns), null, this, true);
```
mx990: I believe we still need to handle whether the section itself is editable, see https://github.com/eclipse-4diac/4diac-ide/blob/6c4216bc4130e47bdf1d13cf3c6a986da8df9afc/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/AbstractEditVarInOutSection.java#L207-L222

How about passing a delegate editable rule to the constructor of `VarDeclarationVisibleEditableRule`?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
```diff
@@ -59,31 +61,44 @@ public void undo() {
...
+            return true;
+        }
+        
+	    if (element.getBlockFBNetworkElement() instanceof final UntypedSubApp subApp && subApp.isUnfolded()) {
+	        return hasInput || hasOutput;
+	    }
+	    
+	    if (element.getBlockFBNetworkElement() instanceof final TypedSubApp tsa
+	            && element instanceof final VarDeclaration vd && vd.isInOutVar()) {
+	        final var typePin = tsa.getType().getInterfaceList().getInterfaceElement(element);
```
mx990: Please use `IInterfaceElement.findInTypeInterface()`.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
```diff
@@ -59,31 +61,44 @@ public void undo() {
...
-		}
-		return false;
+	public static boolean isPinConnected(final IInterfaceElement element) {
+	    final boolean isInput = element.isIsInput();
+	    final boolean hasInput = !element.getInputConnections().isEmpty();
+	    final boolean hasOutput = !element.getOutputConnections().isEmpty();
+
+	    if (element.getFBType() instanceof SubAppType) {
+	        return isInput ? hasOutput : hasInput;
+	    }
```
mx990: I believe this should only apply to InOuts?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
```diff
@@ -59,31 +61,44 @@ public void undo() {
...
-	private static boolean isConnectedInsideSubApp(final IInterfaceElement element) {
-		if (element instanceof final VarDeclaration varDecl && element.isIsInput()
-				&& element.eContainer().eContainer() instanceof final SubAppType sat) {
-			return sat.getFBNetwork().getBlockFBNetworkElements().anyMatch(fb -> {
-				final IInterfaceElement internal = fb.getInterface().getInterfaceElement(varDecl);
-				return (internal != null && !internal.getInputConnections().isEmpty());
-			});
-		}
-		return false;
+	public static boolean isPinConnected(final IInterfaceElement element) {
```
mx990: I would suggest to rename this to `canHidePin` or something similar, since this clearly does more than just check if the pin is connected.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
```diff
@@ -59,31 +61,44 @@ public void undo() {
...
+
+	    if (element.getFBType() instanceof SubAppType) {
+	        return isInput ? hasOutput : hasInput;
+	    }
+
+        if (isInput ? hasInput : hasOutput) {
+            return true;
+        }
+        
+	    if (element.getBlockFBNetworkElement() instanceof final UntypedSubApp subApp && subApp.isUnfolded()) {
```
mx990: I believe this should also apply to input InOuts in collapsed untyped subapps?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
```diff
@@ -59,31 +61,44 @@ public void undo() {
...
+	        return hasInput || hasOutput;
+	    }
+	    
+	    if (element.getBlockFBNetworkElement() instanceof final TypedSubApp tsa
+	            && element instanceof final VarDeclaration vd && vd.isInOutVar()) {
+	        final var typePin = tsa.getType().getInterfaceList().getInterfaceElement(element);
+	        return typePin.isIsInput()
+	                ? !typePin.getOutputConnections().isEmpty()
+	                : !typePin.getInputConnections().isEmpty();
+	    }
```
mx990: I believe this should only apply to inputs, since hiding outputs that are connected on the inside is still valid.

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/messages.properties
mx990, line 1: I would advise against combining all messages in `plugin.properties`, since the *Externalize Strings* wizard in Java code has problems handling messages that are not in a source folder.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationAnnotations.java
```diff
@@ -286,6 +288,49 @@ public static boolean validateVarInOutSubappNetwork(final VarDeclaration varDecl
...
+								MessageFormat.format(Messages.VarDeclarationAnnotations_HiddenPinConnected,
+										varDeclaration.getName()),
+								FordiacMarkerHelper.getDiagnosticData(varDeclaration)));
+			}
+			return false;
+		}
+		return true;
+	}
+
+	private static boolean isPinConnected(final IInterfaceElement element) {
```
mx990: It looks like this is duplicated in `ChangePinVisibilityCommand`. I would suggest to move this to a common helper class.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/ChangePinVisibilityCommand.java
sebHollersbacher: The problem with why I didn't use `IInterfaceElement.findInTypeInterface()` is that the typePins I get do not contain the inside connections which I need to check.

mx990: Ah, thanks for clarifying. Could you please add a comment in the source, so that we do not mistakenly change this later?

mx990 (approved):

azoitl (approved):

PR #2421, opened 2026-04-30
Fix ecc transition splines

vikash1703:


azoitl:
I think before investigating deeper we should do a bit of clean-up first. Draw2d has a `PrecisionPoint` and a `Vector` class. With that your code would be much better readable and easier to understand. 

Also I think it would make sense if you can find a way that you can expose the getDirection to get the TOP, BOTTOM, LEFT, RIGHT, direction of your start and endpoint. Maybe you can even extract some parts into methods. I guess that would also help alot. Especially the changes in the spline connection would benefit from that.

vikash1703:
> I think before investigating deeper we should do a bit of clean-up first. Draw2d has a `PrecisionPoint` and a `Vector` class. With that your code would be much better readable and easier to understand.
> 
> Also I think it would make sense if you can find a way that you can expose the getDirection to get the TOP, BOTTOM, LEFT, RIGHT, direction of your start and endpoint. Maybe you can even extract some parts into methods. I guess that would also help alot. Especially the changes in the spline connection would benefit from that.

@azoitl 
the reason i have created pr is because your understand is better and your suggest would help me yes i will follow the same what your thought.

Thankyou.

vikash1703:
> I think before investigating deeper we should do a bit of clean-up first. Draw2d has a `PrecisionPoint` and a `Vector` class. With that your code would be much better readable and easier to understand.
> 
> Also I think it would make sense if you can find a way that you can expose the getDirection to get the TOP, BOTTOM, LEFT, RIGHT, direction of your start and endpoint. Maybe you can even extract some parts into methods. I guess that would also help alot. Especially the changes in the spline connection would benefit from that.

Hi @azoitl,

SplineConnection.java: I replaced the old getPath() calculation with a clean 7-point design. The class now expects exactly 7 points from BendpointConnectionRouter. A named constant SPLINE_POINT_COUNT = 7 replaces the magic number. I also removed the debug dots for the final clean-up.

ECTransitionFigure.java: I did the full clean-up you asked for. I replaced all manual double arithmetic with Draw2d Vector and PrecisionPoint. I extracted EdgeDirection as a package-visible enum. I extracted the routing logic into focused private methods (isReadyToRoute, calcOrthogonalControlPoint, calcAverageTangent). I also applied your coordinate fix: I call translateToAbsolute() on P4 before passing it to the anchors, and then translateToRelative() on all points before giving them to the router. This fixed the zoom and pan distortion!

ECStateConnectionAnchor.java: I reused the EdgeDirection enum from ECTransitionFigure so the rule is in one place. I fixed getReferencePoint() to call translateToAbsolute() on the model position so the anchor reference is always in absolute screen coordinates as GEF requires.

Two problems I am still facing and would love your advice on:

Problem 1 - Magnetic feel during drag:
When dragging a transition bendpoint, it feels like it snaps or vibrates slightly instead of moving smoothly. I believe this is a jitter loop: refreshVisuals() calls updateBendPoints(), which calls setConstraint() on the router. The router calls route(), which triggers a figure layout change. The EMF adapter sees this and calls refresh() again. I added a JITTER_TOLERANCE of 3 pixels, and it reduced the effect but did not fully stop it. I am not sure if my translateToRelative() approach is creating a small but persistent float-to-int coordinate difference on each cycle that keeps the loop alive. Do you have a suggestion for the right place to break this cycle?

Problem 2 - One transition changes the shape of another:
When I draw a second transition from the same state, the anchor spacing logic in ECStateConnectionAnchor redistributes the slots on the edge. The first transition's anchor point shifts. But the first transition's spline control points (P2 and P6) are not recalculated at the right moment, so its curve shape changes incorrectly. I tried refreshing neighbour transitions from both ECStateEditPart and ECTransitionEditPart earlier, but the timing is off—the refresh sometimes happens before the new anchor slot is finalized. Is there a GEF pattern for triggering a delayed or post-layout refresh that would fire after all anchors have settled?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -140,22 +192,151 @@ public void setConditionText(final String conditionText) {
...
+	// ensures that
+	// very short segments do not disproportionately tilt the curve (preventing
+	// sharp V-shapes).
+	private static Vector calcAverageTangent(final Vector seg1, final double len1, final Vector seg2,
+			final double len2) {
+		final Vector dir1 = seg1.getDivided(len1);
+		final Vector dir2 = seg2.getDivided(len2);
+
+		final double totalLen = len1 + len2;
+		final Vector weighted = new Vector((dir1.x * len1 + dir2.x * len2) / totalLen,
```
azoitl: with multiplying the dir1 and dir2 with the lens you do the oposite of the lines 304 adn 306. I don't think this is a good idea. I would also not dived by total len.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -140,22 +192,151 @@ public void setConditionText(final String conditionText) {
...
+		// transition figure.
+		// This completely fixes the zoom/panning distortion and ensures vector math is
+		// accurate.
+		translateToRelative(p1Abs);
+		translateToRelative(p4Abs);
+		translateToRelative(p7Abs);
+
+		// REFACTOR: Replaced manual double arithmetic with Draw2d PrecisionPoint and
+		// Vector classes
+		// for better readability and easier understanding.
```
azoitl: By imediatly wrapping the points at their creation in a precision point you can save these 3 lines which would make the code better readable.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -140,22 +192,151 @@ public void setConditionText(final String conditionText) {
...
+
+		final PrecisionPoint p2 = calcOrthogonalControlPoint(p1, getSourceAnchor().getOwner(),
+				CTRL_POINT_FACTOR * len1);
+		final PrecisionPoint p6 = calcOrthogonalControlPoint(p7, getTargetAnchor().getOwner(),
+				CTRL_POINT_FACTOR * len2);
+
+		final Vector tangent = calcAverageTangent(seg1, len1, seg2, len2);
+		final double ctrlDist = Math.min(CTRL_POINT_FACTOR * Math.min(len1, len2), MAX_HANDLE_DISTANCE);
+
+		final PrecisionPoint p3 = translate(p4, tangent, -ctrlDist);
```
azoitl: better use len1 and len2 here and not then min.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -44,6 +49,55 @@
...
+	// Absolute/Relative translation.
+	private static final int JITTER_TOLERANCE = 3;
+
+	// FIX: Extracted EdgeDirection enum to expose getDirection (TOP, BOTTOM, LEFT,
+	// RIGHT)
+	// for both this class and ECStateConnectionAnchor as requested.
+	enum EdgeDirection {
+		TOP, BOTTOM, LEFT, RIGHT;
+
+		static EdgeDirection of(final PrecisionPoint anchor, final Rectangle bounds) {
```
azoitl: is this code now duplicated to code below in the ancor?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/SplineConnection.java
```diff
@@ -8,64 +8,74 @@
...
-		p.dispose();
+	protected void outlineShape(final Graphics g) {
+		final PointList points = getPoints();
+		if (points.size() != 7) {
+			super.outlineShape(g);
+			return;
+		}
+		final Path path = new Path(null);
+		try {
+			path.moveTo(points.getPoint(0).x, points.getPoint(0).y);
```
azoitl: please extract the duplicated path generation into a method as it was.

azoitl (changes requested):
@vikash1703 thanks for the update it is now much better readable. 

Regarding your update problems: I think we should first get the base spline and routing alg correct then we can in separate steps fix the update. I think the reason for the update problem is that we are calling the different refereshes to often or at the wrong time. But as said lets first fix the algorithm.

Therefore I also would suggest to remove any checking code like the ready to rout or similar things. As this code my introduce false pasitives that are hard to track later.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
azoitl: Sorry mixed up the code. Meant the code in the ECStateConnectionAnchor.

azoitl (commented):
I thought a bit more about this PR and it get quite big already. I think it would be good to split it up a bit. When good smaller first PR could be to extract `EdgeDirection` into an own file and use the `of` method in the `ECStateConnectionAnchor`. This would make `ECStateConnectionAnchor` cleaner and this PR smaller. 

Furthermore what would help me a lot is if you could keep the order of untouched functions. As you moved setTransitionOrder, getLabel, and getToolTip additional noise was added to the diff. Please avoid that.

vikash1703:
> @vikash1703 thanks for the update it is now much better readable.
> 
> Regarding your update problems: I think we should first get the base spline and routing alg correct then we can in separate steps fix the update. I think the reason for the update problem is that we are calling the different refereshes to often or at the wrong time. But as said lets first fix the algorithm.
> 
> Therefore I also would suggest to remove any checking code like the ready to rout or similar things. As this code my introduce false pasitives that are hard to track later.

@azoitl 
I completely agree with your plan. Let's focus on getting the base spline math and routing algorithm correct first, and we can tackle the update/refresh issues in a separate step.

Just to share what I found regarding the root cause of that refresh loop (the magnetic vibrating feel):
When I drag the transition bendpoint, updateBendPoints() calculates the new control points. Because I have to translate between Absolute and Relative coordinates, there is a tiny 1-pixel rounding difference when converting my PrecisionPoint (doubles) back to integer pixels.

When setConstraint() passes these rounded points to the router, GEF detects a tiny layout change. The EMF adapter catches this layout change and fires refreshVisuals() again. This recalculates the points, causes another 1-pixel rounding shift, and creates an infinite loop. That is what causes the jitter.

But as you suggested, I will ignore this loop for now. I will remove the isReadyToRoute checks so we don't hide any false positives, and I will focus entirely on cleaning up the math, fixing the tangents, and splitting the EdgeDirection into a separate PR.

azoitl:
@vikash1703 thx for sharing your findings. This will definitely help us later.

vikash1703:
@azoitl 
 still it needs improvement first you test at your end and see how spline are reacting to our actions.

This is taking very long time i will give more time to not bother you very much

Thankyou.

azoitl:
@vikash1703 I will test. But it would be great if you could do the `EdgeDirection` refactoring on the base branch in parallel in an independant PR.

vikash1703:
> @vikash1703 I will test. But it would be great if you could do the `EdgeDirection` refactoring on the base branch in parallel in an independant PR.

Hmm just i am doing

vikash1703:
@azoitl 

Have you tested spline behaviour in your local? 

I think two thing have to be corrected the one is when two transition are drawn from same state they change the shape .
and our spline is not draw wider circle like its like pointed when you drag more ! today i will do that .

azoitl:
Yes self transitions are something that we need to treat as well. Again I would move that to a next step. 

I tested a bit and noticed a few things that needs to be checked:

1. I think there is an issue with the new EdgeDirection.of method. When there is an action all transitions are on the left side
2. This PR needs to be rebased on top of latest develop to work
3. at least in my tests the the p1 and p6 where always a bit of
4. This is again for a next PR: all spline control points are now shown as bendpoints when the transition is selected and can be adjusted be the user. This will should not be possible.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -14,17 +14,17 @@
  *   Alois Zoitl - extracted TransitionFigure code and changed to cubic spline
  *   Bianca Wiesmayr, Ernst Blecha - added tooltip
  *   Alois Zoitl - modernized and reworked ECC look
- *******************************************************************************/
+*******************************************************************************/
```
azoitl: please do not delete leading spaces here.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -186,4 +244,58 @@ public ECTransitionToolTipFigure getToolTip() {
...
+
+	private static Vector calcAverageTangent(final Vector seg1, final Vector seg2) {
+		final double len1 = Math.max(seg1.getLength(), EPSILON);
+		final double len2 = Math.max(seg2.getLength(), EPSILON);
+		final Vector dir1 = seg1.getDivided(len1);
+		final Vector dir2 = seg2.getDivided(len2);
+		final Vector tangent = new Vector(dir1.x + dir2.x, dir1.y + dir2.y);
+		final double len = tangent.getLength();
+		if (len < EPSILON) {
+			return new Vector(-dir1.y, dir1.x);
```
azoitl: this line looks very wrong

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -186,4 +244,58 @@ public ECTransitionToolTipFigure getToolTip() {
...
+		final Vector normal = EdgeDirection.of(anchorPt, bounds).toNormal();
+		return translate(anchor, normal, distance);
+	}
+
+	private static Vector calcAverageTangent(final Vector seg1, final Vector seg2) {
+		final double len1 = Math.max(seg1.getLength(), EPSILON);
+		final double len2 = Math.max(seg2.getLength(), EPSILON);
+		final Vector dir1 = seg1.getDivided(len1);
+		final Vector dir2 = seg2.getDivided(len2);
+		final Vector tangent = new Vector(dir1.x + dir2.x, dir1.y + dir2.y);
```
azoitl: vector has a `getAdded` this would be much cleaner

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -186,4 +244,58 @@ public ECTransitionToolTipFigure getToolTip() {
...
+
+	private static PrecisionPoint translate(final PrecisionPoint base, final Vector direction, final double distance) {
+		return new PrecisionPoint(base.preciseX() + direction.x * distance, base.preciseY() + direction.y * distance);
+	}
+
+	private static AbsoluteBendpoint toBendpoint(final PrecisionPoint p) {
+		return new AbsoluteBendpoint(new Point((int) Math.round(p.preciseX()), (int) Math.round(p.preciseY())));
+	}
+
+	private boolean hasSignificantChange(final List<Bendpoint> newConstraints) {
```
azoitl: I would prefere to have it for now without the hasSignificantChange

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/SplineConnection.java
```diff
@@ -14,58 +14,51 @@
...
-					getPoints().getMidpoint().x - (float) startEnd30.x,
-					getPoints().getMidpoint().y - (float) startEnd30.y, getPoints().getMidpoint().x,
-					getPoints().getMidpoint().y);
-
-			p.cubicTo(getPoints().getMidpoint().x + (float) startEnd30.x,
-					getPoints().getMidpoint().y + (float) startEnd30.y, getEnd().x - (float) midEnd30.x,
-					getEnd().y - (float) midEnd30.y, getEnd().x, getEnd().y);
+		final Path p = new Path(null);
+		final PointList points = getPoints();
+		if (7 == points.size()) {
```
azoitl: please no yoda expressions, also I'm not sure if you need that check again.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -186,4 +244,58 @@ public ECTransitionToolTipFigure getToolTip() {
...
+		if (owner instanceof final ECStateFigure stateFigure) {
+			bounds = stateFigure.getNameLabel().getBounds().getCopy();
+			stateFigure.getNameLabel().translateToAbsolute(bounds);
+		} else {
+			bounds = owner.getBounds().getCopy();
+			owner.translateToAbsolute(bounds);
+		}
+		translateToRelative(bounds);
+
+		final Point anchorPt = new Point((int) Math.round(anchor.preciseX()), (int) Math.round(anchor.preciseY()));
```
azoitl: why not just using a toPoint or use the result of the precise point?

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -186,4 +244,58 @@ public ECTransitionToolTipFigure getToolTip() {
...
+		final Vector normal = EdgeDirection.of(anchorPt, bounds).toNormal();
+		return translate(anchor, normal, distance);
+	}
+
+	private static Vector calcAverageTangent(final Vector seg1, final Vector seg2) {
+		final double len1 = Math.max(seg1.getLength(), EPSILON);
+		final double len2 = Math.max(seg2.getLength(), EPSILON);
+		final Vector dir1 = seg1.getDivided(len1);
```
azoitl: would it be a good idea to add a `getNormalized`  method to Vector. With that this whole method could be reduced to 2 lines:
```
final Vector tangent = len1.getNormalized().getAdded(len2.getNormalized);
return tangent.getNormalized();
```

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -141,9 +150,58 @@ public void setConditionText(final String conditionText) {
...
-		final List<Bendpoint> bendPoints = new ArrayList<>();
-		bendPoints.add(new AbsoluteBendpoint(ecTransition.getPosition().toScreenPoint()));
-		getConnectionRouter().setConstraint(this, bendPoints);
+		if (ecTransition == null || ecTransition.getPosition() == null || getSourceAnchor() == null
+				|| getTargetAnchor() == null || getSourceAnchor().getOwner() == null
+				|| getTargetAnchor().getOwner() == null) {
+			return;
+		}
+
+		if (getSourceAnchor().getOwner().getBounds().width == 0
```
azoitl: I don't think this check is needed

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -141,9 +150,58 @@ public void setConditionText(final String conditionText) {
...
+			return;
+		}
+
+		final PrecisionPoint p4 = new PrecisionPoint(ecTransition.getPosition().toScreenPoint());
+		translateToAbsolute(p4);
+
+		final Point loc1 = getSourceAnchor().getLocation(p4);
+		final Point loc7 = getTargetAnchor().getLocation(p4);
+
+		if (loc1 == null || loc7 == null) {
```
azoitl: I don't think that an `Anchor` should ever return null as location. If this is the case for your new anchor I think this should be fixed on anchor side and not here. Then the lines 167 to 175 can be reduced to just to lines and the variables `loc1` and `loc7` would not be needed making the code much better readable.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECTransitionFigure.java
```diff
@@ -141,9 +150,58 @@ public void setConditionText(final String conditionText) {
...
+		}
+
+		final PrecisionPoint p1 = new PrecisionPoint(loc1);
+		final PrecisionPoint p7 = new PrecisionPoint(loc7);
+
+		translateToRelative(p1);
+		translateToRelative(p4);
+		translateToRelative(p7);
+
+		final Vector seg1 = new Vector(p4.preciseX() - p1.preciseX(), p4.preciseY() - p1.preciseY());
```
azoitl: Sorry to bother you with another thing but Vector has a constructor with two points start and end.

vikash1703:
@azoitl 

what i wanted to say is that we are now using new Vector(p1, p4) actually calculates backwards (p1 - p4) in the framework. This twists the line. My old code (p4.x - p1.x) kept the direction forward.
2. Passing p4 in getLocation(p4) makes the anchor jump to the wrong face of the state.

azoitl:
> @azoitl
> 
> what i wanted to say is that we are now using new Vector(p1, p4) actually calculates backwards (p1 - p4) in the framework. This twists the line. My old code (p4.x - p1.x) kept the direction forward. 2. Passing p4 in getLocation(p4) makes the anchor jump to the wrong face of the state.

@vikash1703 That is not true. `Vector`'s implementation for this constructor is calculating end - start as it should do:
```
       public Vector(PrecisionPoint start, PrecisionPoint end) {
		x = end.preciseX() - start.preciseX();
		y = end.preciseY() - start.preciseY();
	}
```

vikash1703:
@azoitl 

Hii i have gone though your pr#1117  now we can use directly now but please when you get free have a look what i have missed so our spline at first creation get curvy.

azoitl:
@vikash1703 it took me a bit to analyze the problem and I think I have to apologize to you. I guided you in the wrong direction. Now that I experimented a bit I think the problem is that `updateBendPoints` is called to early and often. Also I think with your changes we are doing something in `updateBendPoints` we should do somewhere else. 

The problem as you also see it is that as `updateBendPoints` is called to early the states the transition is connected to are not yet fully created or fully drawn. Therefore some of the control points you are calculating get the wrong data and with that we get this loops. 

After looking a bit into the Eclipse GEF Classic and debugging a bit I think the cleaner and also more efficient approach would be to implement an own `ECCTransitionRouter`. This router would then take the original bendpoint as we we had in the `updateBendPoints` method and the connection anchor positions and calculate the full 7 points for our spline and give it to the `ECtransitionFigure` via the `setPoints` method.

The good thing is that the current logic you have in the new `updateBendPoints` can be nearly one to one moved to the new `ECCTransitionRouter` so nothing is lost and your work here was an important step stone.

Do you want to introduce `ECCTransitionRouter` and migrate your code or should I do the basic infrastructure setup and then hand over to you.

azoitl:
I was to quick with hitting the submit button. We could also merge this PR to have it in our code base and then migrate to the `ECCTransitionRouter` in a next step. I would like this approach even a bit more. 

What do you think @vikash1703?

vikash1703:
> @vikash1703 it took me a bit to analyze the problem and I think I have to apologize to you. I guided you in the wrong direction. Now that I experimented a bit I think the problem is that `updateBendPoints` is called to early and often. Also I think with your changes we are doing something in `updateBendPoints` we should do somewhere else.
> 
> The problem as you also see it is that as `updateBendPoints` is called to early the states the transition is connected to are not yet fully created or fully drawn. Therefore some of the control points you are calculating get the wrong data and with that we get this loops.
> 
> After looking a bit into the Eclipse GEF Classic and debugging a bit I think the cleaner and also more efficient approach would be to implement an own `ECCTransitionRouter`. This router would then take the original bendpoint as we we had in the `updateBendPoints` method and the connection anchor positions and calculate the full 7 points for our spline and give it to the `ECtransitionFigure` via the `setPoints` method.
> 
> The good thing is that the current logic you have in the new `updateBendPoints` can be nearly one to one moved to the new `ECCTransitionRouter` so nothing is lost and your work here was an important step stone.
> 
> Do you want to introduce `ECCTransitionRouter` and migrate your code or should I do the basic infrastructure setup and then hand over to you.

@azoitl 

why not i am always happy to make our implementation cleaner thats why i have left and waiting for your observation as i am not able to analyse and also wanted to say that you have guided in a right way so nothing to say sorry you are my teacher i am happy to follow you thought .

thanks

vikash1703:
> I was to quick with hitting the submit button. We could also merge this PR to have it in our code base and then migrate to the `ECCTransitionRouter` in a next step. I would like this approach even a bit more.
> 
> What do you think @vikash1703?

yes then i also dont have to worry about changes sometime i lost my changes 😅 sure we will merge this PR then i will migrate to ECCTransitionRouter in next step

azoitl (approved):

azoitl:
> > I was to quick with hitting the submit button. We could also merge this PR to have it in our code base and then migrate to the `ECCTransitionRouter` in a next step. I would like this approach even a bit more.
> > What do you think @vikash1703?
> 
> yes then i also dont have to worry about changes sometime i lost my changes 😅 sure we will merge this PR then i will migrate to ECCTransitionRouter in next step

I definitely don't want to loose this state. For this you came to far. Also I think it is cleaner when you continue from this state.

vikash1703:
@azoitl 

I have one thing to clarify as i am observing what we have did we will replace each changes that we have done in our ECTransitionFigure to new class and we make it default as we have early  in my understanding and we have to register our new router class in ECCEditPart by overriding.

vikash1703:
> @azoitl
> 
> I have one thing to clarify as i am observing what we have did we will replace each changes that we have done in our ECTransitionFigure to new class and we make it default as we have early in my understanding and we have to register our new router class in ECCEditPart by overriding.

and on thing also i cant able to user getNormalize method right know as it is not merged .

azoitl:
> @azoitl
> 
> I have one thing to clarify as i am observing what we have did we will replace each changes that we have done in our ECTransitionFigure to new class and we make it default as we have early in my understanding and we have to register our new router class in ECCEditPart by overriding.

Exactly!

> and on thing also i cant able to user getNormalize method right know as it is not merged .

Yes I know. Also if it would be merged it would not be immediately available as 4diac IDE is consuming milestone builds and the next milestone build would get available earliest next Friday.

vikash1703:
> > @azoitl
> > I have one thing to clarify as i am observing what we have did we will replace each changes that we have done in our ECTransitionFigure to new class and we make it default as we have early in my understanding and we have to register our new router class in ECCEditPart by overriding.
> 
> Exactly!
> 
> > and on thing also i cant able to user getNormalize method right know as it is not merged .
> 
> Yes I know. Also if it would be merged it would not be immediately available as 4diac IDE is consuming milestone builds and the next milestone build would get available earliest next Friday.

ok and one thing i also have to ask in my ide i dont have my original ectransitiofigure i have modified one and on 4diac ide repo we hjave modified one as we have merged so how i can make it the original default one its a silly question but it save my time

azoitl:
In Eclipse IDE you have the history view there you can see each commit and for each commit the changed files and the for each file the changes. This can help you to identify your changes to extract them. What I also like is the diff view here in Github. you can access it in the menu above under files changed: https://github.com/eclipse-4diac/4diac-ide/pull/2421/changes

There you can also nicely compare the changes and see the before and after state. This is what I use for cases like yours.

PR #2422, opened 2026-05-01
[Replay Debugging] Allow deleting events and timelines

cochicde:
Timelines contains events and other timelines. These can be removed from the GUI.

- Timeline had two similar listener interfaces, one for events added and another for timelines added. These were merged into a StructureListener, which now contains also callbacks for removing events and timelines.
- Since right-click support is added, I changed the way edit parts are catching events. 
  - Resources catch only key events (to move along the timeline with they keyboard)
  - Timelines catch left and right clicks -> this is not quite nice as the events themselves are note edit parts but rather just figures, so the mechanism to catch mouse events is tangled. This requires an extra refactoring, but not in this PR.
  - A timeline without parent cannot be deleted.
  - Undo/redo also work
  - Deleting events remove all events from the selected one to the right end. Timelines spawned from these deleted events are therefore also deleted.
  - Deleting timelines with children, also deletes the children timelines. 

Localisation of strings was also done, as promised.

Screenshot 2026-05-01 16 43 48


cochicde:
I'm not sure where the erros come from. @azoitl is there something wrong with some configuration or is it something I changed? 

```
Error:  Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:5.0.2:test (default-test) on project org.eclipse.fordiac.ide.contractspec.tests: There are test failures.
Error:  
Error:  Please refer to /home/runner/work/4diac-ide/4diac-ide/plugins/org.eclipse.fordiac.ide.contractspec.tests/target/surefire-reports for the individual test results.

```

azoitl:
The detailed output says something about some wrong or missing Junit files. Looks like a version clash or that something could not loaded via maven. We have that sometimes. I hope it is an infra topic and re-triggered the builds.

azoitl:
Looks like we have to switch our XText tests to Junit6 as with todays 2026-06M2 we got the latest XText and they migrated all the tests: https://github.com/eclipse-xtext/xtext/pull/3560

Edit:  
Will do an own PR for this.

azoitl:
with PR #2424 test issues are finally fixed. Please rebase your PR then your PR should run through nicely.

azoitl (changes requested):
A more modern approach is to not use actions directly but use Eclipse Command/Handler pattern. This gives you normally more flexibility of declaring and adding context menus, main menus and toolbars. Also it better decouples the code.

cochicde:
I did the transition to the new approach. I'm not sure about the conventions of names, where to register what, where the constants ID go, who calls whom and so on. Please double check that since I have a  bunch of other menu context on the pipeline after this, for which I will just follow the same pattern.

azoitl:
Ah damn, I wanted to point you yesterday to the `FordiacContextMenuProvider`. There we set all the constants and do default setup. Also GEF has a default delete action that should be activated as soon as you provide a delete command through the right editpolicy. This is nicely explained in this blog: https://vainolo.com/2011/07/10/creating-a-gef-editor-%e2%80%93-part-8-delete-undo-and-redo/

So it is a git of a mixture of actions, GEF Commands, and E3 commands/handlers.

cochicde:
I got back to this PR, but I'm now more confused. As I have it now
1. I have handlers that register to the `IHandlerService` of the site. 
2. When building the menu, I add the corresponding entries using the ID from the handlers.
3. The handlers know when to enable or not, and on execute the send the request to the EditPart
4. The EditPart then gets the commands and gets from my policy the command.
5. The edit part adds the command to the stack of commands.
There's no place for actions anymore. However the `FordiacContextMenuProvider` expect an ActionRegistry, which now I don't have anymore. Also, it tries to add all the `align` actions, which I'm not using. The only thing I could reuse is 

```
	private static void addActionGroups(final IMenuManager menu) {
		menu.add(new Separator(GEFActionConstants.GROUP_EDIT));
		menu.add(new Separator(GROUP_NAVIGATE));
		menu.add(new Separator(GEFActionConstants.GROUP_COPY));
		menu.add(new Separator(GEFActionConstants.GROUP_VIEW));
		menu.add(new Separator(GEFActionConstants.GROUP_FIND));
		menu.add(new Separator(IWorkbenchActionConstants.GROUP_ADD));
		menu.add(new Separator(GEFActionConstants.GROUP_SAVE));
		menu.add(new Separator(GEFActionConstants.GROUP_REST));
		menu.add(new Separator(IWorkbenchActionConstants.GROUP_REORGANIZE));
		menu.add(new Separator(IWorkbenchActionConstants.MB_ADDITIONS));
	}
```

But it seems that I need to bring back the actionRegistry again only to have these separators.

Regarding the delete command,  I'm not sure what would imply that is activated by GEF. Since when an event is selected, you could either delete the events to its right or the whole timeline, I'm not sure if this will introduce more usability issues.

azoitl:
I don't know why you have the handler service in your view. What I see is missing is that you add your commands to any menus or context menus. Is that maybe the thing that is missing?

GEF has its own delete action that will interact with the selected edtiparts and automatically hooks to any menus and context menus. At least this is how we do it on other GEF based editors.

azoitl:
@cochicde you can see in PTR #2453 how @sebHollersbacher did a similar thing for the bulk editor.

cochicde:
I think I'm running in circles. I have now three examples all with different approaches. The link doesn't use handlers, `FordiacContextMenuProvider` uses actions, which I thought I shouldn't use. The PR doesn't use `FordiacContextMenuProvider` which I thought I should use. 

What would be plugin that I can use as a proper example? The two plugins that I see are using  `FordiacContextMenuProvider` are `fbTypeEditor` and `systesConfiguration` which don't have handlers.

azoitl:
Hmh that is bad. I would say the most complete example would be the application editor. This is a rather complicated editor but there you could check the plugin.xml for the command handling and context menu generation. I hope this helps.

cochicde:
I added the `FordiacContextMenuProvider`. Not sure how much it helps though. I couldn't find how to use the handler that would make sense to this case.

cochicde:
Hi @azoitl. I updated the code now to avoid actions and only have commands and handlers. I added the conditions in the plugin.xml. 

This seems more clean and correct, but I still wanted your feedback to see where it can be improved. I have the sense that the code in the handlers can be improved or reused, but I'm not sure. 

This is still a draft because selection, moving with the keyboard and some edge cases are still not fully working, but I wanted to make sure the design is correct before going further. Only 2 commits now, better split

cochicde:
I fixed the navigation and selection issues. The problem to fix was that we always have an event selected. However, navigating is done at the resource level. So now, when an event is selected and any of the navigation key is pressed, the command is bubbled up to the timeline and then the resource edit part which handles these.
I had to therefore changed from actions to commands/handler which made things easier.

Deleteing timelines also require its specisal request and cannot use the standard `REQ_DELETE` since the editPart is selected and it can handle the `REQ_DELETE` request, the request would end up always deleting just the event.

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/plugin.xml
```diff
@@ -24,4 +24,127 @@
...
+      </command>
+
+    </menuContribution>
+  </extension>
+
+  <!-- Key bindings -->
+<extension point="org.eclipse.ui.bindings">
+    <key commandId="org.eclipse.fordiac.ide.debug.replaydebugging.ui.moveLeft"
+         schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
+         contextId="org.eclipse.ui.contexts.window"
```
azoitl: binding keys to the global context is very dangerous. especially so generic keys as the arrow keys. This will lead most probably lead to conflicting key handlers. You can have a look what we did for key handlers and context in the other GEF editors. However in order to get this PR finally merged I think it would be best to  do the key handling in a seperate PR.

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/editpart/TimelineEditPart.java
```diff
@@ -234,6 +256,10 @@ public void propertyChange(final PropertyChangeEvent evt) {
 			safeRefresh(false, true);
 		} else if (evt.getPropertyName().equals(TimelineModel.PROPERTY_TIMELINE_ADDED)) {
 			safeRefresh(true, false);
+		} else if (evt.getPropertyName().equals(TimelineModel.PROPERTY_EVENT_DELETED)) {
```
azoitl: maybe for a future clean-up with latest java this can be done with a switch case in a better readable way.

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/command/DeleteEventsCommand.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.eclipse.fordiac.ide.debug.replaydebugging.core.EventChange;
+import org.eclipse.fordiac.ide.debug.replaydebugging.core.Timeline;
+import org.eclipse.fordiac.ide.debug.replaydebugging.ui.Messages;
+import org.eclipse.gef.commands.Command;
+
+/**
+ * @brief Command to delete an event from a timeline
+ *
+ *        This command supports undo/redo operations through the GEF
```
azoitl: as this is a general property of Eclipse GEF Classic `Command` I don't think this is needed. Same for the commands below.

azoitl (changes requested):
I appreciated that you original split this PR in smaller commits. However as you are now adding "repair" commits on top the idea of small commits and a clean history as we describe it in our [contributing guides](https://eclipse.dev/4diac/doc/development/contribute.html) is not followed. 

For future references if you want to have several smaller commits in your PR you should then also update the individual commits. This is however a bit of work. I had this issue recently when I had a PR with several small commits an issues in the first two or three. 

For this PR I think now the cleaner solution is that you squash them and provide one final commit.

cochicde:
> I appreciated that you original split this PR in smaller commits. However as you are now adding "repair" commits on top the idea of small commits and a clean history as we describe it in our [contributing guides](https://eclipse.dev/4diac/doc/development/contribute.html) is not followed.
> 
> For future references if you want to have several smaller commits in your PR you should then also update the individual commits. This is however a bit of work. I had this issue recently when I had a PR with several small commits an issues in the first two or three.
> 
> For this PR I think now the cleaner solution is that you squash them and provide one final commit.

Sorry for the mess. I'll follow the guidelines better the next time.

I changed to a switch statement, removed the undo/redo comment and  squashed all commits. I'll leave the keybindings for the next PRs.

azoitl (approved):
Thx for all the clean-up and your patience. LGTM now.

PR #2424, opened 2026-05-02
Upgraded XText Tests to Junit6

azoitl:
Not yet sure why it is not working. Maybe we need to wait for a newer Xtend maven plugin.


pom.xml
mx990, line 191: It works for me when specifying `junit6` as `providerHint`:
```xml
					<providerHint>junit6</providerHint>
```

mx990 (changes requested):

pom.xml
azoitl: thx for the feedback. this solved it on my machine.

azoitl:
It looks like github is now also messing with me. I just did a forcepush and the update is visible on my fork but not here.

azoitl:
It looks like Github had an outage. Repushing helped.

mx990 (approved):

PR #2430, opened 2026-05-05
Refactor edge direction

vikash1703:


azoitl:
@vikash1703 please always fetch always from origin (this repository here) and start your development from origin/develop. This should give you always a clean commit history. We also tried to describe this workflow in our contributing guidelines.

vikash1703:
> @vikash1703 please always fetch always from origin (this repository here) and start your development from origin/develop. This should give you always a clean commit history. We also tried to describe this workflow in our contributing guidelines.

sorry i always forgot i  will do it right now

azoitl (changes requested):
As part of this refactoring I expected that the new EdgeDirection will already be used by the ECC connection anchor so that we can see how it shall be used.

Also you need to consider if the state has actions. Then right should not be returned. 

In the end this class should replace all edge logic currently used in the anchor.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECStateConnectionAnchor.java
```diff
@@ -101,40 +86,33 @@ private Point applySpacing(final Rectangle bounds, final Edge edge) {
 		};
 	}
 
-	private List<ECTransition> getTransitionsOnEdgeOrdered(final Edge edge) {
+	private List<ECTransition> getTransitionsOnEdge(final EdgeDirection edge, final Point center) {
```
azoitl, line 89: why did you remove the `Ordered` from the name

vikash1703: > why did you remove the `Ordered` from the name

@azoitl 
"Hi Alois,
I updated this method signature mainly for two reasons:

We replaced the local Edge enum with the new standalone EdgeDirection utility class.

I added Point center as a parameter because the new EdgeDirection.of() method requires the center point of the state's bounds to correctly calculate the direction (Top, Bottom, Left, or Right).

I also dropped 'Ordered' from the method name just to keep it concise and clean, but if you prefer the original name getTransitionsOnEdgeOrdered, I am completely happy to revert the name back.

azoitl: thx for the explanation.I understood the change in parameters I just wanted to get more background on the name change. Fine for me.

azoitl (approved):

PR #2435, opened 2026-05-07
Fix stale type references after project dereference

oberlehner:
This fixes two related issues when dereferencing a project:

1. The target project's TypeLibrary kept stale TypeEntry references from the
   dereferenced project, which could create duplicate type markers.
2. Open system editors could still show/use the old TypeEntry because their
   dependency updater was only listening to the TypeEntry itself, not to the
   target TypeLibrary visibility change.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
+++ b/plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.model.ui.editors;
 
+import java.util.Collection;
 import java.util.Objects;
 
 import org.eclipse.emf.common.notify.Notification;
@@ -49,6 +50,7 @@ import org.eclipse.swt.widgets.Display;
 public class LibraryElementDependencyUpdater extends LibraryElementDependencyTracker {
 
 	private LibraryElement libraryElement;
+	private TypeLibrary typeLibrary;
 	private boolean updating;
 
 	@Override
@@ -67,8 +69,37 @@ public class LibraryElementDependencyUpdater extends LibraryElementDependencyTra
 							.asyncExec(() -> updateDependency(dependency, notification.getOldStringValue()));
 				}
 			}
+		} else if (notification.getNotifier() == typeLibrary
+				&& TypeLibrary.TYPE_ENTRY_NAME_REFERENCES_FEATURE.equals(notification.getFeature())) {
+			// react when a dependency is no longer visible through this library
+			updateTypeLibraryDependency(notification);
 		} else {
 			super.notifyChanged(notification);
+		}
+	}
+
+	private void updateTypeLibraryDependency(final Notification notification) {
+		switch (notification.getEventType()) {
+		case Notification.REMOVE, Notification.REMOVE_MANY ->
+			updateRemovedTypeEntryNameReference(notification.getOldValue());
+		default -> {
+			// ignore
+		}
+		}
+	}
+
+	private void updateRemovedTypeEntryNameReference(final Object object) {
+		switch (object) {
+		case final Collection<?> collection -> {
+			for (final Object value : collection) {
+				updateRemovedTypeEntryNameReference(value);
+			}
+		}
+		case final TypeEntry dependency when getDependencies().contains(dependency) ->
+			Display.getDefault().asyncExec(() -> updateDependency(dependency, dependency.getFullTypeName()));
+		case null, default -> {
+			// ignore
+		}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
```diff
@@ -67,11 +69,40 @@ public void notifyChanged(final Notification notification) {
...
+		case Notification.REMOVE, Notification.REMOVE_MANY ->
+			updateRemovedTypeEntryNameReference(notification.getOldValue());
+		default -> {
+			// ignore
+		}
+		}
+	}
+
+	private void updateRemovedTypeEntryNameReference(final Object object) {
+		switch (object) {
```
mx990: If you distinguish between `REMOVE` and `REMOVE_MANY` above, you can avoid this second switch.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -206,7 +206,7 @@ public DataTypeLibrary getDataTypeLibrary() {
 	}
 
 	public IProject getProject() {
-		return project;
+		return owningProject;
```
mx990: I believe we should keep the name of the field the same as its getter to avoid confusion. I would therefore suggest to keep the current name and rename only shadowing local variables.

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
```diff
@@ -230,12 +261,22 @@ private void install() {
 		if (libraryElement != null && !libraryElement.eAdapters().contains(this)) {
 			libraryElement.eAdapters().add(this);
 		}
+		if (libraryElement != null) {
+			typeLibrary = libraryElement.getTypeLibrary();
+			if (typeLibrary != null && !typeLibrary.eAdapters().contains(this)) {
+				typeLibrary.eAdapters().add(this);
+			}
+		}
```
mx990: I believe if you directly register for notifications from referenced type libraries in `addDependency` below, you should be able to avoid the transitive updates in the type libraries themselves.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
mx990: I believe this has to account for foreign type entries from referenced type libraries by checking if the type library is still referenced, as well as if the type entry exists and is still public in the *referenced* type library.

mx990, line 385: I believe this also needs to send notifications for added name references if the retry was successful.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
oberlehner: done

oberlehner: done

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
oberlehner: done

plugins/org.eclipse.fordiac.ide.model.ui/src/org/eclipse/fordiac/ide/model/ui/editors/LibraryElementDependencyUpdater.java
oberlehner: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -556,30 +586,89 @@ private TypeEntry getBlockTypeEntry(final TypeEntry entry) {
...
+	}
+
+	private boolean hasDataTypeEntryNameReference(final DataTypeEntry entry) {
+		final var dataTypes = dataTypeLib.getDerivedDataTypes().iterator();
+		while (dataTypes.hasNext()) {
+			if (dataTypes.next() == entry) {
+				return true;
+			}
+		}
+		return false;
```
mx990:
```suggestion
		return dataTypeLib.getDerivedTypeEntry(entry.getFullTypeName()) == entry;
```

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #2436, opened 2026-05-08
Remove FB palette resource listener on dispose

oberlehner:


plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBPaletteViewer.java
```diff
@@ -171,7 +174,16 @@ private void setupResourceChangeListener(final IProject project) {
 					commonViewer.refresh();
 				}
 			});
-		});
+		};
+		ResourcesPlugin.getWorkspace().addResourceChangeListener(resourceChangeListener);
+		getControl().addDisposeListener(event -> disposeResourceChangeListener());
```
azoitl: the base class already registers a dispose listener you can just override `handleDispose`

azoitl (changes requested):
I think this should be done on the 3.1.x branch

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBPaletteViewer.java
oberlehner: Are you sure ?
I think the dispose listener is registerd in org.eclipse.gef.ui.parts.AbstractEditPartViewer.hookControl(). 

But the FBPaletteviewer overrides hookControl as empty method.
I moved it to hookControl now

azoitl: No I'm not. To be more precise you are totally right and I'm stupid. As we are showing a CNF inside the palette viewer I had to do not so nice things to get that working.

plugins/org.eclipse.fordiac.ide.application/src/org/eclipse/fordiac/ide/application/editors/FBPaletteViewer.java
```diff
@@ -153,11 +156,20 @@ private void handleTreeExpansion(final String string) {
...
-		// do nothing here! Especially do not call super.hookControl!
+		// Do not call super.hookControl(); PaletteViewer expects a FigureCanvas here.
+		// Register the disposal path that the skipped base hook would install.
+		getControl().addDisposeListener(this::handleDispose);
+	}
+
+	@Override
+	protected void handleDispose(final DisposeEvent event) {
+		disposeResourceChangeListener();
+		super.handleDispose(event);
```
azoitl: I think this super call is as dangerous as the hook control

oberlehner: i removed it

azoitl (approved):

PR #2442, opened 2026-05-08
Added Control Engineering Function Blocks

mayrthom:
As there is currently no module for control engineering available, I implemented one starting with the most basic controllers: P, PI,  PID. Which are implemented in the PID Function Block.

The FB is implemented as Basic FB with a simple anti-windup strategy.

I would like to expand this in the future with additional control FBs if I find the time (thinking of LQR, Kalman, etc.).

### DIFF
--- a/data/typelibrary/controlengineering-3.0.0/MANIFEST.MF
+++ b/data/typelibrary/controlengineering-3.0.0/MANIFEST.MF
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="platform:/resource/org.eclipse.fordiac.ide.library.model/model/library.xsd" Scope="Library">
+    <Dependencies>
+        <Required SymbolicName="core" Version="3.0.0"/>
+    </Dependencies>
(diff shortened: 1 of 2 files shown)


diplfranzhoepfinger:
https://www.holobloc.com/doc/navigator.htm
https://www.holobloc.com/doc/fb/rt/events/E_PWM.htm

azoitl:
> https://www.holobloc.com/doc/navigator.htm https://www.holobloc.com/doc/fb/rt/events/E_PWM.htm

@diplfranzhoepfinger I assume with your pointers you wanted @mayrthom to also look at what FBDK/Holobloc is doing. I must point out from a license perspective we can not take anything from Holoblocs library. Also as @mayrthom nice documented at his other PR PWM has quite some tricks and the his proposed approach is better. Furthermore in this PR there is no PWM block so I have to ask what was the message of your post @diplfranzhoepfinger.

mayrthom:
Interesting is that holobloc put the PWM in the Event package.
But I'm still not sure which would be the correct package.

diplfranzhoepfinger:
> @diplfranzhoepfinger I assume with your pointers you wanted @mayrthom to also look at what FBDK/Holobloc is doing. I must point out from a license perspective we can not take anything from Holoblocs library. .... Furthermore in this PR there is no PWM block so I have to ask what was the message of your post @diplfranzhoepfinger.

there was one originally. 

understand the Licence Situation. clear. cannot be done.

PR #2448, opened 2026-05-11
Avoid export of generic types

MandKastner:


azoitl (approved):

plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
```diff
@@ -58,7 +58,9 @@ public class ExportBuilder extends IncrementalProjectBuilder {
 	public static final String BUILDER_ID = "org.eclipse.fordiac.ide.export.builder"; //$NON-NLS-1$
 	private static final String FORTE_NG_FILTER_ID = "org.eclipse.fordiac.ide.export.exportFilter.forteNg"; //$NON-NLS-1$
 
-	private static final Set<String> fileTypes = Set.of(TypeLibraryTags.ADAPTER_TYPE_FILE_ENDING,
+	static final String GENERIC_CLASS_NAME_ATTRIBUTE = "eclipse4diac::core::GenericClassName"; //$NON-NLS-1$
```
ernstblechaPT, line 61: Could the attribute-name be used from https://github.com/eclipse-4diac/4diac-ide/blob/8fa93be780048ae563e043b0551c58fa8f82f230/plugins/org.eclipse.fordiac.ide.export.forte_ng/src/org/eclipse/fordiac/ide/export/forte_ng/util/ForteNgExportUtil.xtend#L454 or is there a reason to decouple this from a dependency standpoint?

azoitl: Export filters should never directly referenced they are only accessed via their name and the extension point. So adding this dependency would couple it to tightly. The option that could be used is that `ForteNgExportUtil` should use the definition here or in a better location here in the base export plugin.

oberlehner (approved):

PR #2453, opened 2026-05-12
Changed BulkEditor Context Menu to use Eclipse infrastructure

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/handler/GoToElementHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.ui.ide.IDE;
+
+public class GoToElementHandler extends AbstractHandler {
+
+	@Override
+	public Object execute(final ExecutionEvent event) throws ExecutionException {
+		final var selection = HandlerUtil.getCurrentStructuredSelection(event);
+		if (selection.getFirstElement() instanceof final EObject rowObject) {
```
azoitl: in rare cases selection may be null

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/handler/GoToElementHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
+import org.eclipse.fordiac.ide.model.helpers.ModelHelper;
+import org.eclipse.fordiac.ide.model.ui.actions.OpenListenerManager;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.ui.ide.IDE;
+
+public class GoToElementHandler extends AbstractHandler {
+
```
azoitl, line 32: it would be good to add an enablement check method (see other handlers) as the find action (ctrl+3) will list all enabled commands/handlers.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.bulkeditor/src/org/eclipse/fordiac/ide/bulkeditor/handler/GoToElementHandler.java
azoitl: I think you misunderstood me. What I meant is that all handlers should also have a `public void setEnabled(final Object evaluationContext) ` method which will call `setBaseEnabled()` to tell the Eclipse Platform if this command is currently active or not.

azoitl (changes requested):

azoitl (approved):

PR #2456, opened 2026-05-13
Add array-to-array and array-to-values function blocks for BYTE and INT types

franz-hoepfinger-4diac:
Introduces more variants of the existing SI Generic Block to show users that this one is not limited to LREAL which one might think looking to the Library.


franz-hoepfinger-4diac:
i thought i made this PR long time ago, but i cannot find it anymore. 
so here is it.

azoitl:

PR #2358, opened 2026-04-01
Employ CodeQL checks for PR workflow

ruspl-afed:
Fixes #2357

PR #2461, opened 2026-05-14
Create and codeql.yml Workflow

azoitl:
Following @mx990 suggestion in PR #2358 I used githubs advanced codeql setup to create this codeql scanning workflow.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,101 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "develop" ]
+  pull_request:
+    branches: [ "develop" ]
+  schedule:
+    - cron: '40 5 * * 6'
+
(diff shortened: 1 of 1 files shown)


ruspl-afed:
Looks good. It would be great to have license header for this newly added manifest

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,101 @@
+# For most projects, this workflow file will not need changing; you simply need
```
ruspl-afed, line 1: Do we need all the comments in this file to be kept? May be a reference to origin would be enough

azoitl:
@ruspl-afed the file as I committed it is the default workflow generated by github. Therefore I didn't add a copyright header and I also didn't remove any comments. Especially for the review phase. I thin we should improve the file to check for all branches and also pin the versions of the workflows so that dependabot can manage them. WDYT?

mx990:
It looks to me like the autobuild crashes[^1] and the scanning may thus produce incomplete results. I believe we need to either add a manual build step to this workflow or add the CodeQL action to [.github/workflows/maven.yml](https://github.com/eclipse-4diac/4diac-ide/blob/develop/.github/workflows/maven.yml) as an additional step.

[^1]: see https://github.com/eclipse-4diac/4diac-ide/actions/runs/25868842845/job/76018231943#step:5:909

azoitl:
Integrated CodeQL checks into our standard maven workflow with PR #2463. I guess this is the better approach as a our Tycho/Maven build seems not to be handled well by CodeQL. I had a similar experience with dependabot which is also not updating our maven dependencies.

oberlehner (approved):
Good idea !

azoitl:
closing in favor of #2463

PR #2463, opened 2026-05-14
Integrated CodeQL checking into main maven workflow

azoitl:


.github/workflows/maven.yml
```diff
@@ -18,31 +18,59 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
         java: [21]
+        include:
+        - language: actions
+          build-mode: none
+        - language: java-kotlin
+          build-mode: manual
```
mx990: I believe we should only integrate the `java-kotlin` language into our maven workflow and leave the `actions` separate, otherwise we will unnecessarily run our maven build twice. I would suggest directly specifying the `languages` and `build-mode` for the CodeQL actions below.

mx990 (changes requested):

.github/workflows/maven.yml
azoitl: good point. done

mx990 (approved):

oberlehner (approved):

PR #2476, opened 2026-05-16
Fix: correct input reference for WORD_01 in SPLIT_DWORD_INTO_WORDS

franz-hoepfinger-4diac:
The input reference for WORD_01 was mistakenly using IN.%X1 instead of IN.%W1. This update corrects that reference, aligning it with the intended usage.


azoitl:

PR #2479, opened 2026-05-17
Added JSON Parser FBs

mayrthom:
Created a JSON Encoder and Decoder, useful for connecting 4diac to Node-RED. 
The JSON Encoder is modular and can be connected in series for multiple JSON outputs.
Related to eclipse-4diac/4diac-forte#921 

Screenshot


azoitl:
This is an interesting approach. Have you also considered to implement it al communication layer and provide the field names as part of the communication layer parameters? This would reduce the number of required FBs in an application. However with the drawback of being less flexible.

mayrthom:
I havn't thought of implementing it as ComLayer before as it started as spontaneous idea for trying to connect 4diac with Node-RED.


A Json Parser as ComLayer would be very eligant, however I'm not that deep in the ComLayer yet and i just wanted to create a simple versatile approach that creates a Json string.

azoitl:
I like this approach as well as it has other properties that make it interesting. Would you be interested doing a layer in parallel? I think the raw layer can serve as good starting point.

Regarding this PR. Please rebase your commit on the latest head of develop to have a clean history. Also I think this FBs would better fit into the convert package, or?

mayrthom:
I can look into it when I have some time, but I can't promise it now.

I rebased the PR and changed it to the convert package.

azoitl (changes requested):
Please add a copyright header as part of the description field. You can take a look at other FBs in the lib for reference.

azoitl:
> I can look into it when I have some time, but I can't promise it now.

I think it would be great but no pressure.

mayrthom:
Updated it.

azoitl (approved):

PR #2485, opened 2026-05-18
Clear TypeLibrary after project close

oberlehner:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibrary.java
```diff
@@ -608,6 +608,32 @@ private boolean removeReferencedProject(final IProject referencedProject) {
...
+		dataTypeLib.clear();
+		fileMap.clear();
+		packages.clear();
+		duplicates.clear();
+		referencedProjects.clear();
+		eAdapters().clear();
+	}
+
+	private void disposeReferencedProjects() {
+		// TODO clarify behavior when disposing referenced projects
```
mx990: I believe we should remove ourselves from the adapters of any referenced projects here. The type libraries of any referenced projects themselves should remain untouched.

oberlehner: done

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/typelibrary/TypeLibraryManager.java
```diff
@@ -67,6 +67,12 @@ public TypeLibrary getTypeLibrary(final IProject proj) {
 		}
 	}
 
+	public TypeLibrary getTypeLibraryFromProject(final IProject proj) {
```
mx990: I would suggest renaming this to `getTypeLibraryIfPresent`, since we already have a `getTypeLibrary(IProject)`.

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #2486, opened 2026-05-19
Fix new project target selection combobox

oberlehner:
When the new 4diac project wizard is opened in an empty workspace, the
library import page could not preselect the project because it did not
exist yet.

Allow the new project wizard to pass the future project handle to the
library import page while keeping the regular import wizard behavior
unchanged.


azoitl (changes requested):
With this change I can not press finish on the initial page of the new project wizard. This feels not good.

azoitl (approved):

PR #2495, opened 2026-05-23
Ecc new router fix

vikash1703:


vikash1703:
@azoitl 

i will update by tomorrow as i also faced problem everything is now working fine but whgat i got straigth line issue at first time our updatebendpoint throwing null. what i think but i will take tomorrow and fix what is happening where i am wrong

azoitl:
@vikash1703 should I already test or still wait?

vikash1703:
> @vikash1703 should I already test or still wait?

yes you can test and you will feel flicker when dragging transition and straigh line missing and rest of all are good

vikash1703:
> > @vikash1703 should I already test or still wait?
> 
> yes you can test and you will feel flicker when dragging transition and straigh line missing and rest of all are good

and what i think if we somehow do for straigth line and at first user will create small transition straigth and after second time it changes to s  shape then we still got curve but at that time it much better.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECCTransitionRouter.java
```diff
@@ -0,0 +1,130 @@
...
+	public void route(final Connection conn) {
+
+		if (conn.getSourceAnchor() == null || conn.getTargetAnchor() == null) {
+			super.route(conn);
+			return;
+		}
+		final Object constraint = getConstraint(conn);
+		if (!(constraint instanceof final List<?> bendpoints) || bendpoints.isEmpty()
+				|| !(bendpoints.get(0) instanceof final Bendpoint bp)) {
+			super.route(conn);
```
azoitl: This situations happens when layouting is done during dragging bendpoints. If you remove the super.route() the flickering and straight line problem is gone.

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/figures/ECCTransitionRouter.java
```diff
@@ -0,0 +1,130 @@
...
+
+	private static final double CTRL_POINT_FACTOR = 0.3;
+	private static final double MIN_LENGTH = 1.0;
+	private static final double EPSILON = 0.001;
+	private static final double MAX_HANDLE_DISTANCE = 150.0;
+
+	@Override
+	public void route(final Connection conn) {
+
+		if (conn.getSourceAnchor() == null || conn.getTargetAnchor() == null) {
```
azoitl: not sure if this can ever be the case that source or target anchor are null. Also the bendpointrouter from Eclipse GEF Classic is not doing this check. So you could remove it.

However you need to check if the connection is not a ECTransitionFigure. This happens during creating new connections. In this case you should for now call super.route.

azoitl (commented):
All in all it looks good. As I was curios I played a bit with your code and I think I found the reason for the flickering and straight line problems. See comments below.

oberlehner (approved):

azoitl (approved):
Looks and feels amazing!

PR #2498, opened 2026-05-24
Adds workflow which marks inactive issues and PRs as stale

MartinMelikMerkumians:
Adds a workflow to mark inactive issues and PRs as stale and on continued inactivity close them.

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -1,0 +1,28 @@
+---
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+permissions: {}
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mx990 (approved):

azoitl (commented):
Should we add an exception for issues/prs with a bug label as discussed at 4diac FORTE?

mx990:
> Should we add an exception for issues/prs with a bug label as discussed at 4diac FORTE?

How about adding a new label `backlog` or something similar to explicitly mark long-term issues/prs?

MartinMelikMerkumians:
Sounds like a good idea, I wil check and add this to the PR accordingly.

mx990 (approved):

azoitl (approved):
I find that also a good idea.

PR #2500, opened 2026-05-25
Cleaned Instance Property Sheet

azoitl:
The layout of type interface sheet is cleaned and the untyped subapp sheet and the instance property sheet are better seperated.


sebHollersbacher (commented):
Is it intentional that the Data Input/Output tables are removed from the UntypedSubAppPropertySection?

azoitl:
> Is it intentional that the Data Input/Output tables are removed from the UntypedSubAppPropertySection?

Yes. In the next PR new tables are added that at this point work only for typed instances and for untyped subapps there are anyhow the child property sheets that are better suited for editing of untyped subapp interfaces.

sebHollersbacher (approved):

oberlehner (approved):

PR #2510, opened 2026-05-29
Fix dependency update for renamed type entries

oberlehner:
Use the new full type name when updating dependencies after a TypeEntry rename.
Using the old name during a data type move can create a transient missing
datatype and reconfigure configurable FBs with an error type, dropping member
access pins.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
@@ -353,7 +353,7 @@ public abstract class AbstractUpdateBlockFBNElementCommand extends Command
 
 	private void transferInstanceComments() {
 		oldElement.getInterface().getAllInterfaceElements().filter(ie -> !ie.getComment().isBlank()).forEach(ie -> {
-			final IInterfaceElement newIE = newElement.getInterface().getInterfaceElement(ie);
+			final IInterfaceElement newIE = getNewInterfaceElementForPreservingInstanceData(ie);
 			if (newIE != null) {
 				newIE.setComment(ie.getComment());
 			}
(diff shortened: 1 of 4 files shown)


mx990 (changes requested):
As discussed previously in https://github.com/eclipse-4diac/4diac-ide/pull/2304, this would break dependency updates for types renamed outside of a refactoring, where we explicitly *want* an error type to be created.

I believe we need a different approach to retain member access pins. They should also survive if the struct is missing temporarily, for example, when deleting and later restoring the struct. How about creating a `ContainerVarDeclaration` also for error data types in `VarDeclarationFactory.createVarDecl`?

mx990:
I have created PR https://github.com/eclipse-4diac/4diac-ide/pull/2511, which adds a comment explaining why the old type name is used.

azoitl:
> 
> I believe we need a different approach to retain member access pins. They should also survive if the struct is missing temporarily, for example, when deleting and later restoring the struct. How about creating a `ContainerVarDeclaration` also for error data types in `VarDeclarationFactory.createVarDecl`?

I like that idea. I also think we should use `VarDeclarationFactory.createVarDecl` in more places. I haven't yet found the time to search for all `VarDeclaration` creation points where we should use the factory instead.

oberlehner:
> As discussed previously in #2304, this would break dependency updates for types renamed outside of a refactoring, where we explicitly _want_ an error type to be created.
> 
> I believe we need a different approach to retain member access pins. They should also survive if the struct is missing temporarily, for example, when deleting and later restoring the struct. How about creating a `ContainerVarDeclaration` also for error data types in `VarDeclarationFactory.createVarDecl`?

Thx for the hint, this solution makes more sense

oberlehner:
> I have created PR #2511, which adds a comment explaining why the old type name is used.

thx, this makes it much clearer now.

oberlehner:
> > I believe we need a different approach to retain member access pins. They should also survive if the struct is missing temporarily, for example, when deleting and later restoring the struct. How about creating a `ContainerVarDeclaration` also for error data types in `VarDeclarationFactory.createVarDecl`?
> 
> I like that idea. I also think we should use `VarDeclarationFactory.createVarDecl` in more places. I haven't yet found the time to search for all `VarDeclaration` creation points where we should use the factory instead.

you are right, this will be part of a future PR!

Felix-Schmid (commented):
LGTM

Felix-Schmid (approved):

azoitl (approved):

mx990 (approved):

PR #2511, opened 2026-05-29
Add comment explaining usage of old name in dependency updater

mx990:


azoitl (commented):
as we are in feature freeze phase and stabilizing refactoring there I think it would be good to have this PR an 3.2.x.

mx990:
> as we are in feature freeze phase and stabilizing refactoring there I think it would be good to have this PR an 3.2.x.

makes sense, done

azoitl (approved):

PR #2273, opened 2026-03-18
Add typemanagement refactoring test bundle scaffold

retryoos:
- What: New org.eclipse.fordiac.ide.typemanagement.tests test plugin with JUnit 5 and a disabled prototype for rename refactoring tests.
- Why: Provides a dedicated place for upcoming automated tests for org.eclipse.fordiac.ide.typemanagement.refactoring (rename/move/copy/delete).
- Next steps: Add fixture projects and executable tests for STRUCT datatype rename and FB type rename (including undo/redo + marker checks).

PR #2514, opened 2026-05-30
Fix undo of struct type rename leaving .dtp content stale

### LINKED ISSUE #2513: Undo of struct type rename does not revert the .dtp file content
## Summary
Undoing a struct type rename refactoring renames the `.dtp` file back to its original name, but the file's internal `<DataType Name=…>` is not reverted. The persisted file is left inconsistent and the original qualified type name becomes unresolvable.

## Steps to reproduce
1. Have `mypackage::InnerStruct` (file `InnerStruct.dtp`) referenced by `mypackage::OuterStruct`.
2. Rename `InnerStruct` → `InnerStructRenamed` via the type rename refactoring.
3. Perform the undo.

## Expected
- File on disk: `InnerStruct.dtp` with `<DataType Name="InnerStruct">`.
- `mypackage::InnerStruct` resolves and `OuterStruct`'s member reference is valid.

## Actual
- File on disk: `InnerStruct.dtp`, but its content still says `<DataType Name="InnerStructRenamed">`.
- A fresh type-library reload registers `mypackage::InnerStructRenamed` (from a file literally named `InnerStruct.dtp`).
- `OuterStruct`'s member resolves to an `ErrorDataType`.

## Root cause (analysis)
`UpdateTypeEntryChange#perform` returns its undo as `new UpdateTypeEntryChange(newFile, typeEntry, oldName, newName)`. When that undo runs inside the composite change, `findNewResource(this.newName)` looks for the file at the restored name, but at that point in the composite undo the file-rename undo has not yet executed, so the lookup returns `null`, `updateTypeEntryByRename` is skipped, and the content + library revert never happens.

## Reproducer
PR #2273 contains two `@Disabled` tests that demonstrate this:
- `StructDataTypeRenameTest.renameInnerStruct_undoRestoresOriginalState`
- `StructDataTypeRenameTest.renameInnerStruct_redoReappliesRename`

retryoos:
## Summary
Fixes #2513.

`UpdateTypeEntryChange#perform` returned an undo change whose `findNewResource(restoredName)` looked up the file at its restored name. During the LTK composite undo, the file rename undo has not yet executed when this change runs, so the lookup returned `null` and the content and library revert were silently skipped. The file was renamed back on disk, but its internal `<DataType Name=...>` kept the renamed value, leaving the persisted model inconsistent and the original qualified type name unresolvable.

## Fix
- Use the captured file reference when it still resolves (the undo case, where the file is still at its renamed location); fall back to `findNewResource` only on the forward path.
- Derive the target type name from the constructor argument instead of from the file's current filename, so the content update writes the correct name regardless of where the file is currently located.

No public API change. ~29 net lines in one file.

## Reproducer
The two undo/redo tests in #2273 demonstrate the bug. They are currently `@Disabled` there; once this lands and is back-merged into `develop`, they will be re-enabled in that PR.

## Verification
Local Tycho build on `3.2.x` with the reproducer tests pulled in and `@Disabled` removed: `Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/UpdateTypeEntryChange.java
+++ b/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/UpdateTypeEntryChange.java
@@ -99,28 +99,30 @@ public class UpdateTypeEntryChange extends Change {
 
 	@Override
 	public Change perform(final IProgressMonitor pm) throws CoreException {
-		final IFile newFile = findNewResource(newName);
-		if (newFile != null && typeEntry != null) {
-			updateTypeEntryByRename(newFile, typeEntry);
-			return new UpdateTypeEntryChange(newFile, typeEntry, oldName, newName);
+		// On the composite undo, the file rename undo has not run yet, so the file is
+		// still at its renamed location; prefer the captured reference when it still
+		// resolves and fall back to the name lookup for the forward path only.
+		final IFile currentFile = file.exists() ? file : findNewResource(newName);
+		if (currentFile != null && typeEntry != null) {
+			updateTypeEntryToName(currentFile, typeEntry, newName);
+			return new UpdateTypeEntryChange(currentFile, typeEntry, oldName, newName);
 		}
 		return null;
 	}
 
-	private static void updateTypeEntryByRename(final IFile newFile, final TypeEntry entry) {
-		final String newTypeName = TypeEntry.getTypeNameFromFile(newFile);
-
-		if (!Objects.equals(newFile, entry.getFile())) {
+	private static void updateTypeEntryToName(final IFile currentFile, final TypeEntry entry,
+			final String targetTypeName) {
+		if (!Objects.equals(currentFile, entry.getFile()) || !Objects.equals(targetTypeName, entry.getTypeName())) {
 			final TypeLibrary typeLibrary = entry.getTypeLibrary();
 			if (typeLibrary != null) {
 				typeLibrary.removeTypeEntry(entry);
 			}
-			entry.setFile(newFile);
+			entry.setFile(currentFile);
 
 			// update type and typeEditable names
 			final LibraryElement type = entry.copyType();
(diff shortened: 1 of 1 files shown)


oberlehner:
Please use the correct code formater.
 Here we have a lot of whitespaces which produces uncessarry diffs.

retryoos:
> Please use the correct code formater. Here we have a lot of whitespaces which produces uncessarry diffs.

Thank you for the feedback Michael, I kept the original guard shape and just inverted the condition, so the body stays at its original indentation, now diff is down to about 10 lines, so no formatting noise

oberlehner:
LGTM
Please rebase to develop and change the base !

oberlehner:
@retryoos please always assign a reviewer !

mx990:
> Please rebase to develop and change the base !

I would consider this an important fix, so I had asked to base the PR on `3.2.0` in https://github.com/eclipse-4diac/4diac-ide/pull/2273#pullrequestreview-4392614482.

mx990 (approved):

retryoos:
I rebased onto develop and changed the base with the same commit and diff, thank you for the guidance, happy to follow whatever the is decided there

oberlehner:
> > Please rebase to develop and change the base !
> 
> I would consider this an important fix, so I had asked to base the PR on `3.2.0` in [#2273 (review)](https://github.com/eclipse-4diac/4diac-ide/pull/2273#pullrequestreview-4392614482).

Ok then, please change back to 3.2.0 . sry for the circumstances

retryoos:
> > > Please rebase to develop and change the base !
> > 
> > 
> > I would consider this an important fix, so I had asked to base the PR on `3.2.0` in [#2273 (review)](https://github.com/eclipse-4diac/4diac-ide/pull/2273#pullrequestreview-4392614482).
> 
> Ok then, please change back to 3.2.0 . sry for the circumstances

Done, rebased back onto 3.2.x and switched the base. Same commit and diff, thank you for the coordination!

mx990 (approved):
Thank you for providing the fix and your patience. I am also sorry for the confusion.

oberlehner (approved):

PR #2516, opened 2026-05-31
Add cascade tests for struct rename in FB interface and configurable FB

retryoos:
## Summary
Extends the StructRenameTest fixture from #2273 to exercise two forward cascade paths of the struct type rename refactoring that were not covered before:

- **`DataTypeEdit` cascade into a referencing FB type definition's interface variable**, `StructProducer.OUT` (declared as `mypackage::InnerStruct`) updates to `mypackage::InnerStructRenamed` after the rename.
- **`UpdateConfigurableFBModelEdit` cascade into a STRUCT_DEMUX instance's configuration**, the `StructuredType` attribute on a `STRUCT_DEMUX` FB instance in a system Application updates accordingly.

## Files
- `data/StructRenameTest/Type Library/StructProducer.fbt`, service-interface FB with one output `OUT : mypackage::InnerStruct`
- `data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt`, verbatim copy from the `HiddenConnectionTest` fixture (matches the existing test convention of bundling `STRUCT_DEMUX` per fixture)
- `data/StructRenameTest/StructRenameTest.sys`, Application named `App` with a `Producer` FB instance and a `Demux` FB instance configured for `mypackage::InnerStruct`
- `src/.../StructRenameSystemCascadeTest.java`, 2 `@Test` methods following the same per-test workspace-copy isolation as `StructDataTypeRenameTest`

No production code change and no public API change.

## Verification
Local Tycho run, all bundles in the typemanagement tests reactor:
```
TypeManagementRefactoringSmokeTest          1/1 passed
StructDataTypeRenameTest                    5/5 (3 passed, 2 @Disabled, undo behavior tracked in #2513 / #2514)
StructRenameSystemCascadeTest               2/2 passed
BUILD SUCCESS, 0 Checkstyle violations across 17 modules
```

## Relationship to other PRs
Builds on top of #2273 (the test bundle scaffold). I will rebase this branch onto the merged form of #2273 once that lands, so the diff against `develop` shrinks to just the four new files in this PR.

### DIFF
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
@@ -20,26 +20,9 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.m2e.core.maven2Builder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
-		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 		<nature>org.eclipse.pde.PluginNature</nature>
 		<nature>org.eclipse.jdt.core.javanature</nature>
 	</natures>
-	<filteredResources>
-		<filter>
-			<id>1773845733336</id>
-			<name></name>
-			<type>30</type>
-			<matcher>
-				<id>org.eclipse.core.resources.regexFilterMatcher</id>
-				<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
-			</matcher>
-		</filter>
-	</filteredResources>
 </projectDescription>
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
@@ -20,6 +20,7 @@ import java.nio.file.StandardCopyOption;
 import java.util.Comparator;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.resources.IResource;
@@ -30,7 +31,9 @@ import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.fordiac.ide.library.LibraryManager;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
 import org.eclipse.ltk.core.refactoring.Change;
 import org.eclipse.ltk.core.refactoring.CheckConditionsOperation;
 import org.eclipse.ltk.core.refactoring.CreateChangeOperation;
@@ -70,6 +73,36 @@ public final class RefactoringTestSupport {
 		project.open(new NullProgressMonitor());
 		project.refreshLocal(IResource.DEPTH_INFINITE, new NullProgressMonitor());
 		return project;
+	}
+
+	/**
+	 * Link the given standard libraries from the source tree so types they declare
+	 * resolve at test runtime instead of being copied into the fixture.
+	 */
+	public static void linkStandardLibraries(final IProject project, final String... libraryNames) throws CoreException {
+		ensureLibraryFolders(project);
+		final java.nio.file.Path standardLibraries = java.nio.file.Path
+				.of(System.getProperty("user.dir"), "..", "..", "data", "typelibrary") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+				.toAbsolutePath().normalize();
+		for (final String libraryName : libraryNames) {
+			LibraryManager.INSTANCE.importLibrary(project, standardLibraries.resolve(libraryName).toUri(), true, false);
+		}
+		// refresh() only reconciles added or deleted files; drop the cached
+		// TypeLibrary so .sys entries re-parse with the linked libraries in scope.
+		TypeLibraryManager.INSTANCE.removeProject(project);
+		TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
+	}
+
+	private static void ensureLibraryFolders(final IProject project) throws CoreException {
+		final NullProgressMonitor monitor = new NullProgressMonitor();
+		final IFolder standardLibs = project.getFolder(TypeLibraryTags.STANDARD_LIB_FOLDER_NAME);
+		if (!standardLibs.exists()) {
+			standardLibs.create(IResource.VIRTUAL | IResource.FORCE, true, monitor);
+		}
+		final IFolder externalLibs = project.getFolder(TypeLibraryTags.EXTERNAL_LIB_FOLDER_NAME);
+		if (!externalLibs.exists()) {
+			externalLibs.create(IResource.VIRTUAL | IResource.FORCE, true, monitor);
+		}
 	}
 
 	/** Remove a project from the workspace and drop its cached type library. */
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
@@ -55,13 +55,13 @@ public final class RefactoringTestSupport {
 	 * refactorings rename and rewrite type files, so each test must operate on its
 	 * own writable copy instead of mutating the committed fixture in place.
 	 */
-	public static IProject importProjectIntoWorkspace(final String projectName, final Bundle bundle,
-			final IPath bundleRelativePath) throws CoreException, IOException {
+	public static IProject importProjectIntoWorkspace(final String projectName, final String bundleRelativePath)
+			throws CoreException, IOException {
 		final IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
 		final IProject project = root.getProject(projectName);
 		deleteProject(project);
 
-		final java.nio.file.Path source = resolveBundleDirectory(bundle, bundleRelativePath);
+		final java.nio.file.Path source = resolveBundleDirectory(bundleRelativePath);
 		final java.nio.file.Path destination = root.getLocation().append(projectName).toFile().toPath();
 		deleteRecursively(destination);
 		copyRecursively(source, destination);
@@ -83,7 +83,7 @@ public final class RefactoringTestSupport {
 	public static void linkStandardLibraries(final IProject project, final String... libraryNames)
 			throws CoreException, IOException {
 		ensureLibraryFolders(project);
-		final java.nio.file.Path standardLibraries = resolveStandardLibrariesPath();
+		final java.nio.file.Path standardLibraries = TestRepositoryPaths.resolve().standardLibraries();
 		for (final String libraryName : libraryNames) {
 			LibraryManager.INSTANCE.importLibrary(project, standardLibraries.resolve(libraryName).toUri(), true, false);
 		}
@@ -91,16 +91,6 @@ public final class RefactoringTestSupport {
 		// TypeLibrary so .sys entries re-parse with the linked libraries in scope.
 		TypeLibraryManager.INSTANCE.removeProject(project);
 		TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
-	}
-
-	private static java.nio.file.Path resolveStandardLibrariesPath() throws IOException {
-		// Resolve the test bundle's source location through OSGi and walk up to
-		// the repository root, where the standard libraries live under
-		// data/typelibrary.
-		final Bundle bundle = FrameworkUtil.getBundle(RefactoringTestSupport.class);
-		final java.net.URL bundleRootUrl = FileLocator.toFileURL(bundle.getEntry("/")); //$NON-NLS-1$
-		return Paths.get(bundleRootUrl.getPath()).getParent().getParent().resolve("data").resolve("typelibrary") //$NON-NLS-1$ //$NON-NLS-2$
-				.toAbsolutePath().normalize();
 	}
 
 	private static void ensureLibraryFolders(final IProject project) throws CoreException {
@@ -146,9 +136,9 @@ public final class RefactoringTestSupport {
 		return perform.getUndoChange();
 	}
 
-	private static java.nio.file.Path resolveBundleDirectory(final Bundle bundle, final IPath bundleRelativePath)
-			throws IOException {
-		final var url = FileLocator.toFileURL(FileLocator.find(bundle, bundleRelativePath));
+	private static java.nio.file.Path resolveBundleDirectory(final String bundleRelativePath) throws IOException {
+		final Bundle bundle = FrameworkUtil.getBundle(RefactoringTestSupport.class);
+		final var url = FileLocator.toFileURL(FileLocator.find(bundle, new Path(bundleRelativePath)));
 		return Paths.get(url.getPath());
 	}
 
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
@@ -43,6 +43,7 @@ import org.eclipse.ltk.core.refactoring.RefactoringCore;
 import org.eclipse.ltk.core.refactoring.RefactoringStatus;
 import org.eclipse.ltk.core.refactoring.resource.RenameResourceDescriptor;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
 
 public final class RefactoringTestSupport {
 
@@ -79,11 +80,10 @@ public final class RefactoringTestSupport {
 	 * Link the given standard libraries from the source tree so types they declare
 	 * resolve at test runtime instead of being copied into the fixture.
 	 */
-	public static void linkStandardLibraries(final IProject project, final String... libraryNames) throws CoreException {
+	public static void linkStandardLibraries(final IProject project, final String... libraryNames)
+			throws CoreException, IOException {
 		ensureLibraryFolders(project);
-		final java.nio.file.Path standardLibraries = java.nio.file.Path
-				.of(System.getProperty("user.dir"), "..", "..", "data", "typelibrary") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
-				.toAbsolutePath().normalize();
+		final java.nio.file.Path standardLibraries = resolveStandardLibrariesPath();
 		for (final String libraryName : libraryNames) {
 			LibraryManager.INSTANCE.importLibrary(project, standardLibraries.resolve(libraryName).toUri(), true, false);
 		}
@@ -91,6 +91,16 @@ public final class RefactoringTestSupport {
 		// TypeLibrary so .sys entries re-parse with the linked libraries in scope.
 		TypeLibraryManager.INSTANCE.removeProject(project);
 		TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
+	}
+
+	private static java.nio.file.Path resolveStandardLibrariesPath() throws IOException {
+		// Resolve the test bundle's source location through OSGi and walk up to
+		// the repository root, where the standard libraries live under
+		// data/typelibrary.
+		final Bundle bundle = FrameworkUtil.getBundle(RefactoringTestSupport.class);
+		final java.net.URL bundleRootUrl = FileLocator.toFileURL(bundle.getEntry("/")); //$NON-NLS-1$
+		return Paths.get(bundleRootUrl.getPath()).getParent().getParent().resolve("data").resolve("typelibrary") //$NON-NLS-1$ //$NON-NLS-2$
+				.toAbsolutePath().normalize();
 	}
 
 	private static void ensureLibraryFolders(final IProject project) throws CoreException {
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeRenameTest.java
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeRenameTest.java
@@ -12,6 +12,15 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.typemanagement.tests;
 
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.CONVERT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.CORE;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.IEC_61131_3;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.INNER_STRUCT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.INNER_STRUCT_FILE;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.INNER_STRUCT_RENAMED;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.OUTER_STRUCT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.PROJECT_NAME;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.PROJECT_PATH;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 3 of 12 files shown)


tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Copy a bundle-relative project template into a fresh, writable project in the
+	 * test workspace.
+	 *
+	 * The fixture data shipped with the test bundle is read-only template data;
+	 * refactorings rename and rewrite type files, so each test must operate on its
+	 * own writable copy instead of mutating the committed fixture in place.
+	 */
+	public static IProject importProjectIntoWorkspace(final String projectName, final Bundle bundle,
```
oberlehner: please use the `org.eclipse.fordiac.ide.test.model.FordiacProjectLoader`

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/TypeManagementRefactoringSmokeTest.java
```diff
@@ -0,0 +1,25 @@
...
+ * Contributors:
+ *   Dimitrios Kalligaridis - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.typemanagement.tests;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+class TypeManagementRefactoringSmokeTest {
```
oberlehner: We don't need this class

tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
```diff
@@ -0,0 +1,45 @@
...
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
```
oberlehner: why are you using a maven project nature. please try to keep that as the other proejcts. for example -> /org.eclipse.fordiac.ide.test.library



```
<?xml version="1.0" encoding="UTF-8"?>
<projectDescription>
	<name>org.eclipse.fordiac.ide.typemanagement.tests</name>
	<comment></comment>
	<projects>
	</projects>
	<buildSpec>
		<buildCommand>
			<name>org.eclipse.jdt.core.javabuilder</name>
			<arguments>
			</arguments>
		</buildCommand>
		<buildCommand>
			<name>org.eclipse.pde.ManifestBuilder</name>
			<arguments>
			</arguments>
		</buildCommand>
		<buildCommand>
			<name>org.eclipse.pde.SchemaBuilder</name>
			<arguments>
			</arguments>
		</buildCommand>
		<buildCommand>
			<name>org.eclipse.m2e.core.maven2Builder</name>
			<arguments>
			</arguments>
		</buildCommand>
	</buildSpec>
	<natures>
		<nature>org.eclipse.m2e.core.maven2Nature</nature>
		<nature>org.eclipse.pde.PluginNature</nature>
		<nature>org.eclipse.jdt.core.javanature</nature>
	</natures>
	<filteredResources>
		<filter>
			<id>1773845733336</id>
			<name></name>
			<type>30</type>
			<matcher>
				<id>org.eclipse.core.resources.regexFilterMatcher</id>
				<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
			</matcher>
		</filter>
	</filteredResources>
</projectDescription>
```

tests/org.eclipse.fordiac.ide.typemanagement.tests/build.properties
oberlehner, line 1: Please also use the same build.properties as the other plugins.
e.g.:

```
pom.model.packaging = eclipse-test-plugin
source.. = src/
output.. = bin/
bin.includes = META-INF/,\
               data/,\
               .,\
               plugin.xml
```

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FBType Name="STRUCT_DEMUX" Comment="Generic block for demuxing structs">
```
oberlehner: Please dont copy paste Library blocks here.
You can use

oberlehner: You should import them with the library manager: 

Maybe this can work:
```
LibraryManager.INSTANCE.importLibrary(
		project,
		Paths.get("data/typelibrary/convert-1.0.0").toAbsolutePath().toUri(),
		true,
		true);
```

tests/org.eclipse.fordiac.ide.typemanagement.tests/.settings/org.eclipse.jdt.core.prefs
oberlehner, line 1: Please also check if the settings are equal to other 4diac projects. Here we have settings, like code formatter, etc.

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
retryoos: Thank you @oberlehner for pointing this out, the reason I avoided FordiacProjectLoader here is that it links the workspace project to the bundle path, so the rename refactoring ends up mutating the committed fixture in the repo on every test run. Would you prefer I add a writable copy variant to FordiacProjectLoader which copies the bundle data into the workspace location first, so refactoring tests can use the same convention without touching the committed fixture? or do you have another approach in mind? Thank you in advance for your feedback

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/TypeManagementRefactoringSmokeTest.java
retryoos: Removed in 6ecd61c

tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
retryoos: I compared this against test.library/.project and the structure is already identical aparat from the project name and the filter's autogen id, did you have a specific change in mind that I might be missing please lmk

tests/org.eclipse.fordiac.ide.typemanagement.tests/build.properties
retryoos: The pasted template includes plugin.xml because test.library ships one for an extension contribution, but this bundle has no extensions and adding it without the file fails Tycho with "bin.includes value [plugin.xml] do not match any files", so the current file follows the test.model pattern instead, happy to add a real plugin.xml if you would prefer.

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt
retryoos: Thank you I tried this approach but the path data/typelibrary/convert-1.0.0 does not resolve at test runtime, because Tycho's working dir is the test bundle's folder rather than the repo root, so importLibrary returns false and STRUCT_DEMUX never gets registered. What is the way to point at the standard convert library from test? I saw LibraryImportTest bundles its libraries as zips inside the test bundle and extracts them at setup, should I follow this convention?

tests/org.eclipse.fordiac.ide.typemanagement.tests/.settings/org.eclipse.jdt.core.prefs
retryoos: Aligned with test.library's jdt.core.prefs in 6ecd61c

tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
oberlehner: Why do you need the id ?
Please dont add maven project nature here : `<nature>org.eclipse.m2e.core.maven2Nature</nature>`

tests/org.eclipse.fordiac.ide.typemanagement.tests/build.properties
oberlehner: sry this was a bad example. It looks ok now !

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt
oberlehner: I think you first have to create a manifest in your test project.
Try to create the testStructRename project in a deployed 4diac IDE first, and then copy the manifest. 
Then you can use the LibraryManager to resolve the dependency.

Maybe this code snippet helps you:

```

final var convert = LibraryManager.INSTANCE.getStandardLibraries()
		.get("convert").stream()
		.filter(r -> r.version().toString().equals("3.0.0"))
		.findFirst()
		.orElseThrow();

LibraryManager.INSTANCE.importLibrary(project, convert.uri(), true, false);

LibraryManager.INSTANCE.resolveDependencies(project, ManifestHelper.getContainerManifest(project), null);

TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
```

Please use library version 3.0.0 ! 
Do not forget that convert 3.0.0 requires core 3.0.0 ! 

If you manage to load this libraries dynamically in your test it would be great to move the loading mechanisms to `RefactoringTestSupport `

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
oberlehner: Please adress `https://github.com/eclipse-4diac/4diac-ide/pull/2516#discussion_r3361290169` , we need to consider the linking of libraries too when loading a project !

tests/org.eclipse.fordiac.ide.typemanagement.tests/.project
retryoos: Oh you are right, I am sorry, I dropped the m2e.core.maven2Nature and the matching maven2Builder and removed the filteredResources block with the auto gen id, the file is now just the PDE /JDT natures and builders.

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt
retryoos: I generated the project MANIFEST.MF from 4diac ide with convert 3.0.0 and core 3.0.0 and shipped it in the fixture. I then implemented your snippet as RefactoringTestSupport.linkStandardLibraries(IProject) and the blocker for actually calling it is that in the Tycho test runtime, LibraryManager.getStandardLibraries() returns an empty map. Platform.getInstallLocation() resolves to target/work, which has no typelibrary/ folder, so the standard-library scan finds nothing and getStandardLibraries().get("convert") is null. The helper is in place, but what's the way to provision the standard libraries for the Tycho test runtime so I can write it in and drop the bundled STRUCT_DEMUX.fbt?

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
retryoos: The library loading above is still gating this one, because once we can resolve the standard libraries in Tycho I can write in the hlper and revisit whether FordiacProjectLoader becomes usable here. As it is now FordiacProjectLoader.setProjectLocation points the workspace project at the bundle path, so the rename refactoring would mutate the committed fixture. I am happy to add a writable copy variant to it if you would prefer that direction once the library piece lands.

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/Type Library/convert/STRUCT_DEMUX.fbt
oberlehner: There are two possible ways.

Option 1, preferred here: avoid adapting the Maven build and import the libraries directly from the source tree.

`LibraryManager.getStandardLibraries()` is empty in the Tycho test runtime because it only scans `Platform.getInstallLocation()/typelibrary`. In Tycho this is `target/work`, which does not contain the standard libraries. However, we do not need `getStandardLibraries()` for this test. We can call `LibraryManager.importLibrary(...)` with a direct filesystem URI.



Example:

```java
private static void linkStandardLibraries(final IProject project) throws Exception {
	final Path standardLibraries = Path.of(System.getProperty("user.dir"), "..", "..", "data", "typelibrary")
			.toAbsolutePath().normalize();

	LibraryManager.INSTANCE.importLibrary(project, standardLibraries.resolve("core-3.0.0").toUri(), true, false);
	LibraryManager.INSTANCE.importLibrary(project, standardLibraries.resolve("convert-3.0.0").toUri(), true, false);

	TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
}
```

The only caveat is that these links are created under the external library folder, because the URI is a file: URI. For this test that should be fine, since the important part is that the real core and convert libraries are linked and loaded.

Option 2: adapt the Maven build and provision the libraries into Tycho’s install location.

This makes LibraryManager.getStandardLibraries() work, because Tycho starts the test runtime with `-install .../target/work, and LibraryManager scans target/work/typelibrary.`

That would mean copying:

```
data/typelibrary/core-3.0.0
data/typelibrary/convert-3.0.0
```
to:

`${project.build.directory}/work/typelibrary`
during process-test-resources while executing the maven build, for example with maven-resources-plugin.


Given that this test only needs to link the required libraries, I would prefer option 1 and avoid changing the Maven build.

retryoos: Thank you for the advice I implemented option 1, thus linkStandardLibraries now resolves the source tree via Path.of(System.getProperty("user.dir"), "..", "..", "data", "typelibrary") and calls importLibrary for core-3.0.0 and convert-3.0.0 directly. I also dropped the bundled STRUCT_DEMUX.fbt and the shipped project MANIFEST.MF, and the cascade test now loads STRUCT_DEMUX from the linked convert library.

I wanted to note though 2 small changes needed on top of your suggestion to make it pass on this fixture. I create the Standard Libraries / External Libraries virtual folders before importLibrary because otherwise it failed with "parent not accessible" because the fixture project lacks them, and the helper uses the same IResource.VIRTUAL IResource.FORCE with TypeLibraryTags constants pattern as SystemManager.createNew4diacProject. I also dropped the cached TypeLibrary via removeProject before the final refresh() because the .sys file gets parsed during importProjectIntoWorkspace before the libraries are linked, so the Demux instance ends up cached as a plain FB and refresh() alone only reconciles file additions and deletions, not the cached entries.

I noted both in the helper, and now Local Tycho run is green including all 7 tests and cascade.

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
```diff
@@ -0,0 +1,178 @@
...
+		project.open(new NullProgressMonitor());
+		project.refreshLocal(IResource.DEPTH_INFINITE, new NullProgressMonitor());
+		return project;
+	}
+
+	/**
+	 * Link core 3.0.0 and convert 3.0.0 from the source tree so types like
+	 * STRUCT_DEMUX resolve at test runtime instead of being copied into the fixture.
+	 */
+	public static void linkStandardLibraries(final IProject project) throws CoreException {
```
oberlehner: plese do not hardcode the convert-3.0.0 lib here.Extract this as parameter for the method. we may want to test other libraries too in the future!

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeRenameTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+	private void assertOuterMember(final String expectedName, final String expectedQualifiedType) {
+		final VarDeclaration inner = structuredType(OUTER_STRUCT).getMemberVariables().get(0);
+		assertEquals(expectedName, inner.getTypeName());
+		// Comparing the resolved type by identity is fragile because the library
+		// can be reloaded during the refactoring; compare full-qualified names.
+		assertEquals(expectedQualifiedType, PackageNameHelper.getFullTypeName(inner.getType()));
+	}
+
+	private void assertTypeEntry(final IFile typeFile, final String expectedFullTypeName) {
```
oberlehner: please rename to a meaningfull method name like:
asserTypeEntryFullTypeNameEqual

retryoos: Renamed to assertTypeEntryFullTypeNameEqual and all calls updated

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
retryoos: Done, linkStandardLibraries now takes the library directory names as a String varargs parameter, the cascade test calls it with "core-3.0.0", "convert-3.0.0" and future tests can pass whatever libraries they need without touching the helper

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructRenameSystemCascadeTest.java
```diff
@@ -0,0 +1,119 @@
...
+	@BeforeAll
+	static void preloadSystemManager() {
+		SystemManager.INSTANCE.name();
+	}
+
+	@BeforeEach
+	void loadFixture() throws Exception {
+		final Bundle bundle = Platform.getBundle(BUNDLE_NAME);
+		project = RefactoringTestSupport.importProjectIntoWorkspace(PROJECT_NAME, bundle, new Path(PROJECT_PATH));
+		RefactoringTestSupport.linkStandardLibraries(project, "core-3.0.0", "convert-3.0.0"); //$NON-NLS-1$ //$NON-NLS-2$
```
gravemalte: I would pull the hard strings out and put them into a final variable

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeRenameTest.java
```diff
@@ -0,0 +1,163 @@
...
+		// Touch the SystemManager singleton so its resource change listener is
+		// active before any fixture project is imported.
+		SystemManager.INSTANCE.name();
+	}
+
+	@BeforeEach
+	void loadFixture() throws Exception {
+		final Bundle bundle = Platform.getBundle(BUNDLE_NAME);
+		project = RefactoringTestSupport.importProjectIntoWorkspace(PROJECT_NAME, bundle, new Path(PROJECT_PATH));
+		RefactoringTestSupport.linkStandardLibraries(project, "core-3.0.0", "convert-3.0.0"); //$NON-NLS-1$ //$NON-NLS-2$
```
gravemalte: Also here

gravemalte (changes requested):

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructRenameSystemCascadeTest.java
retryoos: Completed I extracted core-3.0.0 and convert-3.0.0 into CORE_LIBRARY and CONVERT_LIBRARY private static final String constants alongside the other constants block, the linkStandardLibraries call uses them

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeRenameTest.java
retryoos: Same fix here, 2 private static final String constants and the call uses them

retryoos:
I rebased onto current dev and squashed everything into a single commit, thus the diff is now down to just the cascade-specific files as the PR description originally promised since #[2273](https://github.com/eclipse-4diac/4diac-ide/pull/2273) was merged and local Tycho run is green, 7 tests, 0 failures, 2 skipped

retryoos:
I added more struct rename tests, including 2 more cascade test methods covering the rest of the ConfigurableFB triad, STRUCT_MUX and F_MOVE instance configurations both update when the struct is renamed, mirroring the existing STRUCT_DEMUX test. For F_MOVE I also linked the iec61131-3 3.0.0 library since that's where it lives.

Also I re-enabled the two @Disabled undo/redo tests in StructDataTypeRenameTest now that #2514's fix is on develop, both pass. Parameterized the configurable FB assertion helper into a single assertConfigurableFBDataType(name, class, expectedType) so the 3 checks share the same logic. So the final status is local tycho 9 tests, 0 failures, 0 errors, 0 skipped.

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
```diff
@@ -72,6 +75,36 @@ public static IProject importProjectIntoWorkspace(final String projectName, fina
...
 	}
 
+	/**
+	 * Link the given standard libraries from the source tree so types they declare
+	 * resolve at test runtime instead of being copied into the fixture.
+	 */
+	public static void linkStandardLibraries(final IProject project, final String... libraryNames) throws CoreException {
+		ensureLibraryFolders(project);
+		final java.nio.file.Path standardLibraries = java.nio.file.Path
+				.of(System.getProperty("user.dir"), "..", "..", "data", "typelibrary") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
```
gravemalte: Why do we have to get a system env here? Am I missing something?

gravemalte (changes requested):

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
retryoos: Thank you for catching this System.getProperty user.dir relied on Tycho launching the test JVM with the bundle directory as cwd which is fragile, I replaced it with an OSGi lookup so the helper resolves the test bundle via FrameworkUtil.getBundle and walks to the repository root from its source location, no system property anymore

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/RefactoringTestSupport.java
```diff
@@ -72,6 +76,45 @@ public static IProject importProjectIntoWorkspace(final String projectName, fina
...
+		TypeLibraryManager.INSTANCE.getTypeLibrary(project).refresh();
+	}
+
+	private static java.nio.file.Path resolveStandardLibrariesPath() throws IOException {
+		// Resolve the test bundle's source location through OSGi and walk up to
+		// the repository root, where the standard libraries live under
+		// data/typelibrary.
+		final Bundle bundle = FrameworkUtil.getBundle(RefactoringTestSupport.class);
+		final java.net.URL bundleRootUrl = FileLocator.toFileURL(bundle.getEntry("/")); //$NON-NLS-1$
+		return Paths.get(bundleRootUrl.getPath()).getParent().getParent().resolve("data").resolve("typelibrary") //$NON-NLS-1$ //$NON-NLS-2$
```
gravemalte: I was thinking about maybe moving this into a helper class (Record)? I find it quite hard to read, and we will use this now multiple times also in #2586.

What do you think @oberlehner

oberlehner: i would just make the method public, since RefactoringTestSupport is already a helper class. Or what would be the benefit of putting that into a record ?

gravemalte: That the object could be constructed and be reused (with the view to other plugin tests, like internal plugins from our project side).
I would like to get out the `.getParent().getParent()....` it's a lot of code that builds up with multiple tests (DRY). A record could hold the data, and we could queue it with an API (not existing yet).

retryoos: I extracted the resolution into a TestRepositoryPaths record with a static resolve() factory exposing repositoryRoot() and standardLibraries(), so the getParent walk lives behind a name and #[2586](https://github.com/eclipse-4diac/4diac-ide/pull/2586) reuses it.

gravemalte:
@oberlehner I'm fine with it now.

oberlehner (approved):

gravemalte (approved):

PR #2518, opened 2026-05-31
[Replay Debugging] Fix colors and make original events read only

cochicde:
- Unify UI colors between application view and the tree-like view. 
- Add matching colors between the state comparison columns and the events seen in the tree-like view
- Allow marking events as read-only and draw a badge icon for those events

Screenshot 2026-05-31 19 11 40


plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/statescomparison/StatesComparisonView.java
```diff
@@ -51,6 +51,9 @@ public class StatesComparisonView extends ViewPart implements ComparisonService.
 
 	private static final String EMPTY = ""; //$NON-NLS-1$
 
+	private static final Color DIFF_HIGHLIGHT_COLOR = new Color(null, 255, 255, 150);
```
azoitl: it would be better to move this to the plugin.xml using the theme api. This allows to better customize it by users and/or provide darkmode colors. 
Also the null is not needed anymore.

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/command/AddToComparisonCommand.java
```diff
@@ -43,28 +44,30 @@ public AddToComparisonCommand(final ComparisonService service, final ReplayNavig
...
 		final var snapshot = replayNavigator.getStateAtEventPosition(eventPosition);
 
-		column = new ComparisonColumn(Long.toString(idCounter++), // unique column ID
-				"", //$NON-NLS-1$ header, empty for now
-				snapshot); // cell data
+		final var rgb = ColorHelper.createRandomColor(COLOR_SATURATION, COLOR_BRIGHTNESS);
+
+		// label empty for now
+		column = new ComparisonColumn(ComparisonService.getInstance().generateUniqueColumnId(), eventPosition, "", //$NON-NLS-1$
+				snapshot, new Color(null, rgb.red, rgb.green, rgb.blue));
```
azoitl: null can be removed.

azoitl (changes requested):

cochicde:
done!

azoitl (approved):
FYI if you later also want to support different colors for darkmode you need to turn your static colors into getters. Otherwise you will not always get the right colors.

PR #2527, opened 2026-06-03
Extracted rename change creation into helper class

oberlehner:


Felix-Schmid (commented):
It seems to me that the work done in the processor and the participant is redundant, since both create the same Change.

mx990:
> It seems to me that the work done in the processor and the participant is redundant, since both create the same Change.

I may be able to shed some light on this. The processor is used when the refactoring is triggered from a graphical context, for example, from a property sheet. There is then another participant[^1] to update the ST code. The participant in this PR is only used with another processor[^2] when the refactoring is triggered from a textual context.

[^1]: see [`org.eclipse.fordiac.ide.structuredtextcore.ui.refactoring.STCoreRenameElementRefactoringParticpant`](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/refactoring/STCoreRenameElementRefactoringParticpant.java)
[^2]: see [`org.eclipse.fordiac.ide.structuredtextcore.ui.refactoring.STCoreRenameElementProcessor`](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/refactoring/STCoreRenameElementProcessor.java)

mx990 (approved):
Thank you for harmonizing both implementations.

Felix-Schmid (approved):
> The processor is used when the refactoring is triggered from a graphical context, for example, from a property sheet. There is then another participant to update the ST code. The participant in this PR is only used with another processor when the refactoring is triggered from a textual context.

I see, that explains it.

azoitl (approved):

PR #2541, opened 2026-06-05
Avoid unnecessary marker update jobs and cleanup marker updates

mx990:
This avoids unnecessary workspace jobs by not forcing an update where possible, particularly when adding types in the type library. It also avoids creating duplicate type library markers.

Furthermore, it directly uses `IResource.deleteMarkers` in builders, where we are guaranteed to hold a workspace lock, and only updates current resources without its children.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
@@ -12,7 +12,6 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.library.builder;
 
-import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -39,7 +38,6 @@ import org.eclipse.fordiac.ide.library.Messages;
 import org.eclipse.fordiac.ide.library.model.library.Manifest;
 import org.eclipse.fordiac.ide.library.model.util.ManifestHelper;
 import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarker;
-import org.eclipse.fordiac.ide.model.errormarker.FordiacMarkerHelper;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
 
 public class LibraryBuilder extends IncrementalProjectBuilder {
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -78,20 +76,24 @@ protected IProject[] build(final int kind, final Map<String, String> args, final
...
-		final SubMonitor progress = SubMonitor.convert(monitor, Messages.LibraryBuilder_CleaningLibrary, 2);
+		final SubMonitor progress = SubMonitor.convert(monitor, Messages.LibraryBuilder_CleaningLibrary,
+				LibraryManager.LIBRARY_FOLDERS.size() + 1);
 
 		// clean manifest library marker
-		FordiacMarkerHelper.updateMarkers(getProject().getFile(LibraryManager.MANIFEST),
-				FordiacErrorMarker.LIBRARY_MARKER, Collections.emptyList(), true);
+		final IFile manifestFile = getProject().getFile(LibraryManager.MANIFEST);
+		if (manifestFile.exists()) {
+			manifestFile.deleteMarkers(FordiacErrorMarker.LIBRARY_MARKER, true, IResource.DEPTH_INFINITE);
```
azoitl: shouldn't this be `IResource.DEPTH_ZERO`?

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
mx990: Yes, thank you for noticing!

oberlehner (approved):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -78,20 +76,24 @@ protected IProject[] build(final int kind, final Map<String, String> args, final
...
-		LibraryManager.LIBRARY_FOLDERS.stream().map(name -> getProject().getFolder(name))
-				.forEach(folder -> FordiacMarkerHelper.updateMarkers(folder, FordiacErrorMarker.LIBRARY_MARKER,
-						Collections.emptyList(), true));
-		progress.worked(1);
-
-		SubMonitor.done(monitor);
+		for (final String name : LibraryManager.LIBRARY_FOLDERS) {
+			final IFolder folder = getProject().getFolder(name);
+			if (folder.exists()) {
+				folder.deleteMarkers(FordiacErrorMarker.LIBRARY_MARKER, true, IResource.DEPTH_INFINITE);
```
azoitl: Stupid question you now have nearly every `DEPTH_ZERO` except here. Is this correct?

mx990: I have now changed it to `DEPTH_ONE`. The reason we cannot use `DEPTH_ZERO` here is that we delete markers for the folders `Standard Libraries` and `External Libraries`, but we also need to include the individual library folders one level below.

azoitl: thx for the clarification.

azoitl (approved):

PR #2567, opened 2026-06-11
Add package rename refactoring

oberlehner:
Package names are now changed through an LTK refactoring instead of direct text editing, so users get the standard refactoring dialog and preview before changes are applied.

The refactoring updates the package name of the selected type and adjusts affected model references, including instances and imports.


Felix-Schmid (commented):
I think we should still use the `PackageSelectionProposalProvider` inside the new dialog.

Felix-Schmid (approved):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/edit/DataTypeEditBuilder.java
```diff
@@ -41,6 +47,38 @@ public static void createStructuredDataTypeChanges(final DataTypeEntry dataTypeE
...
+				}
+			}
+		}
+	}
+
+	private static boolean hasImport(final LibraryElement libraryElement, final String targetTypeName,
+			final Import currentImport) {
+		return libraryElement.getCompilerInfo().getImports().stream()
+				.anyMatch(imp -> imp != currentImport && Objects.equals(imp.getImportedNamespace(), targetTypeName));
+	}
```
mx990: Imports are generally handled by the ST refactoring participant, so this will cause conflicts once we add an ST participant for package changes.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/ChangePackageNameRefactoring.java
```diff
@@ -0,0 +1,157 @@
...
+		} catch (final OperationCanceledException e) {
+			// ignore
+		} catch (final InterruptedException e) {
+			Thread.currentThread().interrupt();
+		} catch (final Exception e) {
+			FordiacLogHelper.logError("Error during package name refactoring", e); //$NON-NLS-1$
+		}
+	}
+
+	public static void run(final TypeEntry typeEntry, final String newPackageName, final Shell shell) {
```
mx990: It seems this method is unused?

mx990 (changes requested):
Please note that this refactoring currently does not update references from ST code.

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/edit/DataTypeEditBuilder.java
oberlehner: Which ST participant does handle this ?
This is not a file rename, and not a file move, so i think the current participants cannot handle that.

My approach now would be to:
1) Create a `org.eclipse.fordiac.ide.typemanagement.refactoring.ChangePackageNameRefactoringProcessor` derived from RenameProcessor  .
Use this processor inside the changePackageNameRefactoring

2) Create  `org.eclipse.fordiac.ide.structuredtextcore.ui.refactoring.STCorePackageNameChangeParticipant ` that calls the STCoreImporter for updating the imports

Do you think this is the cleanest approach, or do you have other suggestions ?

mx990: Yes, I also believe that would be the cleanest approach.

mx990 (approved):

azoitl (approved):

PR #2577, opened 2026-06-15
Start to Restrict API access of Plugins

azoitl:
In order to better hide implementations from other plugins internal implementation plugins are marked with x-internal.

Furthermore as reference example all ECC editor plugins are marked internal as this classes should not be accessed from other plugins.

This PR shall serve as example for future work to better restrict plugin implementations.


Felix-Schmid (approved):

MandKastner (approved):

gravemalte (approved):

mx990 (approved):
I assume the remaining warnings introduced by this PR should be fixed separately?

azoitl:
> I assume the remaining warnings introduced by this PR should be fixed separately?

@mx990 I fixed the most obvious ones. There are several now that need a bit more work and thinking. Therefore I think this will be something for several isolated PRs.

PR #2581, opened 2026-06-16
Add EtherCAT ESI import and typelibrary support

wzj7531:
## Summary

- Add org.eclipse.fordiac.ide.ethercat plugin with ESI import wizard
- Add EtherCAT typelibrary entries (ECMaster, ECBusAdapter, config types)
- Extend deployment for EtherCAT generic FB type names and initial values
- Reuse the existing monitoring MGM connection for online deployment commands (Delete Resource, etc.) when a device is already monitored

## Related PRs

- eclipse-4diac/4diac-forte#968 (PR 1 of 3 — core library; follow-up FORTE PRs planned for IO FBs and ESI)
- eclipse-4diac/4diac-fbe#72

## Test plan

- [ ] mvn install succeeds
- [ ] ESI import wizard generates FB types
- [ ] Deploy EtherCAT application to forte
- [ ] Delete Resource / Kill works while device monitoring is active


azoitl:
Similarly as the 4diac FORTE PR also think it would be good to split this PR in several smaller ones. 

I like the idea of deriving the type information for deployment. But as written in the discussion I would prefer if we could move it to an Attribute. Similar to the type hash attribute. This avoids any potential problems with other users of the description field type.

plugins/org.eclipse.fordiac.ide.ethercat/icons/ethercat.svg
franz-hoepfinger-4diac, line 1: i guess the Licence for the EtherCAT Logo as well as Trademark need to be checked, right ?

azoitl: @franz-hoepfinger-4diac yes this is something we will have to consider here.

azoitl:
@wzj7531 as we have the 4diac FORTE part ready shall we continue with the 4diac IDE PRs? As written above it would be great if you could split this PR into smaller parts. Furthermore I think we need to find a different icon as I'm not sure if we are allowed to use the official EtherCat logo here. Do you know more?

wzj7531:
> @wzj7531 as we have the 4diac FORTE part ready shall we continue with the 4diac IDE PRs? As written above it would be great if you could split this PR into smaller parts. Furthermore I think we need to find a different icon as I'm not sure if we are allowed to use the official EtherCat logo here. Do you know more?

OK, I will draw a icon to replace the original one.

wzj7531:
Thanks for the feedback — I'll split this into smaller PRs as suggested.

Plan:
1. **PR1** — Resolve package-qualified `GenericClassName` for FORTE deployment (Attribute-based, similar to TypeHash; short names like `GEN_PUBLISH` do not override concrete type names)
2. **PR2** — Add EtherCAT types to the IO typelibrary (`ECController`, configs, `ECBusAdapter`, `RevisionNo` attribute declaration)
3. **PR3** — Add the EtherCAT ESI import wizard (no icon for now, to avoid using the official EtherCAT logo)

I will open PR1 and PR2 shortly and follow up with PR3 after those land. This PR will then be closed in favor of the split series.

azoitl:
@wzj7531 thx for the new plan. Looks good to me. Regarding the icon: Please note that we can now have icons in SVG format. This helps for getting better looking icons on different screen scaling factors. In the icons directory you can already find a few that you can start from.

PR #2583, opened 2026-06-16
Clear standard libraries during p2 updates

oberlehner:
Clear standard libraries during p2 updates

During a p2 update, automatic builds can still resolve standard library
dependencies from the product typelibrary. On Windows this can keep files open
while p2 tries to back up/remove them, causing the update to fail.

This clears the standard library cache when the p2 update job starts, so the
builder skips standard library resolution until the updated IDE is restarted.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
@@ -39,6 +39,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
@@ -61,7 +63,10 @@ import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.URIUtil;
+import org.eclipse.core.runtime.jobs.IJobChangeEvent;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobChangeAdapter;
+import org.eclipse.equinox.p2.operations.IProfileChangeJob;
 import org.eclipse.fordiac.ide.library.download.DownloadResult;
 import org.eclipse.fordiac.ide.library.download.IArchiveDownloader;
 import org.eclipse.fordiac.ide.library.model.library.Manifest;
@@ -82,6 +87,8 @@ import org.osgi.framework.VersionRange;
 public enum LibraryManager {
 
 	INSTANCE;
+
+	private static final String UPDATE_JOB_NAME = "Updating Software"; //$NON-NLS-1$
 
 	public static final String LIB_TYPELIB_FOLDER_NAME = "typelib"; //$NON-NLS-1$
 	public static final String PACKAGE_DOWNLOAD_DIRECTORY = ".download"; //$NON-NLS-1$
@@ -116,7 +123,8 @@ public enum LibraryManager {
 			VersionRange.RIGHT_CLOSED);
 
 	private WatchService watchService;
-	private final HashMap<String, List<LibraryRecord>> stdlibraries = new HashMap<>();
+	private final AtomicBoolean standardLibraryResolutionEnabled = new AtomicBoolean(true);
+	private final Map<String, List<LibraryRecord>> stdlibraries = new ConcurrentHashMap<>();
 	private final HashMap<String, List<LibraryRecord>> libraries = new HashMap<>();
 
 	public static final Object FAMILY_FORDIAC_LIBRARY = new Object();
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -480,7 +494,9 @@ public Path[] listArchiveFolders(final Path path) {
 	 * @return map of libraries
 	 */
 	public Map<String, List<LibraryRecord>> getStandardLibraries() {
-		return new HashMap<>(stdlibraries);
+		synchronized (stdlibraries) {
+			return new HashMap<>(stdlibraries);
```
gravemalte: What about a `ConcurrentHashMap` instead of locking?

gravemalte (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -116,7 +122,8 @@ public enum LibraryManager {
 			VersionRange.RIGHT_CLOSED);
 
 	private WatchService watchService;
-	private final HashMap<String, List<LibraryRecord>> stdlibraries = new HashMap<>();
+	private volatile boolean standardLibraryResolutionEnabled = true;
```
mx990: I would suggest using an `AtomicBoolean` instead of a `volatile` field.

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -1065,4 +1077,40 @@ private static Stream<Version> getAvailableVersions(final Map<String, List<Libra
...
+
+	@Override
+	public void sleeping(final IJobChangeEvent event) {
+		// no action needed
+	}
+
+	@Override
+	public void done(final IJobChangeEvent event) {
+		// no action needed
+	}
```
mx990: I would suggest encapsulating the `IJobChangeListener` into an anonymous or nested class instead of adding public methods to `LibraryManager`. I would also suggest inheriting from `JobChangeAdapter` to avoid unused methods.

mx990 (changes requested):

mx990 (approved):

gravemalte (approved):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
```diff
@@ -149,6 +157,14 @@ public static LibraryManagerData init() {
 		}
 
 		LibraryPermission.setLibReadOnly(standardLibraryPath);
+		Job.getJobManager().addJobChangeListener(new JobChangeAdapter() {
+			@Override
+			public void aboutToRun(final IJobChangeEvent event) {
+				if (isUpdateProvisioningJob(event.getJob())) {
+					standardLibraryResolutionEnabled.set(false);
```
azoitl, line 168: Shouldn't this flag at one point be set to true again? Or as after a p2 update there is a restart required and this it setting it to true. Maybe that should be put into a comment.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/LibraryManager.java
gravemalte: We discussed this. Please correct me if I'm wrong. In our case p2 always requires and restart (I don't know if p2 supports hot-reloading), so this flag has to be changed only during the update process. But I agree. A comment would make sense.

azoitl: I as user can say I'm not restarting now. So maybe that should also be considered. If it is smart as user to do that is a different story.

gravemalte: True then its a different story. Then we should switch it I guess.

oberlehner: i can implement the `org.eclipse.core.runtime.jobs.JobChangeAdapter.done(IJobChangeEvent)` method if you want ?

azoitl: Maybe it is a bit over defensive but most of the time we are not overdefinsive it later bites our a**. So I think it would be good if you could do that.

gravemalte: I agree, with the `done(..)` implementation this case should be fixed.

oberlehner: Done

gravemalte (approved):

mx990 (approved):

azoitl (approved):

PR #2586, opened 2026-06-16
Add struct and FB type delete tests with nested SubApp fixture

retryoos:
Adds delete tests for struct data types and FB types.

Struct delete: the .dtp file is removed, the type is gone from the library, and the outer struct member can no longer resolve its type.

FB delete: the .fbt file is removed, the type is gone, and the FB instances stay in place, and the fixture instances a block across three nested subapp levels to check this.

The fixture block is a service interface FB with a struct-typed in/out pin, wired with event and data connections between the instances.

### DIFF
--- a/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeDeleteTest.java
+++ b/tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeDeleteTest.java
@@ -1,0 +1,107 @@
+/*******************************************************************************
+ * Copyright (c) 2026
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Dimitrios Kalligaridis - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.typemanagement.tests;
+
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.CONVERT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.CORE;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StandardLibrary.IEC_61131_3;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.INNER_STRUCT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.INNER_STRUCT_FILE;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.OUTER_STRUCT;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.PROJECT_NAME;
+import static org.eclipse.fordiac.ide.typemanagement.tests.StructRenameTestFixture.PROJECT_PATH;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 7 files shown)


tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeDeleteTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.jupiter.api.Test;
+import org.osgi.framework.Bundle;
+
+class StructDataTypeDeleteTest {
+
+	private static final String BUNDLE_NAME = "org.eclipse.fordiac.ide.typemanagement.tests"; //$NON-NLS-1$
+	private static final String PROJECT_NAME = "StructRenameTest"; //$NON-NLS-1$
+	private static final String PROJECT_PATH = "data/StructRenameTest"; //$NON-NLS-1$
+
+	private static final String CORE_LIBRARY = "core-3.0.0"; //$NON-NLS-1$
```
oberlehner: please move that into a helper class. you have similar constants at multiple places all over your code. Please improve that!

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/StructRenameTest.sys
```diff
@@ -0,0 +1,20 @@
...
+		<SubAppNetwork>
+			<FB Name="Producer" Type="StructProducer" Comment="" x="500.0" y="500.0">
+			</FB>
+			<FB Name="Demux" Type="STRUCT_DEMUX" Comment="" x="1500.0" y="500.0">
+				<Attribute Name="StructuredType" Type="STRING" Value="mypackage::InnerStruct"/>
+			</FB>
+			<FB Name="Mux" Type="STRUCT_MUX" Comment="" x="1500.0" y="1500.0">
+				<Attribute Name="StructuredType" Type="STRING" Value="mypackage::InnerStruct"/>
+			</FB>
+			<FB Name="Move" Type="F_MOVE" Comment="" x="2500.0" y="500.0">
```
oberlehner, line 15: Please add some simple,basic, sifb. These blocks should use the strucuted data types at the interface. 
They also should be expanded at some points, and have connections !
Example:
Image

tests/org.eclipse.fordiac.ide.typemanagement.tests/src/org/eclipse/fordiac/ide/typemanagement/tests/StructDataTypeDeleteTest.java
retryoos: Moved the shared constants into helper types, StandardLibrary for the library names and StructRenameTestFixture for the fixture coordinates, and dropped BUNDLE_NAME entirely by resolving the bundle inside importProjectIntoWorkspace. Now the duplicated constant blocks are gone across all the test classes

tests/org.eclipse.fordiac.ide.typemanagement.tests/data/StructRenameTest/StructRenameTest.sys
retryoos: I am working on it, I will add a basic FB that uses the struct type at its interface, expand the struct pins, and wire connections between the expanded members so the rename and delete cascades are exercised against real connections

retryoos: Added a StructConsumer simple FB that uses InnerStruct at its interface (DI/DO1), gave InnerStruct a few members, and wired Producer.OUT to Consumer.DI as a struct connection in the app. Also I added a cascade assertion so the new FB struct pins are verified to update to InnerStructRenamed on rename

oberlehner:
Please at least read your PR description before you copy that from your AI agent. 

image

retryoos:
> Please at least read your PR description before you copy that from your AI agent.
> 
> image

I apologize for the oversight Michael, I will make sure I won't repeat this moving forward it was a one time mistake

oberlehner (approved):

PR #2589, opened 2026-06-17
Show ECC state name in breakpoints view and stack trace

vikash1703:


plugins/org.eclipse.fordiac.ide.debug/src/org/eclipse/fordiac/ide/debug/EvaluatorDebugStackFrame.java
```diff
@@ -185,6 +186,10 @@ public int getCharEnd() {
 
 	@Override
 	public String getName() {
+		final Object context = getCurrentContext();
+		if (context instanceof final INamedElement element) {
+			return evaluator.getName() + " [" + element.getName() + "]"; //$NON-NLS-1$ //$NON-NLS-2$
+		}
```
mx990: I do not believe it makes sense to add this here for all evaluators. If you want to include the current `ECState` for the `BasicFBEvaluator`, I would suggest to override `getName()` in `BasicFBEvaluator` instead.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/EvaluatorDebugModelPresentation.java
```diff
@@ -113,6 +117,12 @@ public String getText(final Object element) {
 		if (element instanceof final EvaluatorDebugVariable variable) {
 			return getVariableText(variable);
 		}
+		if (element instanceof final EvaluatorModelBreakpoint breakpoint) {
+			final IMarker marker = breakpoint.getMarker();
+			if (marker != null) {
+				return "ECC State Breakpoint: " + marker.getAttribute(IMarker.LOCATION, ""); //$NON-NLS-1$ //$NON-NLS-2$
+			}
```
mx990, line 121: Please follow the convention that the breakpoint text generally starts with the resource name followed by specific information in brackets (e.g., `BasicTest.fbt [State_1]`). Please also use `EvaluatorModelBreakpoint.getQualifiedName()` instead of accessing the marker directly.

plugins/org.eclipse.fordiac.ide.debug.ui/src/org/eclipse/fordiac/ide/debug/ui/EvaluatorDebugModelPresentation.java
```diff
@@ -81,6 +82,9 @@ public IEditorInput getEditorInput(final Object element) {
 
 	@Override
 	public String getEditorId(final IEditorInput input, final Object element) {
+		if (element instanceof org.eclipse.fordiac.ide.model.libraryElement.ECState) {
+			return "org.eclipse.fordiac.ide.fbtypeeditor.ecc.editors.ECCEditor"; //$NON-NLS-1$
+		}
```
mx990: This has no effect, because the source element is not the `ECState`. The `ECCEditor` is also not the correct editor id here, since it is merely a sub-editor of the multi-page FB type editor. I believe this requires more careful planning and is also something for a separate PR.

mx990 (changes requested):

PR #2590, opened 2026-06-17
Add type rename button to Type Info section

oberlehner:
Adds a rename action next to the type name in the Type Info section.

The new magic-wand button opens the standard type rename refactoring flow, so renaming from the properties view behaves like renaming the type file from the explorer: the file, type name, and references are updated through the existing refactoring participants.

Also makes the displayed type name selectable, so users can copy it directly from the properties view.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
+++ b/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
@@ -22,21 +22,36 @@ import org.eclipse.fordiac.ide.gef.widgets.PackageInfoWidget;
 import org.eclipse.fordiac.ide.model.commands.change.ChangeCommentCommand;
 import org.eclipse.fordiac.ide.model.libraryElement.FunctionFBType;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
+import org.eclipse.fordiac.ide.typemanagement.Messages;
+import org.eclipse.fordiac.ide.typemanagement.refactoring.TypeRefactoringHelper;
 import org.eclipse.fordiac.ide.ui.FordiacMessages;
+import org.eclipse.jface.layout.GridDataFactory;
+import org.eclipse.jface.layout.GridLayoutFactory;
+import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionListener;
+import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.layout.GridData;
-import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Text;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.plugin.AbstractUIPlugin;
 import org.eclipse.ui.views.properties.tabbed.TabbedPropertySheetPage;
 
 /** Properties tab which shows the FB type information of the selected FB */
 public abstract class TypeInfoSection extends AbstractDoubleColumnSection {
 
+	private static final String RENAME_ELEMENT_ICON = "icons/full/etool16/tricks.png"; //$NON-NLS-1$
+
 	private PackageInfoWidget typeInfo;
 
 	private Text fbTypeNameText;
 	private Text commentText;
+	private Button renameTypeButton;
+	private Image renameTypeImage;
 
 	private final Adapter typeInfoAdapter = new EContentAdapter() {
 		@Override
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
```diff
@@ -66,24 +82,59 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
 
 	private void createTypeAndCommentSection(final Composite parent) {
 		final Composite composite = getWidgetFactory().createComposite(parent);
-		composite.setLayout(new GridLayout(2, false));
+		GridLayoutFactory.fillDefaults().numColumns(3).equalWidth(false).applyTo(composite);
 		composite.setLayoutData(new GridData(SWT.FILL, 0, true, false));
 		getWidgetFactory().createCLabel(composite, FordiacMessages.TypeName + ":"); //$NON-NLS-1$
-		fbTypeNameText = createGroupText(composite, false);
+		renameTypeButton = getWidgetFactory().createButton(composite, null, SWT.PUSH);
```
azoitl, line 86: I would move the button to the end. This feels for me more natural

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
```diff
@@ -66,24 +82,59 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
...
+		renameTypeButton.setToolTipText(Messages.RenameType_Name);
+		renameTypeButton.setImage(getRenameTypeImage());
+		renameTypeButton.addDisposeListener(e -> disposeRenameTypeImage());
+		renameTypeButton.addSelectionListener(SelectionListener.widgetSelectedAdapter(ev -> openTypeRefactoring()));
+		GridDataFactory.swtDefaults().align(SWT.LEFT, SWT.CENTER).applyTo(renameTypeButton);
+		fbTypeNameText = createGroupStyledText(composite);
 		getWidgetFactory().createCLabel(composite, FordiacMessages.Comment + ":"); //$NON-NLS-1$
+		getWidgetFactory().createLabel(composite, ""); //$NON-NLS-1$
 		commentText = createGroupText(composite, true);
```
azoitl, line 97: `commentText` should now use two columns to also use the space of the button.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
```diff
@@ -66,24 +82,59 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
 
 	private void createTypeAndCommentSection(final Composite parent) {
 		final Composite composite = getWidgetFactory().createComposite(parent);
-		composite.setLayout(new GridLayout(2, false));
+		GridLayoutFactory.fillDefaults().numColumns(3).equalWidth(false).applyTo(composite);
 		composite.setLayoutData(new GridData(SWT.FILL, 0, true, false));
 		getWidgetFactory().createCLabel(composite, FordiacMessages.TypeName + ":"); //$NON-NLS-1$
```
azoitl: as you are improving this property sheet please change the CLable to Label this gives better look and feel.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
```diff
@@ -66,24 +82,59 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
...
+		if (renameTypeImage != null && !renameTypeImage.isDisposed()) {
+			renameTypeImage.dispose();
+		}
+		renameTypeImage = null;
+	}
+
+	private void openTypeRefactoring() {
+		final TypeEntry typeEntry = getType().getTypeEntry();
+		if (typeEntry != null && typeEntry.getFile() != null) {
+			RenameTypeRefactoring.openWizard(typeEntry, fbTypeNameText.getShell());
```
azoitl: Could we somehow inovke the rename command so that we don't need th renametyperefactoring helper classe?

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
```diff
@@ -66,24 +82,59 @@ public void createControls(final Composite parent, final TabbedPropertySheetPage
...
-		fbTypeNameText = createGroupText(composite, false);
+		renameTypeButton = getWidgetFactory().createButton(composite, null, SWT.PUSH);
+		renameTypeButton.setToolTipText(Messages.RenameType_Name);
+		renameTypeButton.setImage(getRenameTypeImage());
+		renameTypeButton.addDisposeListener(e -> disposeRenameTypeImage());
+		renameTypeButton.addSelectionListener(SelectionListener.widgetSelectedAdapter(ev -> openTypeRefactoring()));
+		GridDataFactory.swtDefaults().align(SWT.LEFT, SWT.CENTER).applyTo(renameTypeButton);
+		fbTypeNameText = createGroupStyledText(composite);
 		getWidgetFactory().createCLabel(composite, FordiacMessages.Comment + ":"); //$NON-NLS-1$
+		getWidgetFactory().createLabel(composite, ""); //$NON-NLS-1$
```
azoitl: this emtpy label should definitly not be here.

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
oberlehner: Which rename command do you mean ?
If i call the refactoring wizard directly we would get an LTK depdency to the gef plugin.
I found a better place for the helper method ;)

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
oberlehner: This looks strange in my oppion. All other rename,reorder, delete,... buttons are on the left side. 
I placed the button between the label and the type name. This keeps the action visually associated with the type name and matches the layout used by the new package information rename button.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
oberlehner: Please verify if you like the layout after the replacement of the button

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
azoitl: I tend to dissagree:

We have the buttons on the left only for tables and trees. As soon as we modify a single element we have the button integrated into the text box or at the and. 

If you look at this screenshot:
Image

I read Type Name: button then the name. This totally breaks the reading flow. Because the button is not the type name it modifies the type name.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
azoitl: I thought about invoking the ltk rename command via the Eclipse platform command service infrastructure as we do it with the other reanme refactoring buttons. However I fear this will anyhow not work as this command looks at the current selection and changing the selection before invoking the command may not be a good idea.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
azoitl: Comment text now looks good.

plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/properties/TypeInfoSection.java
oberlehner: However, its anyway incosistent with the package rename refactoring 
image


So we either put both rename buttons on the left side or on the right side. 
Since the buttons of Version Info and Imports are also on the left side i would suggest we put everything on the left side

azoitl: You are right I missed the Package name button. I still think that the type name and package name refactoring buttons should be right otherwise the context is lost. If you put the button on the full left the user sees the button read then text. For the tables there is always a headline giving the context therefore there left is fine or even better.

azoitl (approved):
I'm not fully convienced but having it consistently full left is better then in the middle.

PR #2592, opened 2026-06-17
Export Build should only export existing files

azoitl:
During refactoring we get change notification for the renamed old version of the file which is not existing anymore.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
+++ b/plugins/org.eclipse.fordiac.ide.export/src/org/eclipse/fordiac/ide/export/builder/ExportBuilder.java
@@ -180,7 +180,7 @@ public class ExportBuilder extends IncrementalProjectBuilder {
 				throw new OperationCanceledException();
 			}
 
-			if ((delta.getResource() instanceof final IFile file) && includeInIncrementalBuild(file)) {
+			if ((delta.getResource() instanceof final IFile file) && file.exists() && includeInIncrementalBuild(file)) {
 				exportElement(monitor, file, context);
 			}
 			return true;


mx990 (commented):
If the original type file has been removed, we should probably also delete the generated files?

azoitl:
> If the original type file has been removed, we should probably also delete the generated files?

This is a good point. However our exporter interface currently does not provide that information. So we would need to extend it. @MandKastner WDYT?

MandKastner:
You are right, the generated files should be removed. As we do not want to cache in the builder and currently the relation is unidirectional, i think we only have two options. Change the exporter interface to support removing files in the output dir, or handle remove notifications of exportable elements on the buildpath by cleaning and triggering a full build, otherwise we cannot guarantee the correct state. The builder would be easier, but it comes with a lot of  unnecessary operations.

MandKastner (approved):

oberlehner:
> If the original type file has been removed, we should probably also delete the generated files?

I would suggest to handle this in a delete refactoring participant

MandKastner:
> > If the original type file has been removed, we should probably also delete the generated files?
> 
> I would suggest to handle this in a delete refactoring participant

I think this might not be possible, as in general the output structure is dependent on the specific export filter. I think you would not know which files you have to delete.

azoitl:
> > If the original type file has been removed, we should probably also delete the generated files?
> 
> I would suggest to handle this in a delete refactoring participant

I don't think this is a good idea. First of all it would mean that the export builder information would need to be propagated to the refactoring second we still would need the information of what an export filter is potentially generating.

MandKastner:
As a temporary solution: 
https://github.com/eclipse-4diac/4diac-ide/pull/2599

oberlehner:
> > > If the original type file has been removed, we should probably also delete the generated files?
> > 
> > 
> > I would suggest to handle this in a delete refactoring participant
> 
> I don't think this is a good idea. First of all it would mean that the export builder information would need to be propagated to the refactoring second we still would need the information of what an export filter is potentially generating.

Yes you are right, i misudnerstood the use case at first glance. i think, clearing everything and rebuild is the best solution!

oberlehner (approved):

mx990 (approved):

PR #2594, opened 2026-06-18
Preserve error marker values when saving

oberlehner:
When a member with an initial value and/or comment was removed from a structured data type, the corresponding STRUCT_MUX error pin was initially displayed correctly.

Saving the system, however, discarded the error pin’s initial value from the XML. For pins containing only an initial value, the complete parameter entry was removed. Comment-only pins remained but no longer produced a validation marker.

This change:

- preserves initial values stored on ErrorMarkerInterface during export;
- keeps the parameter entry when only an initial value exists;
- reports missing variables when a preserved comment exists;
- keeps the error pin and Problems View entry after saving.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ErrorMarkerInterfaceAnnotations.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ErrorMarkerInterfaceAnnotations.java
@@ -42,6 +42,22 @@ public final class ErrorMarkerInterfaceAnnotations {
 		return true;
 	}
 
+	public static boolean validateComment(final ErrorMarkerInterface element, final DiagnosticChain diagnostics,
+			final Map<Object, Object> context) {
+		if (hasComment(element) && !isContainedInErrorMarkerFBNElement(element)) {
+			if (diagnostics != null) {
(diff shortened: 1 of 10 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ErrorMarkerInterfaceAnnotations.java
```diff
@@ -39,6 +42,17 @@ public static boolean validateValue(final ErrorMarkerInterface element, final Di
...
+			if (diagnostics != null) {
+				diagnostics
+						.add(new BasicDiagnostic(Diagnostic.ERROR, LibraryElementValidator.DIAGNOSTIC_SOURCE,
+								LibraryElementValidator.ERROR_MARKER_INTERFACE__VALIDATE_VALUE,
+								MessageFormat.format(Messages.ErrorMarkerInterfaceAnnotations_MissingVariableForComment,
+										element.getComment()),
+								FordiacMarkerHelper.getDiagnosticData(element)));
+			}
+			return false;
+		}
```
mx990: I believe this should be moved to a separate `validateComment` model invariant.

mx990 (changes requested):

mx990 (approved):

PR #2608, opened 2026-06-23
Preserve error marker attributes on type rename

oberlehner:
During rename refactoring, affected instances can temporarily go through an
error-marker state. Previously, error-marker pins were only restored when they
had a non-empty value. Attribute-only pins, such as annotated output variables,
were therefore dropped.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
+++ b/plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
@@ -42,10 +42,10 @@ import org.eclipse.fordiac.ide.model.datatype.helper.InternalAttributeDeclaratio
 import org.eclipse.fordiac.ide.model.errormarker.FordiacErrorMarkerInterfaceHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.AdapterFB;
 import org.eclipse.fordiac.ide.model.libraryElement.AdapterType;
-import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
 import org.eclipse.fordiac.ide.model.libraryElement.BlockFBNetworkElement;
 import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableFB;
 import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableMoveFB;
+import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableObject;
 import org.eclipse.fordiac.ide.model.libraryElement.Connection;
 import org.eclipse.fordiac.ide.model.libraryElement.Demultiplexer;
 import org.eclipse.fordiac.ide.model.libraryElement.ErrorMarkerFBNElement;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
```diff
@@ -447,27 +448,63 @@ private static boolean varDeclHasNonInternalAttr(final VarDeclaration vd) {
...
+	}
+
+	private static void copyAttribute(final Attribute attribute, final ConfigurableObject destination) {
+		if (attribute.getAttributeDeclaration() != null) {
+			destination.setAttribute(attribute.getAttributeDeclaration(), attribute.getValue(), attribute.getComment());
+		} else {
+			destination.setAttribute(attribute.getName(), attribute.getType(), attribute.getValue(),
+					attribute.getComment());
+		}
+	}
```
mx990: Would it not be easier to just `EcoreUtil.copy` the whole attribute?

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
```diff
@@ -447,27 +448,63 @@ private static boolean varDeclHasNonInternalAttr(final VarDeclaration vd) {
...
+						final Value value = LibraryElementFactory.eINSTANCE.createValue();
+						value.setValue(errorMarker.getValue().getValue());
+						varDeclaration.setValue(value);
+					}
+					if (errorMarker.isIsInput() && errorMarker.getInputConnections().isEmpty()
+							&& hasValue(errorMarker.getValue())) {
 						// remove errormarker because value was set to pin and no connection needs to be
 						// copied
-						reconnCmds.add(new DeleteErrorMarkerCommand(erroMarker, oldElement));
+						reconnCmds.add(new DeleteErrorMarkerCommand(errorMarker, oldElement));
```
mx990: Why do we delete the error pin in the *old* element in the first place and why only if it has a value?

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.commands/src/org/eclipse/fordiac/ide/model/commands/change/AbstractUpdateBlockFBNElementCommand.java
oberlehner: Good point. i think it does not make any sense here deleting the error marker, since the old element is removed anyway. It is some leftover code,  i was just afraid to delete it.

I will just remove the whole check 

```
if (errorMarker.isIsInput() && errorMarker.getInputConnections().isEmpty()
        && hasValue(errorMarker.getValue())) {
    reconnCmds.add(new DeleteErrorMarkerCommand(errorMarker, oldElement));
}
```

if you agree

mx990: Yes, please do.

mx990 (approved):

azoitl (approved):

PR #2611, opened 2026-06-24
Fix/ecc anchor stream optimization

vikash1703:
Use Java Streams to combine incoming and outgoing transitions during anchor calculation. This prevents mixed transitions from overlapping on the same state edge and avoids the memory overhead of intermediate lists.


azoitl (changes requested):
Unfortunately it is only working for the left and bottom edge. Not for top and left. I currently do not see an obvious reason why this is the case. Maybe one reason is the in line 96 your are not translating to absolute. But I'm not sure.

azoitl (approved):

PR #2628, opened 2026-07-01
Cherry picks from 3.3.x

oberlehner:


oberlehner:
see https://github.com/eclipse-4diac/4diac-ide/pull/2629

PR #2635, opened 2026-07-04
fix: Correct the order of adapter list creation in interface export

franz-hoepfinger-4diac:
Adjust the order of 'createAdapterList' method calls to maintain consistency with the expected sequence for exporting interface lists.

https://github.com/franz-hoepfinger-4diac-org/4diac-ide/pull/19
solves eclipse-4diac#2634


azoitl:
As this is changing the hash calculation it can not be merged as is. We are currently finalizing the XSD files describing the 4diac IDE IEC 61499 xml representation. After that we can identify what changes to loading or saving of XML files is needed. Based on that we want to have a more gracefull update process so that hash migration is possible for our users with upfront notification.

diplfranzhoepfinger:
Understand.

PR #2642, opened 2026-07-06
Manage Library Wizard validation and preview page

MandKastner:


oberlehner:
If i have two conflicts, i only see the latest. The conflict of motor shadows the conflict of lib02. It is also not visible if a select the row motor. 

image

image


After deselection the action from motor, the conflict if lib02 is visible again
image

We can either show the correct conlfict by seleciton, or appending the string so that all conflicts are visbible

oberlehner (approved):

PR #2651, opened 2026-07-09
Fix rename refactoring for structured variable members

oberlehner:


plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
```diff
@@ -75,8 +101,7 @@ private static void createRenameInterfaceChanges(final List<ModelEdit<?>> modelE
 				.getInterfaceElement(List.of(interfaceElement.getName()));
```
mx990: Would it not be enough to just change this to:

```suggestion
final IInterfaceElement instancePin = element.getInterface()
.getInterfaceElement(interfaceElement);
```

and let `InterfaceList.getInterfaceElement(IInterfaceElement)` handle the proper resolution of the member pin?

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
```diff
@@ -62,10 +63,35 @@ private static void createChildChanges(final List<ModelEdit<?>> modelEdits, fina
...
+			}
+		}
+	}
+
+	private static void createRenameInterfaceChanges(final List<ModelEdit<?>> modelEdits,
+			final ContainerVarDeclaration element, final IInterfaceElement interfaceElement, final String newName) {
+		final BlockFBNetworkElement blockElement = element.getBlockFBNetworkElement();
+		if (blockElement != null) {
+			final List<String> path = new ArrayList<>(element.getBlockRelativePath());
+			path.addAll(interfaceElement.getBlockRelativePath());
```
mx990: The block relative path of `interfaceElement` should already include the block relative path of `element` as a prefix if `interfaceElement` is a member of `element`.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
oberlehner: Done

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
oberlehner: Done

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
```diff
@@ -62,21 +63,43 @@ private static void createChildChanges(final List<ModelEdit<?>> modelEdits, fina
...
-					.forEach(element -> createRenameInterfaceChanges(modelEdits, element, interfaceElement, newName));
+			createRenameChanges(modelEdits, newName, result, interfaceElement);
+		}
+	}
+
+	protected static void createRenameChanges(final List<ModelEdit<?>> modelEdits, final String newName,
+			final List<? extends EObject> result, final IInterfaceElement interfaceElement) {
+		for (final EObject element : result) {
+			if (element instanceof final ContainerVarDeclaration container) {
+				createRenameInterfaceChanges(modelEdits, container, interfaceElement, newName);
```
mx990: I believe this can now be replaced with:
```suggestion
				createRenameInterfaceChanges(modelEdits, container.getBlockFBNetworkElement(), interfaceElement, newName);
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/rename/RenameElementRefactoringHelper.java
oberlehner: I think not.
This will produce null in line 99 `final IInterfaceElement instancePin = element.getInterface().getInterfaceElement(interfaceElement);` when renaming a struct member where its instance is part of an expanded pin.

mx990: Ah, I got confused because I thought `interfaceElement` was part of an FB interface, but it can also be a struct member. In that case, using the block-relative path of the struct member also makes little sense.

Would it be possible to have the search return the referencing pin itself instead of its container, so that we can avoid looking up the pin again here?

Otherwise, I believe the most generic way to look up the referencing pin would be:
```java
element.findBySimpleName(interfaceElement.getName()).filter(IInterfaceElement.class::isInstance).map(IInterfaceElement.class::cast).findFirst()
```
In that case, I would also suggest to rename `interfaceElement` to `renamedElement` to make the intent at least a tiny bit more clear.

oberlehner: `Would it be possible to have the search return the referencing pin itself instead of its container, so that we can avoid looking up the pin again here?` 

I think this will make the code much cleaner. 
Getting the container is really confusing in my oppinion. 
I will go for that

oberlehner: > Ah, I got confused because I thought `interfaceElement` was part of an FB interface, but it can also be a struct member. In that case, using the block-relative path of the struct member also makes little sense.
> 
> Would it be possible to have the search return the referencing pin itself instead of its container, so that we can avoid looking up the pin again here?
> 
> Otherwise, I believe the most generic way to look up the referencing pin would be:
> 
> ```java
> element.findBySimpleName(interfaceElement.getName()).filter(IInterfaceElement.class::isInstance).map(IInterfaceElement.class::cast).findFirst()
> ```
> 
> In that case, I would also suggest to rename `interfaceElement` to `renamedElement` to make the intent at least a tiny bit more clear.

Done

mx990 (approved):

PR #2653, opened 2026-07-10
Apply data output properly in the interpreter network and handle ANY types properly

cochicde:
The data is stored in the interpreter network (types of the FBs). When injecting an event, the data must therefore be copied in to the right DataValues.

This operation also is moved aways from the EventManagerProcessor which shoult take care only of processing events ocurrences

Update: while working on this, I realized that whne you have an ANY_BIT type, for example in AND_2, the interprepreter and the watches were showing `DWORD#16#000000` which is not what the user wants to see. On the interpreter side, this problems came from the recursive creation of runtimes, because the complete state is created at the beginning. In the original usage of the interpreter, the variables are being created when they are needed, which smoothly pass the type from the connection to the input data for example. 

The solution here was to look for non-ANY types from the connections of a Data Input or Output and use that type. For the Watches a similar solution is used, passing a variable with the right type, and the interface element to be watche separately.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.debug.replaydebugging/src/org/eclipse/fordiac/ide/debug/replaydebugging/replayer/interpreter/ResourceInterpreterExecutor.java
+++ b/plugins/org.eclipse.fordiac.ide.debug.replaydebugging/src/org/eclipse/fordiac/ide/debug/replaydebugging/replayer/interpreter/ResourceInterpreterExecutor.java
@@ -13,9 +13,7 @@
 
 package org.eclipse.fordiac.ide.debug.replaydebugging.replayer.interpreter;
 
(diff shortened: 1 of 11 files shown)


azoitl (approved):

azoitl (approved):
LGTM. @bwiesmayr any objections?

cochicde:
I tried keeping as separated as possible of everythign that was used by the original interpreter feature. Only when needed I had to change common code, but tried to default to the original intention.

bwiesmayr:
I never considered ANY types so it makes total sense that you had to do separate things for them. I think back when I started we did not allow ANY to ANY connections yet.

bwiesmayr (approved):

PR #2654, opened 2026-07-10
Allow disabling automatic inito in sifb

cochicde:
The SIFB interpreter has currently an automatic INITO output when INIT is triggered. For replay debugging this is not desired as the actual INITO event is replayed if it was actually there.

It's not a nice solution, but at least it allows some flexibility.


Felix-Schmid (approved):
Good enough for now, IMO.

Even in the long run, we will likely have to differentiate between pure interpreter and replay debugging here, since only one has access to actual data for SIFBs.

bwiesmayr (commented):
I agree, good enough, but it was also just a sample for integrating SIFBs at all. I think reading from some models will be the way to go (e.g., try service models). Then we'd need some boolean flag for "simulate SIFB outputs" true/false to be more generic (and it would totally make sense to have that)

If it helps you now, go for it.

cochicde:
I agree. This is just a solution to have the replay debugging working. I like the idea of the service model, as for the replay debugging it would basically be an empty service model for each SIFB. One thing that I see will be missing is some timing parameters, for example, between INIT and INITO, or cyclic IND events. That way we could simulate sccenarios such as slow initialization or burst of events.

Issue #2646, opened 2026-07-07
Custom attributes on expanded struct member pins are lost when the struct member is renamed


retryoos:
When a struct member is renamed, the expanded member pins of its instances keep their configuration under the old member name. The custom attributes and the visibility of the pin are not migrated, and after the system is loaded again they are lost.

I retested this in the UI against develop:

1. Create a struct MyStruct with a BOOL member A
2. Create a simple function block with an output DO1 of type MyStruct, use it in an application and expand DO1 so that DO1.A is visible, and then connect DO1.A to another block
3. Click on the DO1.A pin, and add a custom attribute, which is then persisted as an Attribute child of the parameter DO1.A
4. Rename the struct member A to ARenamed in the data type editor and save

Right after the rename the instance still shows DO1.A with the attribute, and the file still contains the parameter DO1.A with the attribute, pointing to a member that no longer exists. After restarting the IDE the pin is shown as an error marker pin, the connection to it is broken, DO1.ARenamed is not expanded and the Attributes property sheet of the pin is empty, and saving in this state drops the attribute completely.

One more observation from testing when I right clicked on the pin and hovered over Refactor and clicked Rename Element it does nothing, no dialog opens and nothing is written to the error log, and if it's preferred I can open a separate issue for that if you want.

retryoos:
Screenshots of the reproduction:

Attribute on DO1.A before the rename
<img width="1420" height="1018" alt="Image" src="https://github.com/user-attachments/assets/aee78f18-fd4f-4081-8020-03dd1a3df339" />

MyStruct before the rename
<img width="1418" height="533" alt="Image" src="https://github.com/user-attachments/assets/76ad3cc5-8833-4390-ab12-180545de6e41" />

Rename Element from Refactor after right clicking on the pin, clicking it does nothing
<img width="837" height="642" alt="Image" src="https://github.com/user-attachments/assets/199ca28f-3118-432d-92d7-2fd237dab91e" />

After renaming A to ARenamed in the data type editor in the same session, the pin is still DO1.A and the attribute is still there
<img width="1413" height="741" alt="Image" src="https://github.com/user-attachments/assets/efe8cce5-64ef-4081-9e6d-bd61c78fded8" />

After restarting the IDE, DO1.A is an error marker pin, the connection is broken, ARenamed is not expanded and the Attributes property sheet of the pin is empty
[...]

PR #2656, opened 2026-07-10
Disable rename element refactoring for pins of block instances

retryoos:
Rename Element is now only enabled for pins in the type editors, like struct members or the interface pins of an FB type. For pins of block instances in an application it is disabled instead of silently doing nothing, as discussed in #2646.


retryoos:
@oberlehner lastly whenever you have a moment on this one, and if there anything you need from me please let me know

Issue #699, opened 2024-11-05
Set a default profile for a new Device


pedro-ricardo:
Hello everyone, things are evolving quite fast around here, it is looking awesome!

My problem is that I have a device that does not support all the communication profiles available in 4diac, so users cat get quite confused if they forgot to take a look at the Profile options after adding the device.

To make this experience better I thought about three points.

1. I'd like to issue an error if an unsupported profile is set. But I feel like it should already be doing this by using the output of `IDeviceManagementInteractorProvider.supports()` function.

2. Upon creating a new device it will be nice to be able to configure what will be the `DEFAULT_PROFILE` that will come with it once dragged into the screen.
I thought that it could be controlled though the `.dev` file in the TypeLibrary just like one can set the default value for `MGR_ID` for example.

3. Maybe another interesting feature is being able to restrict the available profiles for a device in the implementation of `IDeviceManagementInteractorProvider`. But again it looks like something that could come from the output of `supports()` function that is already there.

How do you feel about this? Does it make sense?

azoitl:
Yes this totally makes sense and is for quite some time on my wish list. I'm not yet sure how to implement it but I think it should be fully driven by the type file (`.dev`). Otherwise any changes and updates would need changes to 4diac IDE code and I don't feel this is a good idea. What we could use for this would be our new Attribute infrastructure, there we could specify an array of strings with the supported profiles and the first one being the default. What do you think?

pedro-ricardo:
I agree that changing the device type file (`.dev`) would be a good approach, it indeed looks better for maintainability.
How can I help?

azoitl:
I think the steps to be done would be the following:

1. Create a new attribute type in the core library of 4diac IDE. In difference to my comment above I guess this attribute should be only a STRING and the values then a `,` separated list.
2. Add the attribute to the device types in 4diac IDE, and add as value the list of supported device profiles. As we do not have an editor for that you would need to add this in XML
3. Update the create device command to check this attribute and the list of available device profiles and automatically set the first available from the list
4. rework the device property sheet to filter the device profiles to be selectable according to the list in the device type. 

With that you would already cover a major part of your use case. The open points would then be what to do if a device profile requested is not there. But I guess this can be discussed when we have the first basic infrastructure.

PR #2671, opened 2026-07-17
Add attribute to device type to allow filter of available Profiles

pedro-ricardo:
This Merge Request implements #699 as I undestood it was suggested.
It adds a new attribute `<Attribute Name="eclipse4diac::core::SupportedProfiles" />` to be parsed in device files (`.dev`). This attribute value is a comma separated list written as `STRING` type (single quoted), i.e. `Value="'HOLOBLOC,DynamicTypeLoad'"` 

If the attribute does not exist or if it's empty it fall back to the previous allow all behaviour. Otherwise the values written in the attribute are used to filter the `DeviceManagementInteractorFactory.INSTANCE.getAvailableProfileNames()` list.

I used the `TypeHash` attribute as an example for adding this new `SupportedProfiles` one. I hope it is a good example.

I haven't commited changes to the `data/typelibrary/system-3.0.0/typelib/devices` directly. In my point of view we can add those where it is convenient after the first review.

In order to see this implementation in action add the following line to a device file. This example here filters out the `FBDK2` from the available Profiles in FORTE_PC device.
```diff
--- a/data/typelibrary/system-3.0.0/typelib/devices/FORTE_PC.dev
+++ b/data/typelibrary/system-3.0.0/typelib/devices/FORTE_PC.dev
@@ -5,4 +5,5 @@
   <VersionInfo Organization="fortiss GmbH" Version="1.0" Author="Alois Zoitl" Date="2015-10-31" Remarks="Initial version" />
   <CompilerInfo packageName="iec61499::system"/>
   <VarDeclaration Name="MGR_ID" Type="WSTRING" InitialValue="&#34;localhost:61499&#34;" Comment="Device manager socket ID" />  
+  <Attribute Name="eclipse4diac::core::SupportedProfiles" Value="'HOLOBLOC,OPC UA,DynamicTypeLoad,Interpreter'"/>
 </DeviceType>
```

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/DEVImporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/DEVImporter.java
@@ -1,6 +1,6 @@
 /********************************************************************************
- * Copyright (c) 2008, 2009, 2011 - 2017  Profactor GmbH, TU Wien ACIN, fortiss GmbH
- * 				 2020 Johannes Kepler University Linz
+ * Copyright (c) 2008  Profactor GmbH, TU Wien ACIN, fortiss GmbH, 
+ *                                Johannes Kepler University Linz, Aimirim STI
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -13,6 +13,7 @@
  *    - initial API and implementation and/or initial documentation
  *  Alois Zoitl - Changed XML parsing to Staxx cursor interface for improved
  *  			  parsing performance
+ *  Pedro Ricardo - Added parsing of the SupportedProfiles attribute
  ********************************************************************************/
 package org.eclipse.fordiac.ide.model.dataimport;
 
@@ -24,9 +25,11 @@ import java.util.List;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.stream.XMLStreamException;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.fordiac.ide.model.LibraryElementTags;
+import org.eclipse.fordiac.ide.model.dataimport.exceptions.TypeImportException;
 import org.eclipse.fordiac.ide.model.libraryElement.DeviceType;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementFactory;
@@ -146,10 +149,13 @@ public class DEVImporter extends TypeImporter {
 
 	}
 
-	private void parseDeviceTypeAttribute() {
+	private void parseDeviceTypeAttribute() throws XMLStreamException, TypeImportException {
 		if (isProfileAttribute()) {
 			parseProfile();
+		} else {
+			parseGenericAttributeNode(getElement());
 		}
+		proceedToEndElementNamed(LibraryElementTags.ATTRIBUTE_ELEMENT);
 	}
 
 	private void parseProfile() {
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/DeviceProfileHelper.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/DeviceProfileHelper.java
@@ -1,0 +1,62 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Aimirim STI
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Pedro Ricardo - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.model.helpers;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
+import org.eclipse.fordiac.ide.model.libraryElement.ConfigurableObject;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
+import org.eclipse.fordiac.ide.model.value.StringValueConverter;
+
+/** 
+ * Helper for the 'SupportedProfiles' attribute that a device type may
+ * declare in its type file (.dev) to restrict the set of communication profiles
+ * available for its devices. */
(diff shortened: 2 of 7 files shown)


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/DeviceProfileHelper.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
```
azoitl: please add here the copyright holder. In your case most probably the company you are working for.

Please fix this also in the subsequent commits.

plugins/org.eclipse.fordiac.ide.systemconfiguration/src/org/eclipse/fordiac/ide/systemconfiguration/properties/DeviceSection.java
azoitl: please add your company to the list of copyright holders.

plugins/org.eclipse.fordiac.ide.systemconfiguration/src/org/eclipse/fordiac/ide/systemconfiguration/commands/DeviceCreateCommand.java
```diff
@@ -83,16 +89,31 @@ public void execute() {
...
 	private void setDeviceProfile() {
-		String profile;
-		if ((null != device.getType().getProfile()) && !"".equals(device.getType().getProfile())) { //$NON-NLS-1$
-			profile = device.getType().getProfile();
-		} else {
-			profile = Platform.getPreferencesService().getString(UIPreferenceConstants.FORDIAC_UI_PREFERENCES_ID,
-					UIPreferenceConstants.P_DEFAULT_COMPLIANCE_PROFILE, "", null); //$NON-NLS-1$
+		final DeviceType type = device.getType();
+		String profile = getDefaultSupportedProfile(type);
+		if (null == profile) {
```
azoitl: please change these old Yoda expressions to the right for, e.g., profile == null

azoitl (changes requested):
@pedro-ricardo thx for this work. Despite my minor comments it looks very good. There are two general points:

1. As per our contribution guidelines we would like to have a clean commit history. Do you think some of your commits should be merged together?
2. Not sure if this is for another commit, but how would you handle device profiles for which there is plugin registered. Should this be checked with the builder and result in an error marker?

pedro-ricardo:
Thank you for the review, I'll make the changes as requested.

As for que questions, I tend to commit a lot indeed but I can re-write it for sure. How would you prefer it? For a small feature like this should it be a single commit ?

As for question number 2, could you give me a practical example so I can check ?

azoitl:
> Thank you for the review, I'll make the changes as requested.
> 
> As for que questions, I tend to commit a lot indeed but I can re-write it for sure. How would you prefer it? For a small feature like this should it be a single commit ?

I normally prefer the commits in such a way as you would like to have them ending up in the repository. I'm also a fan of smaller commits and I'm happy with a PR having several commits. This makes reviewing and later bug searching easier. In your example maybe splitting into model, parseing, and ui could be a good structure.

> 
> As for question number 2, could you give me a practical example so I can check ?

You can see this in the model with the model invariant operations. These all start with `validate`. For example there are several of these in the `VarDeclaration`object. But I think this would most probably a different PR.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/helpers/DeviceProfileHelper.java
pedro-ricardo: Solved in 18e0e37

plugins/org.eclipse.fordiac.ide.systemconfiguration/src/org/eclipse/fordiac/ide/systemconfiguration/properties/DeviceSection.java
pedro-ricardo: Solved in 18e0e37

plugins/org.eclipse.fordiac.ide.systemconfiguration/src/org/eclipse/fordiac/ide/systemconfiguration/commands/DeviceCreateCommand.java
pedro-ricardo: Solved in 2c2d1c9

pedro-ricardo:
I've added two more commits with the requested changes ao you can check. If it's ok I will re-write all commits to match the pattern suggested.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/DEVImporter.java
```diff
@@ -1,6 +1,7 @@
 /********************************************************************************
  * Copyright (c) 2008, 2009, 2011 - 2017  Profactor GmbH, TU Wien ACIN, fortiss GmbH
  * 				 2020 Johannes Kepler University Linz
+ * 				 2026 AIMIRIM STI
```
azoitl: Unfortunately you touched files with old copyright headers that should be updated and fixed. The new format for the copyright statment according to the eclipse project handbook is:

 Copyright (c) creation year list of copyright holders

for example for this file:
```
* Copyright (c) 2008  Profactor GmbH, TU Wien ACIN, fortiss GmbH, 
*                                Johannes Kepler University Linz, AIMIRIM STI
```

pedro-ricardo: Ok, I wasn't aware of the outdated example. I'll fix it soon.

azoitl: Thx for your patience with me. I should have mentioned from the beginning.

azoitl: Thx for the fixes. Looks good to me now.

pedro-ricardo:
Done the commit re-write

azoitl (approved):
Thx for the clean-up and again for this very cool feature.

pedro-ricardo:
I forgot to mention but I've looked up into the validation you said ... but it seems that I had to add changes in other 3 plugins. In my opinion it makes sense as a follow-up work because it might increase the "footprint" of this PR.
But that is your call.

azoitl:
> I forgot to mention but I've looked up into the validation you said ... but it seems that I had to add changes in other 3 plugins. In my opinion it makes sense as a follow-up work because it might increase the "footprint" of this PR. But that is your call.

I expected that. I just didn't merge immediately as I wanted to wait for the CI builds. However I then forgot. Merging now.

PR #2675, opened 2026-07-20
Fixed error connection to correctly handle missing fb's/pins

sebHollersbacher:


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -210,53 +223,20 @@ public IInterfaceElement getSourceEndpoint() {
...
-			return isEmptyConnection() ? null : connection;
+			return connection;
 		}
 
 		public boolean isMissingConnectionDestination() {
 			return connectionState
 					.containsAll(EnumSet.of(ConnectionState.DEST_MISSING, ConnectionState.SOURCE_ENDPOINT_EXISTS));
 		}
 
-		public boolean isValidConnection() {
```
oberlehner, line 221: why do you delete this methods ?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
```diff
@@ -460,7 +383,7 @@ private static void parseConnectionRouting(final Connection connection, final Co
 
 	public enum ConnectionState {
 		VALID, SOURCE_MISSING, SOURCE_ENDPOINT_MISSING, DEST_MISSING, DEST_ENDPOINT_MISSING, SOURCE_EXITS,
-		SOURCE_ENDPOINT_EXISTS, DEST_EXISTS, DEST_ENPOINT_EXITS, MISSING_TYPE, DATATYPE_MISSMATCH, DUPLICATE
```
oberlehner, line 463: why do you delete MISSING_TYPE ?
We also have the case that the fb is still in the network, but only the type is deleted.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
sebHollersbacher: My mistake, I thought I could remove them as they were not used in the project

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/ConnectionHelper.java
sebHollersbacher: It was never used and also set incorrectly
I now fixed it so it's working correctly.

oberlehner (approved):

azoitl (approved):

PR #2685, opened 2026-07-21
Update tracecompass to latest version

cochicde:


azoitl (changes requested):
First of all you should update the target platform. The latest tracecompass version according to their release page https://download.eclipse.org/tracecompass/releases/ is 12.0. 
Normally we try to avoid having exact versions in manifest files to reduce maintenance effort.

plugins/org.eclipse.fordiac.ide.product/org.eclipse.fordiac.ide.product.target
```diff
@@ -13,8 +13,7 @@
  * Contributors:
  *   See git history
  *******************************************************************************
--->
```
azoitl, line 16: can you please bring back the newline that was here?

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.product/org.eclipse.fordiac.ide.product.target
cochicde: done

azoitl (approved):

PR #2699, opened 2026-07-24
Replaced if case cascads with switch statements recommended by sonar

azoitl:


sebHollersbacher (changes requested):
We should also include the `case null` in the switch statements to avoid a possible `NullPointerException`

azoitl:
> We should also include the `case null` in the switch statements to avoid a possible `NullPointerException`

Good point. I always forget that this is needed for the switch instanceof pattern matching.

sebHollersbacher (approved):

PR #2704, opened 2026-07-27
feat: Remove ASN1Tag from Struct template

franz-hoepfinger-4diac:
Eliminates the ASN1Tag as it is no longer required in the Struct.dtp template.


franz-hoepfinger-4diac:
https://github.com/franz-hoepfinger-4diac-org/4diac-ide/pull/22 Review here.

azoitl:
I fear before we can remove the ASN1 tag from the XML we also not to change 4diac FORTE and the 4diac FORTE exporter.

franz-hoepfinger-4diac:
> I fear before we can remove the ASN1 tag from the XML we also not to change 4diac FORTE and the 4diac FORTE exporter.

you mean like this ? 
https://github.com/eclipse-4diac/4diac-ide/pull/2704/changes/36cf2b97a86d46a993182c51e513b30c5bea5ebc

BTW: FORTE is hardcoded, so ASN1 can be removed, but is not wired to the IDE. 
or do you think remove the ASN1 COM layer totally from all Targets ??? 

But this seems to hard. This would remove the FBDK COM Layer completely. if i understood right, FBDK is a core component in FORTE. Isn't it ? 

the less destructive Version would be: 
https://github.com/franz-hoepfinger-4diac-org/4diac-forte/pull/22
(only remove the Tag, but it is Questionable if we need this, as it was hardcoded anyhow)

the total removal would be: 
https://github.com/franz-hoepfinger-4diac-org/4diac-forte/pull/23
but i fear this is totally not what we want. This would destroy the Full Com Stack.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/DataTypeImporter.java
```diff
@@ -105,10 +102,6 @@ protected IChildHandler getBaseChildrenHandler() {
 		};
 	}
 
-	private void parseASN1Tag() throws XMLStreamException {
-		proceedToEndElementNamed(LibraryElementTags.ASN1_TAG);
```
franz-hoepfinger-4diac, line 109: Importing old Struct DTD would give a Parse Error ? 
just asking ?

azoitl:
As written this is a bigger thing that can not just be done in a simple PR. As described in our contribution guidelines these kind of changes should be first discussed in own discussion point. Therefore I'm setting this PR for now to draft.

For starting a discussion on how we will handle the ASN1 tag on 4diac IDE and 4diac FORTE side a thorough discussion is needed. @franz-hoepfinger-4diac can you do such an analysis?

diplfranzhoepfinger:
> For starting a discussion on how we will handle the ASN1 tag on 4diac IDE and 4diac FORTE side a thorough discussion is needed. @franz-hoepfinger-4diac can you do such an analysis?

https://github.com/eclipse-4diac/4diac-ide/discussions/2725

here you go. yes. I am Ready for such a analysis.

PR #2712, opened 2026-07-30
Fix builder abort affecting all projects

mx990:
This changes the library builder to add an error marker for unresolvable projects on the project itself instead of the manifest, which is then checked by subsequent builders to ignore the project.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
+++ b/plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
@@ -12,7 +12,6 @@
  *******************************************************************************/
 package org.eclipse.fordiac.ide.library.builder;
 
-import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


MandKastner (approved):

plugins/org.eclipse.fordiac.ide.library/src/org/eclipse/fordiac/ide/library/builder/LibraryBuilder.java
```diff
@@ -79,22 +76,12 @@ protected IProject[] build(final int kind, final Map<String, String> args, final
 
 	@Override
 	protected void clean(final IProgressMonitor monitor) throws CoreException {
-		final SubMonitor progress = SubMonitor.convert(monitor, Messages.LibraryBuilder_CleaningLibrary, 3);
+		final SubMonitor progress = SubMonitor.convert(monitor, Messages.LibraryBuilder_CleaningLibrary, 1);
 
-		// clean manifest library marker
-		final IFile manifestFile = getProject().getFile(LibraryManager.MANIFEST);
-		if (manifestFile.exists()) {
```
azoitl, line 86: There may still be markers at the manifest file (see line 716 of `LibraryManager` below). Shouldn't we remove them on clean?

mx990: The new version uses `deleteMarkers` with `IResource.DEPTH_INFINITE` to delete all library markers within the project, including the manifest file.

azoitl: Stupid me. I totally missed that. This heat is not good for my brain.

azoitl (approved):

PR #2714, opened 2026-07-31
Fix connection policy not updating style field

Felix-Schmid:
The style field of the `ConnectionMovementHighlightEditPolicy` is no longer updated since PR #2699, causing the cursor to still be displayed correctly, but the connection tool is always used.


azoitl (approved):
Thx for finding this issue.

PR #2717, opened 2026-08-01
[Replay Debugging] Fix zooming issue in timeline view and highlight steps also when selecting the watch

cochicde:
- Zooming for connections is applied only after everything is calculated
- When the user selects a monitoring value (watch) it also hightlights steps in the timeline, similar to selecting the actual interface.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/SelectionService.java
+++ b/plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/SelectionService.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.fordiac.ide.deployment.debug.ui.annotation.WatchValueAnnotation;
 import org.eclipse.fordiac.ide.model.libraryElement.FBNetworkElement;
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/SelectionService.java
```diff
@@ -85,6 +86,8 @@ public void selectionChanged(final IWorkbenchPart part, final ISelection selecti
 			} else if (ep.getModel() instanceof final org.eclipse.fordiac.ide.model.libraryElement.Connection mo) {
 				lastSelectedDatapoints.add(mo.getSource().getQualifiedName());
 				lastSelectedDatapoints.add(mo.getDestination().getQualifiedName());
+			} else if (ep.getModel() instanceof final WatchValueAnnotation watch) {
```
azoitl: by now this if else cascade would read much cleaner with switch instanceof pattern matching

azoitl (changes requested):

plugins/org.eclipse.fordiac.ide.debug.replaydebugging.ui/src/org/eclipse/fordiac/ide/debug/replaydebugging/ui/SelectionService.java
cochicde: done

azoitl (approved):

PR #2728, opened 2026-08-05
Fix broken connections for untyped subapps in resource deployment data

mx990:
There were broken connections (between different resources) for untyped subapps in the resource deployment data, because we did not go back to the resource network when looking for the external interface element.

This changes `getSubAppExternalElement()` to mirror the behavior of `getSubAppInternalElement()` and go back to the mapped untyped subapp.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/data/ResourceDeploymentData.java
+++ b/plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/data/ResourceDeploymentData.java
@@ -44,7 +44,6 @@ import org.eclipse.fordiac.ide.model.libraryElement.Resource;
 import org.eclipse.fordiac.ide.model.libraryElement.SubApp;
 import org.eclipse.fordiac.ide.model.libraryElement.SubAppType;
 import org.eclipse.fordiac.ide.model.libraryElement.TypedSubApp;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/data/ResourceDeploymentData.java
```diff
@@ -292,10 +291,16 @@ private static IInterfaceElement getSubAppInternalElement(final IInterfaceElemen
 	}
 
 	private static IInterfaceElement getSubAppExternalElement(final IInterfaceElement element, final SubApp subApp) {
-		if (subApp instanceof UntypedSubApp) {
+		final InterfaceList interfaceList;
+		if (subApp instanceof TypedSubApp) {
+			interfaceList = subApp.getInterface();
```
azoitl, line 296: what if the typed subapp is mapped?

mx990: If the typed subapp is mapped, `subApp` will be in the resource network, since that is where we started.

azoitl: ah

azoitl (approved):

PR #2745, opened 2026-08-14
Cleanup util plugin by removing ColorManager

Felix-Schmid:


azoitl (approved):

azoitl:
Github copilot is wrong here. Since a few Eclipse versions colors do not consume os handles anymore.

PR #2780, opened 2026-08-26
Add generic E_D_FF_1..5 Service Interface FBTypes (BOOL family)

franz-ms-muc:
Interface-only declarations (D1..Dn/Q1..Qn BOOL, EO fires only if at least
one Di differs from the latched Qi), following the same architecture as
E_REND_2/3/4 / OR_2..9: no BasicFB body, GenericClassName='GEN_E_D_FF'.
Native FORTE implementation (GEN_E_D_FF_fbt.h/.cpp) still needed before
these can actually run - interface declaration only for now.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01WLvA6NB3nnZU73938b3pAU

### DIFF
--- a/data/typelibrary/events-3.0.0/typelib/E_D_FF_1.fbt
+++ b/data/typelibrary/events-3.0.0/typelib/E_D_FF_1.fbt
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FBType Name="E_D_FF_1" Comment="Data latch (d) flip flop for 1 BOOL, EO only on change of at least one (generic FB)">
+	<Identification Standard="61499-1 Annex A" Description="Copyright (c) 2017, 2026 fortiss GmbH, HR Agrartechnik GmbH&#10;&#10;This program and the accompanying materials are made&#10;available under the terms of the Eclipse Public License 2.0&#10;which is available at https://www.eclipse.org/legal/epl-2.0/&#10;&#10;SPDX-License-Identifier: EPL-2.0">
+	</Identification>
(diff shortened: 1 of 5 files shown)


azoitl:
@franz-ms-muc as you also have the any version why a bool version as well. This just pollutes the library.

franz-hoepfinger-4diac:
> @franz-ms-muc as you also have the any version why a bool version as well. This just pollutes the library.

you are right. 
i made it just to stream-line with the current E_D_FF.fbt // E_D_FF_ANY.fbt
we could also have changed E_D_FF into ANY, and not make 2 Versions there.

i guess either we need a Discussion 
-or- 
a Master Decision from you. 
i would be fine with any solution.

azoitl:
`E_D_FF` is defined in the standard so this one I do not want to be changed.

For the new block we are more flexible. As the new Blocks would anyhow be `E_D_FF_x` it would be different then the one from the standard so we could that one also give the any IOs. And such simplify the lib. WDTY?

franz-hoepfinger-4diac:
> `E_D_FF` is defined in the standard so this one I do not want to be changed.

which i fully understand. 
> 
> For the new block we are more flexible. As the new Blocks would anyhow be `E_D_FF_x` it would be different then the one from the standard so we could that one also give the any IOs. And such simplify the lib. WDTY?

I only used the ANY Versions of these blocks up to now. 
this the BOOL Versions can be dropped.

franz-ms-muc:
closing in Favour of https://github.com/eclipse-4diac/4diac-ide/pull/2781

PR #2786, opened 2026-08-27
Fix package name and descriptions in signalprocessing typelib

franz-ms-muc:
Qualifies the ST PACKAGE declarations with the eclipse4diac:: prefix to match CompilerInfo packageName, corrects the swapped FIELDBUS_PERCENT_TO_WORD/FIELDBUS_WORD_TO_PERCENT descriptions, and cleans up comments and formatting.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XFZNahFYaa5F1bd2iKfPZi
https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-ide/pull/33

PR #2814, opened 2026-09-02
Qualify PACKAGE declaration with eclipse4diac:: prefix in signalprocessing typelib

franz-ms-muc:
FIELDBUS_PERCENT_TO_WORD, FIELDBUS_WORD_TO_PERCENT, SCALE, and SCALE_LIM declared `PACKAGE signalprocessing;` in their ST body, but their `CompilerInfo packageName` is `eclipse4diac::signalprocessing` - the unqualified PACKAGE declaration doesn't match. Qualifies all four to `PACKAGE eclipse4diac::signalprocessing;` to match.

Split out of #2786, which bundled this with unrelated description and formatting fixes - those are in #2786's other half, opened separately.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01KZMwou1aMoeNFPiXrJszsa

### DIFF
--- a/data/typelibrary/signalprocessing-3.0.0/typelib/FIELDBUS_PERCENT_TO_WORD.fct
+++ b/data/typelibrary/signalprocessing-3.0.0/typelib/FIELDBUS_PERCENT_TO_WORD.fct
@@ -2,7 +2,7 @@
 <Function Name="FIELDBUS_PERCENT_TO_WORD" Comment="Convert a REAL in the Range 0.0% to 100.0% (0.0 to 1.0) to a WORD Value Range 0-FAFF">
 	<Identification Standard="SAE J1939; ISO 11783" Description="Copyright (c) 2023 HR Agrartechnik GmbH     &#10;   &#10;This program and the accompanying materials are made    &#10;available under the terms of the Eclipse Public License 2.0    &#10;which is available at https://www.eclipse.org/legal/epl-2.0/    &#10;    &#10;SPDX-License-Identifier: EPL-2.0">
 	</Identification>
-	<VersionInfo Organization="HR Agrartechnik GmbH" Version="3.0.1" Author="Franz Höpfinger" Date="2026-08-27" Remarks="updated description and cleaned up formatting">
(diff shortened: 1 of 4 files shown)


azoitl (changes requested):
please increase minor version number to indicate package name change. 
Please fix merge conflict.

franz-ms-muc:
> please increase minor version number to indicate package name change. 

again, the Package name is **NOT** changed. 
the rest i will address now.

azoitl:
> > please increase minor version number to indicate package name change.
> 
> again, the Package name is **NOT** changed. the rest i will address now.

are you shure? I thought for functions the function body is defining evrything and the rest is just there because it is.

franz-ms-muc:
> > > please increase minor version number to indicate package name change.
> > 
> > 
> > again, the Package name is **NOT** changed. the rest i will address now.
> 
> are you shure? 
no

> I thought for functions the function body is defining evrything and the rest is just there because it is.
if i am looking to FORTE it seems to be other way round ? 

https://github.com/eclipse-4diac/4diac-forte/blob/0535a2dc3b7fab036a8dc7e89d8de6c2c48f70cd/modules/signalprocessing/src/eclipse4diac/signalprocessing/FIELDBUS_PERCENT_TO_WORD_fct.cpp#L58

https://github.com/eclipse-4diac/4diac-forte/blob/0535a2dc3b7fab036a8dc7e89d8de6c2c48f70cd/modules/signalprocessing/src/eclipse4diac/signalprocessing/FIELDBUS_PERCENT_TO_WORD_fct.cpp#L37

franz-ms-muc:
BTW i did update the ~~minor~~ maintenance  Version, 
but other day you said we do not do update the Version 2 times in a Release cycle ? 

i like it more this way, as otherwise we must merge the Change Comment in the VersionInfo ... so pls have a look.

azoitl:
> BTW i did update the minor Version, but other day you said we do not do update the Version 2 times in a Release cycle ?

As I commented in another PR you mixed up minor and maintenance version. And yes it should not be bumped twice in a cycle.

> 
> i like it more this way, as otherwise we must merge the Change Comment in the VersionInfo ... so pls have a look.

azoitl (changes requested):
As Inow better understand the problem and what is the root cause of it, I have to say that you where right. IT should not be a new maintainence version but you shoul add the documentation of your change to the 3.0.1 version.

azoitl (approved):

PR #2817, opened 2026-09-02
Fix false unused-import warning for imports used in System Configuration

### LINKED ISSUE #1764: Unused Import Warning (false warning)
Image

Image

Image

Martin Jobst @mx990 --> That is because we are currently not checking inside System Configuration

franz-ms-muc:
## Summary
- The whole-project builder that computes used/unused imports (`STAlgorithmInitialValueBuilderParticipant`) pruned the entire `SystemConfiguration` subtree, so `Device`/`Resource`/FB-instance parameters referencing an import (e.g. a Global Constant used as a Device's `MGR_ID` or as an FB instance parameter inside a `Resource`) were never seen and got reported as `Unused import`.
- Every elementary data type (e.g. `WSTRING`, `BOOL`) extends `AnyType`, so the pre-existing `varDeclaration.getType() instanceof AnyType` check in `validateValue` already covers concrete-typed parameter values such as `MGR_ID` — removing the prune is sufficient to let them be parsed and their import usage registered; no widening of that condition is needed.
- Replaced the blanket `SystemConfiguration` prune with a narrower one for `FBNetworkElement`s that are the mapped-to side of a `Mapping` (the shadow copy of an Application FB instance inside a `Resource`'s `FBNetwork`), since those are already validated as part of the Application network and would otherwise be processed a second time.

Fixes #1764

An unrelated `ErrorDataType` hardening for the same `validateValue` methods was split out per review into #2826.

## Test plan
- Reproduced the false-positive on a real project (`.sys` files using Global Constants as `Device`/`Resource`/FB-instance `Parameter` values, e.g. `MGR_ID="C_11"`, FB `Parameter ID="Q1_REMOTE_WRITE"`) — warnings for those imports disappeared after a rebuild with the fix applied.
- Verified imports that genuinely don't exist are still reported as `The import '...' does not exist` errors.
- Verified genuinely unused imports are still reported as `Unused import` warnings.
- Built the affected plugins plus the `org.eclipse.fordiac.ide.product` reactor via `mvn install -pl plugins/org.eclipse.fordiac.ide.product,plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui,plugins/org.eclipse.fordiac.ide.structuredtextcore.ui -am -DskipTests` — `BUILD SUCCESS`.

## References

- https://github.com/eclipse-4diac/4diac-ide/pull/2817
- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-ide/pull/35
- https://github.com/eclipse-4diac/4diac-ide/pull/2826
- https://github.com/eclipse-4diac/4diac-ide/issues/1764

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
+++ b/plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
@@ -9,6 +9,8 @@
  *
  * Contributors:
  *   Martin Jobst - initial API and implementation and/or initial documentation
+ *   Franz Höpfinger - fix false unused-import warnings for values used only
+ *                      in System Configuration
  *******************************************************************************/
 package org.eclipse.fordiac.ide.structuredtextalgorithm.ui.builder;
 
@@ -41,10 +43,10 @@ import org.eclipse.fordiac.ide.model.helpers.PackageNameHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
 import org.eclipse.fordiac.ide.model.libraryElement.BaseFBType;
 import org.eclipse.fordiac.ide.model.libraryElement.ECTransition;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetworkElement;
 import org.eclipse.fordiac.ide.model.libraryElement.Import;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementPackage;
-import org.eclipse.fordiac.ide.model.libraryElement.SystemConfiguration;
 import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
 import org.eclipse.fordiac.ide.model.value.TypedValueConverter;
 import org.eclipse.fordiac.ide.structuredtextalgorithm.resource.STAlgorithmResource;
@@ -157,7 +159,8 @@ public class STAlgorithmInitialValueBuilderParticipant implements IXtextBuilderP
 				throw new OperationCanceledException();
 			}
 			switch (allContents.next()) {
-			case final SystemConfiguration _ -> allContents.prune();
+			case final FBNetworkElement fbNetworkElement when fbNetworkElement.isMapped()
+					&& fbNetworkElement.getMapping().getTo() == fbNetworkElement -> allContents.prune();
 			case final Attribute attribute -> {
 				validateType(attribute, delta, typeUsageCollector, ignoreWarnings, context, monitor);
 				validateValue(attribute, delta, typeUsageCollector, variableUsageValidator, ignoreWarnings, context,


plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -228,6 +226,14 @@ protected void validateValue(final VarDeclaration varDeclaration, final IResourc
...
 				issues.replaceAll(issue -> ValidationUtil.convertToModelIssue(issue, varDeclaration.getValue()));
 			}
+		} else if (!value.isBlank() && ValidationUtil.isContainedInSystemConfiguration(varDeclaration)) {
+			// Device/Resource/FB instance Parameter values keep their concrete type (e.g.
+			// copied from the DeviceType or the FB's interface pin), but may still reference
+			// an imported symbol (e.g. a Global Constant) by unqualified name. Parse the
+			// value only to register that usage, discarding any parse issues, since the
+			// value format itself is not validated here (see the AnyType branch above).
+			validate(varDeclaration.getValue(), delta, typeUsageCollector, variableUsageValidator, new ArrayList<>(),
+					context);
```
mx990: No, we should only skip elements that are already handled elsewhere, that is, elements mapped from an application network. If it is an original value set by the user, we want *both* type usage and issues.

I believe we should instead add the following above in place of `SystemConfiguration`:
```java
case final FBNetworkElement fbne when fbne.isMapped() && fbne.getMapping().getTo() == fbne -> allContents.prune();
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
franz-ms-muc: Good point — done. Folded the System Configuration path into the existing AnyType branch so both type usage and issues are reported the same way for original user-set values, and added the prune you suggested for FBNetworkElements that are the mapped-to side of a Mapping, to avoid double-processing the Application-mapped shadow copies. See 051bd5f47.

franz-ms-muc:
Re the Copilot review (https://github.com/eclipse-4diac/4diac-ide/pull/2817#pullrequestreview-5089481133): that's expected and intentional, not a regression — @mx990 explicitly asked for this in an earlier review comment ("If it is an original value set by the user, we want *both* type usage and issues"), which is why the System-Configuration path was folded into the same branch as the AnyType case instead of discarding issues. The PR description was stale (still described the earlier discard-issues version); updated it to match the current behavior.

franz-ms-muc:
I realize this could go to 3.3.x as well, as this is a Bugfix .... hold on ... 

EDIT: done with https://github.com/eclipse-4diac/4diac-ide/pull/2820 with Cherry-pick, as a Rebase was not possible. 
hope this helps.

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -219,8 +222,11 @@ protected void validateValue(final VarDeclaration varDeclaration, final IResourc
 			final IBuildContext context, final IProgressMonitor monitor) throws CoreException {
 		final String value = getValue(varDeclaration);
 		final List<Issue> issues = new ArrayList<>();
-		// do not parse value if blank or variable has invalid type
-		if (!value.isBlank() && varDeclaration.getType() instanceof AnyType) {
+		// AnyType variables always need parsing; System Configuration Parameters (e.g.
+		// Device/Resource/FB instance parameters) keep a concrete type but may still
+		// reference an imported symbol, such as a Global Constant, by unqualified name
+		if (!value.isBlank() && (varDeclaration.getType() instanceof AnyType
+				|| ValidationUtil.isContainedInSystemConfiguration(varDeclaration))) {
```
mx990: Please revert this change. The condition `varDeclaration.getType() instanceof AnyType` checks that we have a *valid* type (e.g., not an error type) and has nothing to do with the system configuration issue. We should not attempt to parse initial values with an invalid type, since type errors are already handled elsewhere.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
mx990: > Removing the prune alone was not enough, since Device/Resource/FB
instance parameter values keep a concrete type (e.g. WSTRING) rather
than AnyType, and value parsing for import-usage tracking was gated on
AnyType.

To clarify, concrete types (e.g., `WSTRING`) inherit from `AnyType`, so they are already covered by the condition.

franz-ms-muc: Good catch on the underlying concern, but I don't think a full revert is right here — this branch is exactly what fixes the reported bug, and reverting it would reintroduce the false-positive `Unused import` warnings.

On `instanceof AnyType` being unrelated to System Configuration: the actual shipped device type disagrees — `data/typelibrary/system-3.0.0/typelib/devices/FORTE_PC.dev:7`:
```xml
<VarDeclaration Name="MGR_ID" Type="WSTRING" InitialValue="&#34;localhost:61499&#34;" Comment="Device manager socket ID" />
```
`MGR_ID` (the exact Device Parameter from the bug report) is concretely `WSTRING`, not `AnyType`, so the pre-existing branch alone never parses it — that's the whole reason the import-usage tracking missed it in the first place. This was also verified end-to-end against a real project (multiple `.sys` files using Global Constants as `MGR_ID`/FB-instance parameter values): the false warnings only disappear with this branch in place.

On the actual valid part of your concern (parsing values with an invalid/unresolved type): you're right that we shouldn't do that. I added an explicit exclusion instead of a revert — `ErrorDataType` is now excluded from *both* the `AnyType` and the System Configuration path (it previously slipped through the `AnyType` check too, since `ErrorDataType extends AnyDerivedType extends AnyType`). See d4952ab18.

Happy to adjust further if there's a different angle I'm missing.

franz-ms-muc: Correction to my reply above — I was wrong, sorry for the noise. You're right that `WSTRING` (and every other elementary type) extends `AnyType`: `WstringType extends AnyStringType extends AnyCharsType extends AnyElementaryType extends AnyType` (same for `BoolType`). So `MGR_ID` was already covered by the pre-existing `instanceof AnyType` check all along — the `SystemConfiguration` prune removal plus your `FBNetworkElement`-mapped-to prune is the actual fix, and my `isContainedInSystemConfiguration` widening was unnecessary.

Reverted it (and removed the now-unused `ValidationUtil.isContainedInSystemConfiguration` helper). Kept the `ErrorDataType` exclusion, since that part still stands independent of the above. See b4f2db7f3.

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -219,8 +223,12 @@ protected void validateValue(final VarDeclaration varDeclaration, final IResourc
...
 		final List<Issue> issues = new ArrayList<>();
-		// do not parse value if blank or variable has invalid type
-		if (!value.isBlank() && varDeclaration.getType() instanceof AnyType) {
+		// do not parse value if blank or variable has invalid type; ErrorDataType is
+		// excluded too, since unresolved types are already reported elsewhere and
+		// re-parsing their value would only add noise (note that it would otherwise
+		// pass this check, since ErrorDataType extends AnyDerivedType extends AnyType)
+		if (!value.isBlank() && varDeclaration.getType() instanceof AnyType
+				&& !(varDeclaration.getType() instanceof ErrorDataType)) {
 			final LibraryElement featureType = STCoreUtil.getFeatureType(varDeclaration);
```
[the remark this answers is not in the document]
franz-ms-muc: This is based on a mistaken premise (same one a human reviewer raised and I initially agreed with, then verified and retracted — see the thread on the original widened version of this line): every elementary `DataType` in this model extends `AnyType`, not just the generic/type-inferred ones. Concretely: `WstringType extends AnyStringType extends AnyCharsType extends AnyElementaryType extends AnyType` (same chain for `BoolType`, etc.). So `varDeclaration.getType() instanceof AnyType` is already true for a concrete `WSTRING`-typed variable such as `MGR_ID` — there's no additional `isContainedInSystemConfiguration` widening needed.

This was also verified end-to-end: rebuilding a real project with only the `SystemConfiguration`-prune removal (no widening) makes the false `Unused import` warnings for `MGR_ID`/FB-instance parameter values disappear, while genuinely unused/nonexistent imports are still correctly flagged. No change needed here.

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -219,8 +223,12 @@ protected void validateValue(final VarDeclaration varDeclaration, final IResourc
...
 		final String value = getValue(varDeclaration);
 		final List<Issue> issues = new ArrayList<>();
-		// do not parse value if blank or variable has invalid type
-		if (!value.isBlank() && varDeclaration.getType() instanceof AnyType) {
+		// do not parse value if blank or variable has invalid type; ErrorDataType is
+		// excluded too, since unresolved types are already reported elsewhere and
+		// re-parsing their value would only add noise (note that it would otherwise
+		// pass this check, since ErrorDataType extends AnyDerivedType extends AnyType)
+		if (!value.isBlank() && varDeclaration.getType() instanceof AnyType
+				&& !(varDeclaration.getType() instanceof ErrorDataType)) {
```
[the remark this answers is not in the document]
franz-ms-muc: Good catch, this was a real gap. Fixed: for `ErrorDataType`, we now still call `validate(...)` to register import/variable usage from the value, just with a throwaway issues list so it doesn't produce a second, misleading `INITIAL_VALUE_MARKER` for a variable whose type is already reported as broken elsewhere. See 5486d9531.

franz-ms-muc:
Following up on the same split requested on #2820: removed the ErrorDataType exclusion from this PR too (4a4c9ded3) — it's now its own standalone PR: #2826 (covers both validateValue overloads, VarDeclaration and Attribute).

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -157,7 +160,8 @@ protected void validateContents(final Resource resource, final IResourceDescript
 				throw new OperationCanceledException();
 			}
 			switch (allContents.next()) {
-			case final SystemConfiguration _ -> allContents.prune();
+			case final FBNetworkElement fbNetworkElement when fbNetworkElement.isMapped()
+					&& fbNetworkElement.getMapping().getTo() == fbNetworkElement -> allContents.prune();
 			case final Attribute attribute -> {
```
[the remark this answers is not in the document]
franz-ms-muc: Correct, stale — that helper was reverted per review (turned out unnecessary, since every elementary type already extends AnyType). Updated the PR description to match the current diff.

plugins/org.eclipse.fordiac.ide.structuredtextalgorithm.ui/src/org/eclipse/fordiac/ide/structuredtextalgorithm/ui/builder/STAlgorithmInitialValueBuilderParticipant.java
```diff
@@ -225,8 +229,18 @@ protected void validateValue(final VarDeclaration varDeclaration, final IResourc
 			try {
 				new TypedValueConverter((DataType) featureType, true).toValue(value);
 			} catch (final Exception e) {
-				validate(varDeclaration.getValue(), delta, typeUsageCollector, variableUsageValidator, issues, context);
-				issues.replaceAll(issue -> ValidationUtil.convertToModelIssue(issue, varDeclaration.getValue()));
+				if (varDeclaration.getType() instanceof ErrorDataType) {
+					// the type itself is unresolved and already reported elsewhere; still
+					// register any import/variable usage referenced by the value, but
+					// discard the issues so we don't create a second, likely misleading
+					// INITIAL_VALUE_MARKER for it
```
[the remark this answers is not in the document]
franz-ms-muc: Stale — that ErrorDataType branch (and its comment) was removed from this PR entirely, split into #2826 per review feedback. The comment you're quoting is accurate again for the current code, which no longer has that branch.

azoitl:
@franz-ms-muc why do we have two PRs for the same thing? what is the difference betwwen this PR and #2820?

franz-ms-muc:
close this in Favour of https://github.com/eclipse-4diac/4diac-ide/pull/2820 as this one merges to develop, and https://github.com/eclipse-4diac/4diac-ide/pull/2820 merges to 3.3.x, and 3.3.x will later anyhow be merged into Develop.

PR #2818, opened 2026-09-02
Create error library element on load errors in library element resource

mx990:


azoitl (approved):
Fixes the problem. As this is for me a kind of bug fix should we move this to 3.3.x?

Maybe for another PR and for develop: can we somehow get rid of the code duplication for error types between resources and typeentries?

mx990:
> Fixes the problem. As this is for me a kind of bug fix should we move this to 3.3.x?

Good to hear. Yes, moved to 3.3.x.

> Maybe for another PR and for develop: can we somehow get rid of the code duplication for error types between resources and typeentries?

Yes, that is also on my list.

PR #2819, opened 2026-09-02
Load directly from resource in library element provider

mx990:
The library element provider did copy stale library elements from the type entry when `synchronize()` was called before the resource change listener had a chance to `refresh()` the type entry.

This loads library elements directly from a resource in the library element provider to always receive an up-to-date copy.

This PR depends on https://github.com/eclipse-4diac/4diac-ide/pull/2818 to create error library elements in case of load errors, otherwise there will be an exception.


azoitl (approved):
Fixes my original problem. As this is for me a kind of bug fix should this also be moved to 3.3.x?

mx990:
> Fixes my original problem. As this is for me a kind of bug fix should this also be moved to 3.3.x?

Good to hear. Yes, moved to 3.3.x.

Issue #2816, opened 2026-09-02
Organize Imports is partly broken


franz-hoepfinger-4diac:
4diac IDE nightly
Version: 3.3.0
Build id: 2026-08-27_2003

Organize Imports does Find GCFs from the current directory, but 
those in "External Libraries" are no more recognized. 
this was working before. (namely April 2026 Version)

do you need a Minimal Sample project ? 

Thanks, 
Franz

franz-ms-muc:
Found two independent root causes under this one issue, split into separate PRs:

- #2828: External-library import resolution — FB-instance parameters mirrored through an adapter interface (e.g. `Input`/`Output` pins on IXA/QXA-style wrapper FBs) were never registered with a search-support factory at all, so Organize Imports silently skipped them regardless of whether the referenced constant was local or external. This also explains the wildcard-import-deletion case.
- #2829: The `TypeHash` bookkeeping attribute was pulling in a needless, functionally unnecessary import.

franz-ms-muc:
closing, as https://github.com/eclipse-4diac/4diac-ide/pull/2828 is merged, and https://github.com/eclipse-4diac/4diac-ide/pull/2829 is closed.

PR #2828, opened 2026-09-04
Register missing VarDeclaration search factories

franz-ms-muc:
## Summary
- `Organize Imports` (and, by the same mechanism, the unused-import check) never correctly detected usage of a Global Constant referenced from an FB-instance `Parameter` that is mirrored through an adapter interface (e.g. an `Input`/`Output` pin on an IXA/QXA-style wrapper FB), regardless of whether the referenced constant was defined locally or in an external library. In the worst case (a wildcard import combined with such a parameter) the existing import got deleted outright instead of kept or corrected.
- Root cause: such parameters are modeled as `ContainerVarDeclaration` or `VarConfigInstance` - both subtypes of `VarDeclaration` - but the `org.eclipse.fordiac.ide.model.search.factory` extension point in `org.eclipse.fordiac.ide.model.search.st/plugin.xml` only declared the plain `VarDeclaration` type as a source. Since `ISearchFactory`'s registry does an exact-class lookup (no supertype matching), `ISearchFactory.createSearchSupport()` returned `null` for these subtypes, so `OrganizeImportsCommand`'s whole-model traversal silently skipped them - the import-usage computation never saw the reference at all.
- Registered the two subtypes alongside `VarDeclaration`, both in the extension point (what the running IDE actually uses) and in `StructuredTextSearchFactory.register()` (the equivalent path used by unit tests, so both stay in sync).

Verified against a real reproduction project: an FB-instance `Output` parameter referencing an external-library Global Constant is now correctly recognized, its import gets re-added when missing, and it's no longer dropped when running Organize Imports with a wildcard import present.

This addresses the "External Libraries" part of #2816. A second, independent root cause reported under the same issue (the `TypeHash` attribute pulling in a needless import) is fixed separately in #2829.

## Test plan
- Added `VarDeclarationSearchSupportTest.test{ContainerVarDeclaration,VarConfigInstance}HasSearchSupport`, each asserting `ISearchFactory.createSearchSupport(...)` returns a `VarDeclarationSearchSupport` instance instead of `null`. Confirmed these tests fail without the fix and pass with it.
- Manually verified end-to-end in a real project: an `Output` parameter referencing an external-library constant is correctly re-imported by Organize Imports after the fix (previously required manually re-adding the import, and a related wildcard-import case had the import deleted outright).

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/StructuredTextSearchFactory.java
+++ b/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/StructuredTextSearchFactory.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2022, 2023 Martin Erich Jobst
+ * Copyright (c) 2022 Martin Erich Jobst
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -9,15 +9,18 @@
  *
  * Contributors:
  *   Martin Jobst - initial API and implementation and/or initial documentation
+ *   Franz Höpfinger - register VarDeclaration subtypes for search support
  */
 package org.eclipse.fordiac.ide.model.search.st;
 
 import org.eclipse.fordiac.ide.model.data.DirectlyDerivedType;
 import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
+import org.eclipse.fordiac.ide.model.libraryElement.ContainerVarDeclaration;
 import org.eclipse.fordiac.ide.model.libraryElement.ECTransition;
 import org.eclipse.fordiac.ide.model.libraryElement.STAlgorithm;
 import org.eclipse.fordiac.ide.model.libraryElement.STFunctionBody;
 import org.eclipse.fordiac.ide.model.libraryElement.STMethod;
+import org.eclipse.fordiac.ide.model.libraryElement.VarConfigInstance;
 import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
 import org.eclipse.fordiac.ide.model.search.ISearchFactory;
 import org.eclipse.fordiac.ide.model.search.ISearchSupport;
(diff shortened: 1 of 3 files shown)


diplfranzhoepfinger:
This as well
@azoitl I guess this needs to go to 3.3.x right?

plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/StructuredTextSearchFactory.java
```diff
@@ -57,6 +61,9 @@ public static void register() {
 		ISearchFactory.Registry.INSTANCE.registerFactory(STMethod.class, factory);
 		ISearchFactory.Registry.INSTANCE.registerFactory(STFunctionBody.class, factory);
 		ISearchFactory.Registry.INSTANCE.registerFactory(VarDeclaration.class, factory);
+		ISearchFactory.Registry.INSTANCE.registerFactory(ContainerVarDeclaration.class, factory);
+		ISearchFactory.Registry.INSTANCE.registerFactory(LocalVariable.class, factory);
```
mx990: `LocalVariable` is obsolete and not used anymore. I have created PR https://github.com/eclipse-4diac/4diac-ide/pull/2832 to remove it.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/StructuredTextSearchFactory.java
franz-ms-muc: Thanks, was already wondering, as `LocalVariable` is a concept which is not existing in IEC 61499.

mx990 (changes requested):
Please also change the commit message to be less verbose. We do not need all the details about where and how the problem originally occurred. That belongs in the issue. The fact that subtypes of `VarDeclaration` were not registered in both the search factory and plugin is enough for the first paragraph.

plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/StructuredTextSearchFactory.java
franz-ms-muc: @mx990 issue addressed, LocalVariable removed everywhere.

mx990 (approved):

azoitl (approved):

PR #2829, opened 2026-09-04
Exclude TypeHash attribute from Organize Imports

franz-ms-muc:
## Summary
- Whenever a type file carries the tool-managed `<Attribute Name="eclipse4diac::core::TypeHash" Value="''"/>` bookkeeping attribute, Organize Imports added a functionally useless `<Import declaration="eclipse4diac::core::TypeHash"/>` to it.
- `TypeHash` is a real, resolvable `AttributeType` (`data/typelibrary/core-3.0.0/typelib/TypeHash.atp`, package `eclipse4diac::core`), so `AttributeSearchSupport.getImportedNamespaces()` correctly found it had a non-empty package and added its namespace - but its XML `Name` is already the fully-qualified name, so resolution never actually depends on ST scoping or an import. `TypeHash` isn't in `InternalAttributeDeclarations.allAttributes` either (that list only covers synthetic Java-only attributes like `VAR_CONFIG`/`VISIBLE`), so none of the existing `isInternalAttribute()` checks caught this case.
- Added a specific exclusion for `TypeHash` in `AttributeSearchSupport.getImportedNamespaces()`.

This is one of the symptoms reported under #2816 - a separate, independent root cause from the External-Library import-resolution issue also reported there (unrelated code paths, fixed separately in #2828).

## Test plan
- Added `AttributeSearchSupportTest.testTypeHashAttributeExcludedFromImportedNamespaces` (asserts the namespace is not reported) and `testOtherAttributeIncludedInImportedNamespaces` (asserts a non-`TypeHash` attribute declaration still is, so the general mechanism isn't weakened). Confirmed the first test fails without the fix and passes with it.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/AttributeSearchSupport.java
+++ b/plugins/org.eclipse.fordiac.ide.model.search.st/src/org/eclipse/fordiac/ide/model/search/st/AttributeSearchSupport.java
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *   Martin Jobst - initial API and implementation and/or initial documentation
+ *   Franz Höpfinger - exclude TypeHash attribute from imported namespaces
  */
 package org.eclipse.fordiac.ide.model.search.st;
 
@@ -20,6 +21,7 @@ import org.eclipse.emf.ecore.EObject;
 import org.eclipse.fordiac.ide.model.data.AnyType;
 import org.eclipse.fordiac.ide.model.helpers.PackageNameHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.Attribute;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
 import org.eclipse.fordiac.ide.structuredtextalgorithm.util.STAlgorithmParseUtil;
 import org.eclipse.xtext.parser.IParseResult;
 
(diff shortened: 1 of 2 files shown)


diplfranzhoepfinger:
@azoitl I guess this needs to go to 3.3.x right?

franz-ms-muc:
One Question here: My Sourcery Review generated one Comment: 
https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-ide/pull/39#discussion_r3932273174
it is about `eclipse4diac::core::TypeHash` vs. legacy short name `TypeHash`.
while this is legit, i guess the old naming will be deprecated soon. 

what do you think ?

mx990:
While it is true that resolution does not depend on the import, how the name is displayed in the attribute section does. Otherwise, adding imports for attributes would not make any sense at all.

The `TypeHash` attribute is also only intended for use in standard libraries and special cases in library development. It is definitely not intended for general use. I therefore do not believe we should add a special exception for it here.

franz-ms-muc:
> The TypeHash attribute is also only intended for use in standard libraries and special cases in library development. It is definitely not intended for general use. I therefore do not believe we should add a special exception for it here.

I agree. This PR https://github.com/eclipse-4diac/4diac-ide/pull/2829 is no head-burner, while https://github.com/eclipse-4diac/4diac-ide/pull/2828 definitely is. 

BTW: 
the same is true for `eclipse4diac::core::GenericClassName` (as well as `eclipse4diac::core::TypeHash` discussed here)
plus the Legacy Versions of the 1.0.0 Lib `GenericClassName` and `TypeHash`.

the End-user only makes "sub" Files, no "fbt" Files, and the `TypeHash` is only relevant for fbt. so this is not something annoying the End-user a lot. Also this import does not generate a Warning.

having 

```
		<Import declaration="eclipse4diac::core::GenericClassName"/>
		<Import declaration="eclipse4diac::core::TypeHash"/>
```

in a fbt File does not annoy the User as the IDE Silently ignore is, and does not produce a "unused" Warning, so nothing bad happens. 


So, 

- we could let this Fix out, as End-users are unaffected and Lib Desingers should know how to handle ? 
- Or we could extend it to also not generate the `GenericClassName` Import as well, 
- Or we could add these 2 Imports to all affected Files in the Typelib, if the right decision is that Attributes should have this Import declaration, and as well drop this fix.

mx990:
I believe we should just leave this as-is for now.

PR #2833, opened 2026-09-04
Filter content assist by expected type

### LINKED ISSUE #1853: CTRL + SPACE show too much stuff
Hi, 

4diac-ide_3.0.0-win32.win32.x86_64_rc_2025-11-19_1607

using:
https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac_training1/releases/tag/v0.21

open any Übung. 
say Uebung_001

and Type in "Input_I1" the CTRL + SPACE 

it show too much constants. here are only a few possible. others are not type matched. 

in this case: only "Input_I1" till "Input_I10" or so ... 
but there are shown about 1000 constants.

franz-ms-muc:
## What & why

Value-position content assist (an FB instance parameter's `Value`, a
global constant reference, or any other typed initializer field)
proposed every visible symbol regardless of its declared type. A
single field could list hundreds/thousands of type-mismatched global
constants, standard functions, and bare data type names alongside the
handful of actually relevant candidates - see #1853.

## Fix

- `completeSTFeatureExpression_Feature`: boost proposals matching the
  field's expected type to the top (via the existing, previously
  content-assist-unused, resource-level expected-type tracking),
  instead of excluding non-matches outright. A struct-typed candidate
  that doesn't itself match may still be the correct start of a member
  access whose type only resolves later (e.g. `aStruct.field`), which
  a hard filter would incorrectly hide - see
  [this review comment](https://github.com/eclipse-4diac/4diac-ide/pull/2833#pullrequestreview-5117146512).
  Declared types are compared **by URI string**, not by resolving them
  into objects: elementary types (e.g. `UINT`) are in-memory singletons
  addressed by a synthetic URI that cannot be loaded as a resource -
  attempting to resolve it either throws (with a forced load) or
  silently lets everything through (without one).
- Functions, FB types, and bare data type names are still excluded,
  but only in a genuine constant-only position (an initializer value),
  where none of them is ever valid syntax regardless of type - this is
  a syntactic exclusion, unrelated to the type-matching above, and
  never applies in a general algorithm/method body expression (e.g. an
  assignment's right-hand side), where a function call is fine.
- The same-shaped `TypeName#value` typed-literal completions
  (`STStructInitializerExpression`, `STNumericLiteral`,
  `STDateLiteral`, `STTimeLiteral`, `STTimeOfDayLiteral`,
  `STDateAndTimeLiteral`, `STStringLiteral`) keep the stricter exact-
  match **filter** rather than a ranking boost, since there is no
  member-access escape hatch for choosing a type name - if the type
  doesn't match here, the literal is never valid.

## Verified

Manually, live, in a built IDE, against the real repro project linked
in #1853 (`u16ObjId` on a `Softkey_IXA` instance expecting `UINT`, and
`Output` on a `logiBUS_QXA` instance expecting a `logiBUS_DO_S`
struct): empty field, partial prefix, and full match all now rank
type-compatible constants first. Also verified the member-access case
directly (`OutputNumber_c6_N.u16ObjId`, a struct-typed FB instance
navigated via `.` to reach a compatible field) still offers
`OutputNumber_c6_N` as a candidate, and that a general ST algorithm
assignment's right-hand side is unaffected (still shows every visible
symbol, matching pre-PR behavior).

No automated test is included: there is currently no test
infrastructure in this repo for Xtext content-assist/proposal-provider
behavior (a real Eclipse workbench/SWT completion popup), and adding
that infrastructure from scratch felt disproportionate to this fix -
flagging this gap rather than skipping it silently.

Fixes #1853.

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/contentassist/STCoreProposalProvider.java
+++ b/plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/contentassist/STCoreProposalProvider.java
@@ -15,18 +15,25 @@
  *       - exclude proposals based disallowing qualified names only
  *       - add proposal for callables
  *       - enable proposals for global variables in packages
+ *   Franz Höpfinger
+ *       - rank feature expression proposals by expected type and exclude
+ *         non-values in constant-only positions
  */
 package org.eclipse.fordiac.ide.structuredtextcore.ui.contentassist;
 
 import java.util.Collections;
 import java.util.List;
 
+import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EReference;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.fordiac.ide.globalconstantseditor.globalConstants.GlobalConstantsPackage;
 import org.eclipse.fordiac.ide.globalconstantseditor.globalConstants.STVarGlobalDeclarationBlock;
+import org.eclipse.fordiac.ide.model.data.DataType;
 import org.eclipse.fordiac.ide.model.helpers.ImportHelper;
 import org.eclipse.fordiac.ide.model.libraryElement.ICallable;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElement;
 import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementPackage;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeEntry;
 import org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary;
@@ -34,7 +41,11 @@ import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
 import org.eclipse.fordiac.ide.structuredtextcore.resource.STCoreResourceDescriptionStrategy;
 import org.eclipse.fordiac.ide.structuredtextcore.services.STCoreGrammarAccess;
 import org.eclipse.fordiac.ide.structuredtextcore.stcore.STCorePackage;
+import org.eclipse.fordiac.ide.structuredtextcore.stcore.STExpression;
 import org.eclipse.fordiac.ide.structuredtextcore.stcore.STImport;
+import org.eclipse.fordiac.ide.structuredtextcore.stcore.STInitializerExpression;
+import org.eclipse.fordiac.ide.structuredtextcore.stcore.STResource;
+import org.eclipse.fordiac.ide.structuredtextcore.stcore.util.STCoreUtil;
 import org.eclipse.fordiac.ide.structuredtextcore.util.STCoreRegionString;
 import org.eclipse.jface.text.contentassist.ICompletionProposal;
 import org.eclipse.jface.viewers.StyledString;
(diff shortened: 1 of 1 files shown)


mx990 (changes requested):
I do not believe we should straight up filter out any proposals that do not match the expected type. First of all, the expected type may not always be available or accuracte while typing. More importantly, though, later expressions may actually change the meaning, for example:
```st
VAR_TEMP
  X : MyStruct; // (VAR1 : INT, ...)
  Y : INT;
END_VAR
Y :=
```
With your change, triggering content assist after `Y := ` would filter out `X`, since `MyStruct` is not assignable to `INT`. However, the user may have wanted to type `X.VAR1`.

I believe the most we can and should do here is sort the proposals to have those with matching types at the top.

franz-ms-muc:
> I do not believe we should straight up filter out any proposals that do not match the expected type.

@mx990 Thanks for your comment, 
i was totally not aware that the `.` representation is allowed in 4diac. 

just gave it a test, 
at least i get not Error, (did not test a Deploy) 

image
 

just also Realized, that, if the Import is there before (impossible with organitze Imports, but givent the case) there is also a Syntax completion: 

image



so, yes, i must address this, THANKS.

franz-ms-muc:
@mx990 changed it, and tested it. 

now the directly matching things comes first, 
the indirect matching come after.

plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/contentassist/STCoreProposalProvider.java
```diff
@@ -84,10 +95,163 @@ protected static boolean isKeywordWorthyToPropose(final Keyword keyword) {
...
+		// name never valid syntax - exclude those there. Everything else is merely
+		// ranked by expected-type match, never excluded: a struct-typed candidate may
+		// still be the correct start of a member access (e.g. "aStruct.field") even
+		// though its own type does not match, and there is no attempt here to search
+		// struct members or apply numeric widening rules to decide that up front.
+		final boolean constantOnly = isConstantOnlyPosition(model);
+		final Predicate<IEObjectDescription> filter = description -> isVisible(description)
+				&& (!constantOnly || isValueCandidate(description));
+		final Function<IEObjectDescription, ICompletionProposal> factory = getTypeRankedProposalFactory(context,
+				grammarAccess.getSTFeatureNameRule().getName(), model, expectedType);
```
mx990: As mentioned already, please do not attempt to filter out proposals based on possibly incomplete or inaccurate information while the user is typing.

plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/contentassist/STCoreProposalProvider.java
```diff
@@ -84,10 +95,163 @@ protected static boolean isKeywordWorthyToPropose(final Keyword keyword) {
...
+	public void completeSTStringLiteral_Type(final EObject model, final Assignment assignment,
+			final ContentAssistContext context, final ICompletionProposalAcceptor acceptor) {
+		completeTypedLiteralType(model, assignment, context, acceptor);
+	}
+
+	protected void completeTypedLiteralType(final EObject model, final Assignment assignment,
+			final ContentAssistContext context, final ICompletionProposalAcceptor acceptor) {
+		final LibraryElement expectedType = getExpectedType(model);
+		final Predicate<IEObjectDescription> filter = description -> isVisible(description)
+				&& isLiteralTypeCompatible(description, expectedType);
```
mx990: As mentioned already, please do not attempt to filter out proposals based on possibly incomplete or inaccurate information while the user is typing.

plugins/org.eclipse.fordiac.ide.structuredtextcore.ui/src/org/eclipse/fordiac/ide/structuredtextcore/ui/contentassist/STCoreProposalProvider.java
```diff
@@ -84,10 +95,163 @@ protected static boolean isKeywordWorthyToPropose(final Keyword keyword) {
...
+			public ICompletionProposal apply(final IEObjectDescription candidate) {
+				final ICompletionProposal proposal = super.apply(candidate);
+				if (proposal instanceof final ConfigurableCompletionProposal configurableProposal
+						&& isExpectedType(model, candidate, expectedType)) {
+					configurableProposal.setPriority(configurableProposal.getPriority() + TYPE_MATCH_PRIORITY_BOOST);
+				}
+				return proposal;
+			}
+		};
+	}
```
mx990, line 120: The proper way to adjust the priority of proposals is via `STCoreContentProposalPriorities`, not here.

mx990 (changes requested):

Issue #2584, opened 2026-06-16
Comments on Service disappear


franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-ide/issues/2520

i guess i have found a similar one than the one above.

```xml
	<Service RightInterface="RESOURCE" LeftInterface="APPLICATION" Comment="Subscribe to a PUBLISH_10 Block">
	</Service>
```

becomes

```xml
	<Service RightInterface="RESOURCE" LeftInterface="APPLICATION">
	</Service>
```

on a re-save.

franz-hoepfinger-4diac:
just re-confirm: this is still the case in 4diac-ide_3.3.0-win32.win32.x86_64_nightly_2026-08-21_2006

franz-ms-muc:
Fixed in #2835.

Issue #2520, opened 2026-06-01
Transition Comments: Disappear


diplfranzhoepfinger:
<img width="734" height="210" alt="Image" src="https://github.com/user-attachments/assets/dc18a13b-b658-4281-998a-7131e3becac1" />

The Storing of the Comments work. 
they are present in the XML File. PERFECT. 

but at re-opening, they are no more displayed. 
another "Save" deletes them.

PR #2835, opened 2026-09-04
Fix Service comment being read onto the wrong element

franz-ms-muc:
## What & why

Opening a `.fbt` file with a `ServiceInterfaceFBType`'s `Service`
`Comment` attribute set, then saving it again, silently dropped the
comment - see #2584. Same symptom class as #2520 (`ECTransition`
comments), fixed for that element in #2521.

## Root cause

`BlockTypeImporter.parseService(FBType type)` called
`readCommentAttribute(type)` - but `readCommentAttribute(INamedElement)`
sets the comment on the `INamedElement` it's given, and `type` here is
the *containing* `FBType`, not the `Service` being parsed (`Service`
isn't itself an `INamedElement`, so it couldn't be passed directly).
This silently set the `FBType`'s own comment instead of the
`Service`'s.

Since `FbtExporter`'s `addService()` already correctly writes
`sfb.getService().getComment()`, a comment set through the UI *was*
present in the file after the first save - but because it was never
read back into `Service.comment` on the next import, that in-memory
value was empty, and the following save wrote out an empty comment,
i.e. none at all.

## Fix

Read the comment value and set it on `type.getService()` directly,
using the same `readCommentAttribute()` (no-arg, returns
`Optional<String>`) helper introduced by #2521 for the analogous fix.

## Verified

Added `ServiceCommentImportExportTest`: builds a `ServiceInterfaceFBType`
with a `Service` comment in memory, exports it via `FbtExporter`,
re-imports the result via `FBTImporter`, and asserts the comment
survived the round trip. Confirmed it fails (`expected: <...> but was:
<>`) without the fix and passes with it.

Also verified manually: an `.fbt` with a `Service Comment="..."`
attribute now keeps the comment across an open/save cycle instead of
losing it.

- https://github.com/eclipse-4diac/4diac-ide/issues/2584
- https://github.com/eclipse-4diac/4diac-ide/issues/2520
- https://github.com/eclipse-4diac/4diac-ide/pull/2521

### DIFF
--- a/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/BlockTypeImporter.java
+++ b/plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataimport/BlockTypeImporter.java
@@ -20,6 +20,7 @@
  *  Martin Jobst - refactor marker handling
  *  Alois Zoitl  - updated for new adapter FB handling
  *  Martin Jobst - extract interface importer
+ *  Franz Höpfinger - fix Service comment being read onto the wrong element
  ********************************************************************************/
 package org.eclipse.fordiac.ide.model.dataimport;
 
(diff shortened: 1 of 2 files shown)


franz-ms-muc:
> No automated test is included

NOW we have tests as well. Test was tested, also Counter-Test made.

azoitl (changes requested):
Commit message is still horrible.

franz-ms-muc:
> Commit message is still horrible.

i fixed it.

azoitl (approved):

PR #2743, opened 2026-08-13
Fix missing adapter initial values for composite FBs in forte_ng export

mx990:
The export did not include initial values for adapters in composite FBs.

Issue #2737, opened 2026-08-11
Struct-typed value on a plain FB instance parameter is not exported


franz-hoepfinger-4diac:
When a composite FB contains a regular (non-adapter) FB instance whose input variable has a struct type, and that variable is given a value — either a global-constant reference or a bare struct literal like `(code := 255, bCyclic := TRUE)` — the forte_ng export generates an empty assignment instead of the actual value:

```cpp
fb_BtnFirst->var_InputEvent = ;
```

This is a compile error in the generated C++ code, but 4diac IDE itself shows no validation error on the source `.fbt`.

Reproduction: `ScrollFS_PHYS.fbt`

https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac_training1/blob/feature/scroll/Ventilsteuerung/4diacIDE-workspace/.lib/isobus-3.0.0/typelib/UT/Q/ScrollFS_PHYS.fbt

Related to #2743, which fixed the same symptom for `AdapterDeclaration` parameters — this is the same bug for a plain FB instance's own input variable.

Root cause and fix: #2836.

<img width="417" height="363" alt="Image" src="https://github.com/user-attachments/assets/52fa2988-b369-40da-a686-fb6d487dc77f" />

azoitl:
This FB should have compile errors in 4diac IDE. In that case you should not be able to export it or we can not correctly export it. I'm more concerned why you are not getting any compile errors in 4diac IDE. Is build automatically turned off?

franz-hoepfinger-4diac:
found a second, much simpler case: 

the "TRUE" here is not exported in a Composite FB. 

<img width="1222" height="471" alt="Image" src="https://github.com/user-attachments/assets/8c639f75-c852-4ef2-961c-8508a938ecc5" />

mx990:
Is it just missing the value, as in the example above, or is the entire initial value assignment not exported for the adapter? If it is the latter, I have provided a fix in PR https://github.com/eclipse-4diac/4diac-ide/pull/2743.

franz-hoepfinger-4diac:
> Is it just missing the value, as in the example above, or is the entire initial value assignment not exported for the adapter? If it is the latter, I have provided a fix in PR [#2743](https://github.com/eclipse-4diac/4diac-ide/pull/2743).

i experienced both: 
https://github.com/eclipse-4diac/4diac-ide/issues/2737#issue-5119781080
here it exported the Assignment but miss the Value, 

here
[...]

PR #2836, opened 2026-09-05
Register missing VarDeclaration types for forte_ng codegen

franz-ms-muc:
## What & why

A regular (non-adapter) FB instance's struct-typed input variable,
given a value (either a plain global-constant reference, or a bare
struct literal like `(code := 255, bCyclic := TRUE)`), was exported
with an empty right-hand side, producing invalid C++ - see #2737. PR
#2743 already fixed the same symptom for adapter parameters; this
covers a separate, still-open path for plain FB instances.

## Root cause

Same bug class as #2828 (a different registry, same mistake). When a
regular `VarDeclaration`'s declared type is a `StructuredType`,
`InterfaceListCopier`/`VarDeclarationFactory` creates a
`ContainerVarDeclaration` for the per-instance copy (used for later
struct-member decomposition) instead of a plain `VarDeclaration`. The
forte_ng codegen's `ILanguageSupportFactory.Registry` does an
exact-class lookup (no supertype matching) keyed by
`org.eclipse.fordiac.ide.export.language.supportFactory`'s declared
`<source type="...">` entries, which only listed the plain
`VarDeclaration` type. So `ILanguageSupportFactory.createLanguageSupport("forte_ng", ...)`
returned `null` for the `ContainerVarDeclaration` case *before* ever
reaching `StructuredTextSupportFactory`'s own `instanceof VarDeclaration`
dispatch (which already handles it correctly, polymorphically) -
`CompositeFBImplTemplate`'s
`fbNetworkInitialVariableLanguageSupport` map then legitimately stored
`null` for that variable, and `generateFBNetworkInitialValue` silently
generated nothing for it.

## Fix

Register `ContainerVarDeclaration` and `VarConfigInstance` alongside
`VarDeclaration`, both in the extension point (what the running IDE
actually uses) and in `StructuredTextSupportFactory.register()` (the
equivalent path used by unit tests, so both stay in sync).

## Verified

Added `CompositeFBStructValueExportTest`: builds a composite FB
containing a plain (non-adapter) FB instance whose struct-typed input
variable has a struct-literal value, runs the actual forte_ng export
headlessly, and asserts the generated `setFBNetworkInitialValues()`
assignment is non-empty. Confirmed it fails
(`fb_LeafInstance->var_InputEvent = ;`) without the fix and passes
(`... = forte::CIEC_TestScrollStruct(255_DINT, true_BOOL);`) with it.

- https://github.com/eclipse-4diac/4diac-ide/issues/2737

### DIFF
--- a/tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
+++ b/tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
@@ -1,0 +1,112 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Franz Höpfinger
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Franz Höpfinger - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.test.export;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.fordiac.ide.model.data.DataFactory;
+import org.eclipse.fordiac.ide.model.data.StructuredType;
+import org.eclipse.fordiac.ide.model.datatype.helper.IecTypes.ElementaryTypes;
+import org.eclipse.fordiac.ide.model.helpers.BlockInstanceFactory;
+import org.eclipse.fordiac.ide.model.libraryElement.CompositeFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.FB;
+import org.eclipse.fordiac.ide.model.libraryElement.FBNetwork;
+import org.eclipse.fordiac.ide.model.libraryElement.LibraryElementFactory;
+import org.eclipse.fordiac.ide.model.libraryElement.SimpleFBType;
+import org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;
+import org.eclipse.fordiac.ide.model.resource.FordiacTypeResourceFactory;
+import org.eclipse.fordiac.ide.model.typelibrary.FBTypeEntry;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
+import org.eclipse.fordiac.ide.test.model.typelibrary.DataTypeEntryMock;
+import org.eclipse.fordiac.ide.test.model.typelibrary.FBTypeEntryMock;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that a struct literal assigned to a struct-typed input variable of
+ * a plain (non-adapter) FB instance is exported with that value as the
+ * generated assignment's right-hand side.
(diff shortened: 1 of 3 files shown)


tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.eclipse.fordiac.ide.model.typelibrary.FBTypeEntry;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibrary;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryManager;
+import org.eclipse.fordiac.ide.model.typelibrary.TypeLibraryTags;
+import org.eclipse.fordiac.ide.test.model.typelibrary.DataTypeEntryMock;
+import org.eclipse.fordiac.ide.test.model.typelibrary.FBTypeEntryMock;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Reproduces https://github.com/eclipse-4diac/4diac-ide/issues/2737: a
```
franz-ms-muc: @azoitl is this Link in the Code OK, or should be removed ???

franz-ms-muc (commented):
the Rest of the Code for me looks clean, logical, minimal. 
The Test was tested and counter-Tested.

tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
```diff
@@ -0,0 +1,121 @@
...
+		FordiacTypeResourceFactory.INSTANCE
+				.createResource(URI.createFileURI(LEAF_FB_NAME + TypeLibraryTags.FB_TYPE_FILE_ENDING_WITH_DOT))
+				.getContents().add(leafType);
+
+		final FB leafInstance = BlockInstanceFactory.createFBInstanceForTypeEntry(leafEntry);
+		leafInstance.setName(LEAF_INSTANCE_NAME);
+		leafInstance.setInterface(leafEntry.getInterface().instanceCopy());
+		leafInstance.setTypeEntry(leafEntry);
+		leafInstance.getInterface().getInputVars().stream().filter(v -> STRUCT_VAR_NAME.equals(v.getName()))
+				.findFirst().orElseThrow().setValue(createValue(STRUCT_LITERAL_VALUE));
```
mx990:
```suggestion
		leafInstance.getInterface().getVariable(STRUCT_VAR_NAME).setValueString(STRUCT_LITERAL_VALUE);
```

tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
mx990: I would personally prefer if we avoided direct links to GitHub in the source code.

mx990 (changes requested):
@azoitl Should this fix also go in `3.3.x`?

tests/org.eclipse.fordiac.ide.test.export/src/org/eclipse/fordiac/ide/test/export/CompositeFBStructValueExportTest.java
azoitl: Even more I would say the test comment, if even necessary, should explain what it tests and what the expected result is, not what was wrong.

azoitl (changes requested):
Yes I also think this should be in 3.3.x as it is a rather critical bug fix.

@franz-ms-muc can you please rebase on top of 3.3.x and update the PR's target branch.

mx990 (approved):

azoitl (approved):

PR #2843, opened 2026-09-07
Added InstanceAttributes for TypedSubApps

sebHollersbacher:


oberlehner:
image

It is not verified if the element is really part of the subapp instance.
For example adding random attributes in the normal atttribute sheet like a.b.c ist listed as subapp instance attribute after a restart of 4diac. 

We need to check if the element is really accessible .

Also a GOTO function would be nice.

PR #2846, opened 2026-09-07
Introduce isMemberAccessPin helper function to interfaceelement

azoitl:


plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
```diff
@@ -1737,6 +1737,11 @@
         <details key="body" value="return org.eclipse.fordiac.ide.model.libraryElement.impl.InterfaceElementAnnotations.getBlockRelativePath(this);"/>
       </eAnnotations>
     </eOperations>
+    <eOperations name="isMemberAccessPin" lowerBound="1" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean">
+      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+        <details key="body" value="return eContainer() instanceof org.eclipse.fordiac.ide.model.libraryElement.VarDeclaration;"/>
```
mx990: Since this is in `IInterfaceElement`, would it not make more sense to check for `IInterfaceElement` instead of `VarDeclaration`?

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
mx990, line 1: It seems this has become a large diff due to save actions. Please re-generate.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
```diff
@@ -285,8 +284,7 @@ private void addConnection(final Connection connection, final FBNetwork fbNetwor
 	}
 
 	private static boolean isExportableConnectionEndpoint(final IInterfaceElement endPoint) {
-		return (endPoint != null)
-				&& (endPoint.eContainer() instanceof InterfaceList || endPoint.eContainer() instanceof VarDeclaration);
+		return (endPoint != null) && (endPoint.eContainer() instanceof InterfaceList || endPoint.isMemberAccessPin());
```
mx990:
```suggestion
		return endPoint != null && endPoint.getInterfaceList() != null;
```

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
azoitl: Interesting I do not remember that I saved that. Will regenerate and test.

plugins/org.eclipse.fordiac.ide.model/model/lib.ecore
azoitl: I thought about that. Decided against it if we later want to add other member access pins. At that point I'm not so sure anymore. The reason for having it in IInterfaceElement is that it makes its users much simpler.

azoitl: After a second thought I think it is better to be more specific for now and widen later when needed. Changed as suggestested.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/dataexport/FBNetworkExporter.java
azoitl: done

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
azoitl: it looks like the file was wrongly formated bevore. I deleted the whole file and regenerated it. There was one generated not inside. @mx990 @oberlehner should this one be preserved?

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
```diff
@@ -330,32 +354,30 @@ public DataType getType() {
...
-	 *
-	 * @generated NOT
+	 * <!-- begin-user-doc -->
+	 * <!-- end-user-doc -->
+	 * @generated
 	 */
 	@Override
-	public void setType(final DataType newType) {
-		java.util.Objects.requireNonNull(newType,
-				"DataType of a VarDeclaration must not be null! " + getQualifiedName());
```
mx990, line 348: It looks like this got lost re-generating the code.

mx990 (changes requested):

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
mx990: As commented below, I believe we should preserve it.

azoitl: good then I will implement it in the right EMF idiom as it should have been.

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
azoitl: added again

plugins/org.eclipse.fordiac.ide.model/src-gen/org/eclipse/fordiac/ide/model/libraryElement/impl/VarDeclarationImpl.java
azoitl: done

mx990 (approved):

PR #2865, opened 2026-09-10
skip projects without package registry

oberlehner:


ernstblechaPT (approved):
lgtm

disconnect could be handled with a finally-block

ernstblechaPT (approved):

azoitl (approved):

Issue #2872, opened 2026-09-11
DualHysteresis: no direct UP/DOWN transition, causes spurious Neutral cycle


franz-ms-muc:
## Description

`DualHysteresis` (`data/typelibrary/signalprocessing-3.0.0/typelib/DualHysteresis.fbt`) implements a two-sided hysteresis with three states: `UP`, `DOWN`, `Neutral`. The ECC only has transitions from `UP`/`DOWN` back to `Neutral`, and from `Neutral` to `UP`/`DOWN` — there is no direct `UP -> DOWN` or `DOWN -> UP` transition.

## Problem

If `INPUT` crosses both switch-on thresholds within a single `REQ` event (a scan-to-scan jump larger than `2 * (ABS(DEAD) + ABS(HYSTERESIS))`), the FB cannot reach the correct state in that same event:

1. FB is in `UP`, `REQ` fires.
2. The only outgoing transition from `UP` is `UP -> Neutral` (`INPUT < MI + ABS(DEAD)`), which is satisfied since `INPUT` dropped far below `MI`.
3. FB lands in `Neutral`, outputs `DO_UP = FALSE`, `DO_DOWN = FALSE`, and the `REQ` event is consumed.
4. The `Neutral -> DOWN` transition is also guarded by `REQ`, so it is not evaluated again in the same cycle — only unconditional (`Condition="1"`) transitions chain within one event in IEC 61499 ECC semantics.

Result: one scan cycle with neither `DO_UP` nor `DO_DOWN` set, before the FB reaches `DOWN` on the next `REQ`.

This is a narrow edge case (requires a large single-scan swing), but it is a real, silent one-cycle output glitch that a downstream consumer (e.g. valve/relay control) could observe.

## Suggested fix

Add `UP -> DOWN` and `DOWN -> UP` transitions with the same switch-on conditions as `Neutral -> DOWN`/`Neutral -> UP`, evaluated before the existing switch-off-to-`Neutral` transitions (ECC evaluates a state's outgoing transitions in document order, first match wins):

```
UP   --REQ[INPUT <= MI - ABS(DEAD) - ABS(HYSTERESIS)]--> DOWN    (new)
UP   --REQ[INPUT <  MI + ABS(DEAD)]------------------->  Neutral (existing)
DOWN --REQ[INPUT >= MI + ABS(DEAD) + ABS(HYSTERESIS)]--> UP      (new)
[...]

PR #2873, opened 2026-09-11
Clarify DualHysteresis docs, guard signs with ABS

franz-ms-muc:
## What

- Spells out the actual switch-on/switch-off formulas in the FBType and VarDeclaration comments instead of the generic "can be 0.1 meaning 10%" wording.
- Wraps `DEAD` and `HYSTERESIS` in `ABS()` in the ECC transition conditions.
- Bumps VersionInfo to 3.1.

## Why

`DEAD`/`HYSTERESIS` were used unwrapped in the threshold math, so a negative parameter value would silently invert which side of `MI` a switch point falls on. `ABS()` makes the switch points match the documented formulas regardless of parameter sign.

Prerequisite for the direct UP/DOWN transition fix in a follow-up PR (related to #2872).


franz-ms-muc:
https://github.com/eclipse-4diac/4diac-ide/pull/2874 is merged, so no need to merge this. closing.

PR #2875, opened 2026-09-13
Add common error library element factories

mx990:
This adds common factories for error library elements and data types to avoid duplicate code in type entries and library element resources. Also improve uri handling in type library manager.


plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/impl/ErrorLibraryElementFactoryImpl.java
```diff
@@ -0,0 +1,171 @@
...
+		this.delegate = delegate;
+	}
+
+	@Override
+	public ErrorLibraryElement create(final String fullTypeName, final EClass eClass) {
+		if (eClass.getEPackage() == DataPackage.eINSTANCE) {
+			return ErrorDataTypeFactory.INSTANCE.create(fullTypeName, eClass);
+		}
+		if (eClass.getEPackage() != LibraryElementPackage.eINSTANCE) {
+			throw new IllegalArgumentException("Not a valid class " + eClass); //$NON-NLS-1$
```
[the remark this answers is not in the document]
mx990: @copilot If we intend to support extension packages providing their own library elements, then the check in `CreateDataTypeEntry` would need to be narrowed instead of broadening the scope of the error factory.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/ErrorLibraryElementFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ *   Martin Jobst - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.model.libraryElement;
+
+import org.eclipse.emf.ecore.EClass;
+import org.eclipse.fordiac.ide.model.libraryElement.impl.ErrorLibraryElementFactoryImpl;
+
+public interface ErrorLibraryElementFactory {
+
+	ErrorLibraryElementFactory INSTANCE = new ErrorLibraryElementFactoryImpl(LibraryElementFactory.eINSTANCE);
```
azoitl, line 20: would we ever change this to something else? If not we could also set it in the constructor to simplify the code. Or would this help for tests.

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/data/ErrorDataTypeFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ *   Martin Jobst - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+package org.eclipse.fordiac.ide.model.data;
+
+import org.eclipse.emf.ecore.EClass;
+import org.eclipse.fordiac.ide.model.data.impl.ErrorDataTypeFactoryImpl;
+
+public interface ErrorDataTypeFactory {
+
+	ErrorDataTypeFactory INSTANCE = new ErrorDataTypeFactoryImpl(DataFactory.eINSTANCE);
```
azoitl, line 20: as below would we ever provide something else?

plugins/org.eclipse.fordiac.ide.model/src/org/eclipse/fordiac/ide/model/libraryElement/ErrorLibraryElementFactory.java
mx990: I have added a default constructor to the error factories. That way, we can avoid passing the factory here and also allow passing a different factory in case of future tests or extensions.

azoitl (approved):

PR #2884, opened 2026-09-15
Create ECC state on commit of the inline editor

vikash1703:
Double-clicking on the empty ECC canvas opens an inline text editor with a template state name. The state is only created when the user commits the name, so cancelling the edit leaves the model unchanged.

The root edit part forwards the open request to the ECC edit part, which delegates to a direct edit policy that creates the state from the entered name and the clicked position.


plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/policies/ECCCreateStateDirectEditPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+	public void performDirectEdit(final LocationRequest request) {
+		creationLocation = request.getLocation().getCopy();
+		new ECCStateCreationDirectEditManager(getHost(), creationLocation).show();
+	}
+
+	@Override
+	protected Command getDirectEditCommand(final DirectEditRequest request) {
+		if (creationLocation != null && request.getCellEditor() != null
+				&& request.getCellEditor().getValue() instanceof final String name && !name.isBlank()) {
+			final Point location = creationLocation.getCopy();
```
azoitl: please use `request.getLocation().getCopy()` Then you do not need to store the creationLocation. This would be cleaner.

azoitl (changes requested):
Works very nicely. I have only a minor remark for the edit policy

plugins/org.eclipse.fordiac.ide.fbtypeeditor.ecc/src/org/eclipse/fordiac/ide/fbtypeeditor/ecc/policies/ECCCreateStateDirectEditPolicy.java
vikash1703: @azoitl 
Done i have used request.getLocation().getCopy() to make it cleaner but i encounter if we do not override DirectEditRequest createDirectEditRequest() then we do not get state while double click so i added that as well as same FB network do in NewInstanceDirectEditManager i have done in managerclass.

azoitl (approved):

